PR #2, opened 2026-05-22
feat: introduce base utilities including Status, Defs, Executor, and MemoryPool

lxy-9602:
### Purpose

No Linked issue.

Introduce the foundational utility modules from alibaba/paimon-cpp to bootstrap the apache/paimon-cpp codebase. These modules provide core infrastructure (status handling, type definitions, memory management, executor) that other components depend on.

- Introduce `Status`, `Result<T>`, and related utilities (`status.cpp`, `status_test.cpp`, `status_utils.h`)
- Introduce type definitions and constants (`defs.h`, `defs.cpp`, `type_fwd.h`)
- Introduce `Executor` interface and default implementation (`executor.h`, `executor.cpp`)
- Introduce `MemoryPool` and `Bytes` with unit tests (`memory_pool.h/cpp`, `memory_pool_test.cpp`, `bytes.h/cpp`, `bytes_test.cpp`)
- Introduce shared utilities: `compare.h`, `macros.h`, `string_builder.h`, `result.h`, `traits.h`, `visibility.h`
- Introduce test harness (trimmed to status assertion macros only)
- Preserve third-party attribution comments (Apache Arrow, RocksDB/LevelDB, Google LLC/LiteRT)

### Tests

- `status_test.cpp` — Status and Result type correctness
- `memory_pool_test.cpp` — MemoryPool allocation and lifecycle
- `bytes_test.cpp` — Bytes buffer operations

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -206,6 +206,7 @@ This product includes code from RocksDB.
 
 * endian utility in src/paimon/common/utils/math.h
 * uuid utility in src/paimon/common/utils/uuid.h
+* Assert utilities in src/paimon/testing/utils/testharness.cpp and src/paimon/testing/utils/testharness.h
 
 Copyright: 2011-present, Facebook, Inc.  All rights reserved.
 Copyright: 2011 The LevelDB Authors. All rights reserved.
--- a/LICENSE
+++ b/LICENSE
@@ -243,3 +243,25 @@ THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 --------------------------------------------------------------------------------
+
+This product includes code from Apache Arrow.
+
+* Core utilities:
+  * docs utilities in docs/ directory
+  * basic utilities in
+    - include/paimon/compare.h
+    - include/paimon/macros.h
+    - include/paimon/status.h
+    - include/paimon/string_builder.h
+    - src/paimon/common/utils/status.cpp
+* Build system modules:
+  * cmake_modules/BuildUtils.cmake
+  * cmake_modules/DefineOptions.cmake
+  * cmake_modules/san-config.cmake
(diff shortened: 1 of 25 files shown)


leaves12138 (changes requested):
Thanks for migrating the base utilities. I compared the added files with the existing source snapshot; most of the code matches the source migration with ASF header normalization, and a lightweight public-header syntax check passes.

One blocker before merge: this PR introduces files that carry third-party attribution markers, but it does not update `LICENSE` / `NOTICE` for those newly introduced files.

Examples:
* `include/paimon/compare.h`, `include/paimon/macros.h`, `include/paimon/status.h`, `include/paimon/string_builder.h`, and `src/paimon/common/utils/status.cpp` are marked as adapted from Apache Arrow / LevelDB.
* `src/paimon/testing/utils/testharness.cpp` and `src/paimon/testing/utils/testharness.h` are marked as adapted from RocksDB / LevelDB.

Please add the corresponding Apache Arrow and RocksDB / LevelDB attribution/license entries to `LICENSE` and `NOTICE` in this PR, scoped to the files introduced here. After that I can re-review the migration.

leaves12138 (changes requested):
Thanks for the update. The Apache Arrow attribution has been added and covers the Arrow-derived files introduced here.

There is still one small licensing blocker: the RocksDB / LevelDB license text and NOTICE are already present from PR #5, but the `LICENSE` section currently lists only `src/paimon/common/utils/math.h` and `src/paimon/common/utils/uuid.h` under RocksDB. This PR also introduces RocksDB-derived assertion utilities in `src/paimon/testing/utils/testharness.cpp` and `src/paimon/testing/utils/testharness.h`, and those paths should be added to the existing RocksDB section to keep the attribution list accurate.

After that update, the migration looks good to me.

leaves12138 (approved):
Thanks for the update. The Apache Arrow attribution is now present, and the existing RocksDB / LevelDB section now includes the `testharness.cpp` / `testharness.h` files introduced by this PR. The migrated files are consistent with the staged source migration, and the lightweight public-header syntax check passes. LGTM.

PR #3, opened 2026-05-22
chore: add apidoc doc-config files

lucasfang:
### Purpose

This PR adds two new documentation-related files:

Doxyfile
.gitignore

### DIFF
--- a/apidoc/.gitignore
+++ b/apidoc/.gitignore
@@ -1,0 +1,19 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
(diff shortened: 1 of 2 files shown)


leaves12138 (approved):
LGTM. I checked the added apidoc files against the migrated source snapshot: the content is effectively the existing Doxygen configuration with ASF headers, `apidoc/.gitignore` covers the generated `html/` and `xml/` directories, and the file formatting/newline checks look good.

Non-blocking note: until `include/` is migrated, running Doxygen will warn that `../include` does not exist. The config is intended to be run from the `apidoc/` directory, consistent with the existing source workflow.

lxy-9602:
LGTM

PR #6, opened 2026-05-22
chore: add thirdparty toolchain and thirdparty diff

lucasfang:
### Purpose

add thirdparty toolchain and thirdparty diff patch

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -265,3 +265,34 @@ Copyright: 2016-2024 The Apache Software Foundation.
 Home page: https://arrow.apache.org/
 License: https://www.apache.org/licenses/LICENSE-2.0
 --------------------------------------------------------------------------------
+
+This product includes code from cppjieba.
+
+* cppjieba utility in src/paimon/global_index/lucene/ directory
+
+Copyright: 2013
+Home page: https://github.com/yanyiwu/cppjieba
+License: https://opensource.org/licenses/MIT
+
+The MIT License (MIT)
+
+Copyright (c) 2013
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the "Software"), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+the Software, and to permit persons to whom the Software is furnished to do so,
+subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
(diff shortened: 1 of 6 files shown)


leaves12138 (changes requested):
Thanks for migrating the third-party toolchain and patch files. I checked the imported patch files against the source snapshot, and the diff files themselves match the migration source.

I found two blockers before merge:

1. `cmake_modules/ThirdpartyToolchain.cmake` is not a pure migration of the current source snapshot. It changes the Arrow/ORC dependency policy to allow `SYSTEM`/`AUTO` to resolve to system Arrow/ORC and removes the guard that forced bundled Arrow/ORC because paimon-cpp applies project-specific patches. Since this PR also imports `arrow.diff` and `orc.diff`, using unpatched system Arrow/ORC can silently bypass those required patches. Please either revert this toolchain file to the source snapshot behavior for this migration PR, or split and justify the system-Arrow/ORC policy change with tests in a separate PR.

2. This PR introduces third-party patch files for Arrow, ORC, and cppjieba, but the Apache repository's `LICENSE`/`NOTICE` files are not updated with the corresponding third-party attribution/license entries. Please add the relevant Apache Arrow, Apache ORC, and cppjieba entries when introducing these files.

The rest of the staged migration shape looks fine.

leaves12138:
I re-checked this after the clarification.

`DefineOptions.cmake` does block explicit `-DArrow_SOURCE=SYSTEM` and `-DORC_SOURCE=SYSTEM`, but it does not fully prevent system Arrow/ORC in the current PR:

* if `PAIMON_DEPENDENCY_SOURCE=SYSTEM` is used and `Arrow_SOURCE` / `ORC_SOURCE` are left empty, `paimon_get_dependency_source()` falls back to `PAIMON_DEPENDENCY_SOURCE`, so Arrow/ORC can still resolve as `SYSTEM`;
* if `PAIMON_DEPENDENCY_SOURCE=AUTO` (the default) and system Arrow/ORC are found, this PR's `paimon_configure_dependency_root()` keeps `AUTO`, then `resolve_dependency()` can choose `ArrowAlt` / `ORCAlt` as system dependencies.

The source snapshot had an additional guard in `ThirdpartyToolchain.cmake` that forced Arrow/ORC `AUTO` to `BUNDLED` and rejected Arrow/ORC `SYSTEM`, because these dependencies need project-specific patches. That guard is removed in this PR, so the blocker still stands.

The LICENSE/NOTICE attribution blocker also still stands.

leaves12138 (changes requested):
Re-reviewed the latest update. The Arrow/ORC dependency policy blocker is fixed, and the cppjieba attribution was added.

One licensing blocker remains: the current LICENSE/NOTICE update still does not cover the newly added Apache ORC-derived patch file, and the Apache Arrow section still does not list the newly added Arrow-derived files (`cmake_modules/ThirdpartyToolchain.cmake` and `cmake_modules/arrow.diff`). Please update LICENSE/NOTICE to accurately cover these newly distributed files before merge.

leaves12138 (approved):
Re-reviewed the latest update. The patched Arrow/ORC dependency policy is enforced, and LICENSE/NOTICE now cover the added Arrow/ORC/cppjieba files. I do not see further blockers.

PR #10, opened 2026-05-25
feat: introduce basic data types including decimal, blob and timestamp

lszskye:
### Purpose
Introduce the basic data types module including:

- **Decimal**: A 128-bit decimal type compatible with Java Paimon's `DecimalData`. 
- **Blob**: A binary large object type that supports reading from file paths or blob descriptors. 
- **Timestamp**: An immutable timestamp type representing milliseconds + nanos-of-millisecond since epoch.
- **Logger**: A pluggable logging framework with a default glog-based adapter, supporting custom logger registration and thread-safe initialization.

Internal utilities added:
- `BlobDescriptor`: Serializable descriptor for blob references with versioned binary format (v1/v2 compatible with Java).
- `BlobDefs`: Constants for blob file format (magic number, header layout, etc.).
- `BlobUtils`: Helper functions for separating blob/non-blob fields in Arrow schemas and arrays.

### Tests
- `BlobTest`
- `BlobDescriptorTest`
- `BlobUtilsTest`
- `DecimalTest`
- `TimestampTest`
- `LoggerTest`

### API and Format

This change adds new public headers under `include/paimon/data/`:
- `blob.h` — `Blob` class
- `decimal.h` — `Decimal` class
- `timestamp.h` — `Timestamp` class

And extends `include/paimon/logging.h` with the `Logger` interface.

No existing API or storage format is changed.

### DIFF
--- a/include/paimon/data/blob.h
+++ b/include/paimon/data/blob.h
@@ -1,0 +1,113 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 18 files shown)


leaves12138 (approved):
Thanks for migrating the basic data types and logging utilities. I compared the added files with the source snapshot; they match the staged migration with ASF header normalization. The PR is based on the current main and merges cleanly. I also checked line endings/trailing whitespace and ran a lightweight public-header syntax smoke check; no blocker found.

Non-blocking note: some implementation/test files still depend on components expected to arrive in later migration PRs, such as filesystem, IO streams, string/decimal utilities, and executor futures. That looks fine for the staged migration.

LGTM.

PR #11, opened 2026-05-25
feat: introduce common/memory module

lxy-9602:
### Purpose

No Linked issue.

### Introduce

- `MemorySegment` — core memory abstraction wrapping a `Bytes` buffer with typed get/put operations, bounds checking, and bulk copy/compare support (`memory_segment.h/cpp`)
- `MemorySegmentUtils` — utility functions for cross-segment operations including multi-segment copy, compare, equality check, and hash (`memory_segment_utils.h/cpp`)

### Tests

- `memory_segment_test.cpp` — MemorySegment typed read/write, boundary checks, bulk operations
- `memory_segment_utils_test.cpp` — Multi-segment copy, compare, hash correctness

### DIFF
--- a/include/paimon/io/byte_order.h
+++ b/include/paimon/io/byte_order.h
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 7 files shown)


leaves12138 (changes requested):
Thanks for the contribution. I found one correctness issue that should be fixed before merge.

`MemorySegment::Compare` compares each 8-byte chunk as a native-endian integer. On little-endian platforms this does not preserve lexicographical byte order. For example, `[0x01, 0, 0, 0, 0, 0, 0, 0]` should compare greater than `[0x00, 0x01, 0, 0, 0, 0, 0, 0]` byte-wise, but the current implementation reads them as `1` and `256` and returns the opposite result.

Please either compare byte-by-byte, or use an endian-safe/big-endian 8-byte comparison, and add a regression test covering this case.

leaves12138 (approved):
Re-reviewed the latest update. The previous MemorySegment::Compare correctness blocker is fixed: the 8-byte fast path now compares big-endian values, so lexicographical byte order is preserved on little-endian platforms. I do not see further blockers.

PR #13, opened 2026-05-25
feat: introduce data type with JSON serialization

lszskye:
### Purpose
Introduce the Paimon data type system for apache-paimon-cpp, including:

- **DataType**: Base class representing Paimon data types, with Arrow type mapping and JSON serialization. Supports atomic types (BOOLEAN, TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE, STRING, BYTES, DATE, DECIMAL, TIMESTAMP, BLOB) and complex types (ARRAY, MAP, ROW).
- **DataField**: Represents a field in a row type with id, name, type, and optional description. Provides bidirectional conversion between Paimon DataField and Arrow Field/Schema, including metadata handling for field ids.
- **RowType**: A structured type containing named fields, serialized as JSON with field definitions.
- **ArrayType**: A list/array type with element type.
- **MapType**: A key-value map type. Note: keys are always non-nullable due to Apache Arrow limitations.
- **RowKind**: Represents row kinds (INSERT, UPDATE_BEFORE, UPDATE_AFTER, DELETE) with byte value and short string serialization.
- **DataTypeJsonParser**: A comprehensive type parser supporting the full Paimon type syntax (including parameterized types like `DECIMAL(10,2)`, `TIMESTAMP(9) WITH LOCAL TIME ZONE`, etc.).

### Tests

- `DataTypeTest`
- `DataFieldTest`
- `DataTypeJsonParserTest`
- `RowKindTest`

### DIFF
--- a/src/paimon/common/types/array_type.h
+++ b/src/paimon/common/types/array_type.h
@@ -1,0 +1,57 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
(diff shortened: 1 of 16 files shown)


leaves12138 (changes requested):
Thanks for the contribution. I found one parser correctness issue that should be fixed before merge.

`Tokenize` is missing a `break` after handling `CHAR_END_SUBTYPE` (`>`). This makes every `>` token also emit a `BEGIN_PARAMETER` (`(`) token at the same position. As a result, SQL-style nested type strings using subtype syntax, such as `ROW<id INT>` or `ARRAY<INT>`, cannot be parsed correctly because the token stream contains an extra `(` after each `>`.

Please add the missing `break` and include a regression test for a type string containing `>`.

leaves12138 (approved):
Re-reviewed the latest update. The previously reported `CHAR_END_SUBTYPE` fallthrough has been fixed. I did not find further blockers in this round.

PR #18, opened 2026-05-26
feat: add common data structures (BinarySection, BinaryString, InternalRow)

lxy-9602:
### Purpose

No Linked issue.

### Introduce

- `BinarySection` — base class for binary data backed by a MemorySegment, providing offset/size-based access (`binary_section.h/cpp`)
- `BinaryString` — variable-length binary string with comparison, copy, hash, and substring operations (`binary_string.h/cpp`)
- `DataDefine` — variant type definitions (`NullableDataField`, `DataField`) for Paimon columnar data (`data_define.h`)
- `DataGetters` / `DataSetters` — abstract interfaces for typed columnar read/write access (`data_getters.h`, `data_setters.h`)
- `InternalRow` — abstract row interface with typed field accessors and field count (`internal_row.h/cpp`)
- `InternalArray` — abstract array interface for columnar array data (`internal_array.h`)
- `InternalMap` — abstract map interface wrapping key/value InternalArrays (`internal_map.h`)

### Tests
- `binary_section_test.cpp` — BinarySection construction and memory access
- `binary_string_test.cpp` — BinaryString comparison, copy, hash, substring operations
- `internal_row_test.cpp` — InternalRow field access and JoinedRow correctness

### DIFF
--- a/src/paimon/common/data/binary_section.cpp
+++ b/src/paimon/common/data/binary_section.cpp
@@ -1,0 +1,64 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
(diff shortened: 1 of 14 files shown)


leaves12138 (changes requested):
Thanks for the contribution. I found one blocker that should be fixed before merge.

`BinarySection::HIGHEST_FIRST_BIT` is defined as `0x80L << 56`. In C++, this left-shifts a signed integer into the sign bit, which is undefined behavior when the result is not representable as a signed `long`. This mask is used by `ReadBinary` to distinguish inline fixed-length data from variable-length data, so compiler-dependent behavior here can break binary/string decoding.

Please define these masks using unsigned arithmetic and cast after the shift, for example `static_cast<int64_t>(0x80ULL << 56)` (and preferably use unsigned masks consistently for the related `0x7F` mask as well).

leaves12138 (approved):
Thanks for the update. The previous signed-shift issue is fixed, and I did not find any remaining blocking issue in this round.

PR #19, opened 2026-05-26
feat: add Range, RangeHelper, RowRangeIndex, and ByteRangeCombiner ut…

dalingmeng:
### Purpose

No Linked issue.

Introduce range-related utility modules to the apache/paimon-cpp codebase. These modules provide range representation, range manipulation helpers, row range indexing with binary search, and byte range combining adapted from Apache ORC.

- Introduce `Range` struct for inclusive integer range representation (`include/paimon/utils/range.h`, `range.cpp`)
- Introduce `RangeHelper` template for range splitting and merging operations (`range_helper.h`)
- Introduce `RowRangeIndex` for efficient intersection queries over sorted non-overlapping ranges using binary search (`include/paimon/utils/row_range_index.h`, `row_range_index.cpp`)
- Introduce `ByteRangeCombiner` for combining byte ranges, adapted from Apache ORC (`byte_range_combiner.h`, `byte_range_combiner.cpp`)
- Update `LICENSE` to add `byte_range_combiner` files under the Apache ORC attribution entry

### Tests

- `range_test.cpp` — Range construction, count, intersection, merge, and string representation
- `range_helper_test.cpp` — RangeHelper split and merge operations
- `row_range_index_test.cpp` — RowRangeIndex intersection queries and edge cases
- `byte_range_combiner_test.cpp` — ByteRangeCombiner combining logic

### API and Format

No API or format changes.

### Documentation

No documentation changes.

### DIFF
--- a/include/paimon/utils/range.h
+++ b/include/paimon/utils/range.h
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "paimon/visibility.h"
+
+namespace paimon {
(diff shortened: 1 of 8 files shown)


src/paimon/common/utils/byte_range_combiner.h
```diff
@@ -0,0 +1,37 @@
...
+
+// Adapted from Apache ORC
+// https://github.com/apache/orc/blob/main/c%2B%2B/src/io/Cache.hh
+
+#pragma once
+
+#include <vector>
+
+#include "paimon/result.h"
+#include "paimon/utils/read_ahead_cache.h"
```
leaves12138: This header is not present in the PR head tree ( /  are both missing), but  uses  from it. As-is, this new header cannot compile. Please add the missing header/ definition, or include the correct existing header before using .

leaves12138 (changes requested):
Thanks for the update. I found one compile-blocking issue:  depends on  via , but that header is not present in this PR/head tree, so including  cannot compile. Please either add the missing / definition in this PR, include the correct existing header, or move  into a header that is actually part of the branch before using it here.

leaves12138 (approved):
Thanks for the update. The previous missing-header blocker is resolved after removing ByteRangeCombiner from this PR, and I did not find any remaining blocking issue in the updated range/row-range changes.

PR #22, opened 2026-05-26
feat: introduce binary row format with reader/writer support

lszskye:
### Purpose
Introduce the binary row data format, compatible with Java Paimon's `BinaryRow` serialization format. This module provides efficient in-memory row representation with fixed-length and variable-length parts within a single `MemorySegment`.

Key components:
- **BinaryWriter**: Abstract interface defining the write protocol.
- **AbstractBinaryWriter**: Template-method base class implementing variable-length field writing (strings, bytes, decimals, timestamps, arrays, rows, maps) with auto-growing buffer management.
- **BinaryRowWriter**: Concrete writer for `BinaryRow`, supporting all Paimon data types.
- **BinaryRow**: Immutable binary row representation backed by `MemorySegment`. Implements `InternalRow` for reading all field types, supports null tracking via bit set, hashing, equality comparison, and copy operations.
- **BinaryDataReadUtils**: Static utilities for reading typed fields (Timestamp, Decimal, BinaryString, Array, Row, Map) from memory segments with proper offset/size decoding.
- **BinaryArray**: A binary implementation of `InternalArray` backed by a single `MemorySegment`. Supports all Paimon data types. Provides bulk conversion methods (`ToIntArray`, `ToLongArray`, etc.) and factory methods (`FromIntArray`, `FromLongArray`).
- **BinaryArrayWriter**: Writer for `BinaryArray`, inheriting from `AbstractBinaryWriter`. 
- **BinaryMap**: A binary implementation of `InternalMap` backed by a single `MemorySegment`.

### Tests
- `BinaryRowTest`
- `BinaryRowWriterTest`
- `BinaryArrayTest`
- `BinaryMapTest`

### DIFF
--- a/src/paimon/common/data/abstract_binary_writer.cpp
+++ b/src/paimon/common/data/abstract_binary_writer.cpp
@@ -199,24 +199,24 @@ int32_t AbstractBinaryWriter::RoundNumberOfBytesToNearestWord(int32_t num_bytes)
         return num_bytes + (8 - remainder);
     }
 }
+
 template <typename T>
 void AbstractBinaryWriter::WriteBytesToFixLenPart(MemorySegment* segment, int32_t field_offset,
                                                   const T& bytes, int32_t len) {
-    int64_t first_byte = len | 0x80;  // first bit is 1, other bits is len
-    int64_t seven_bytes = 0L;         // real data
+    const uint64_t first_byte =
+        static_cast<uint64_t>(len) | 0x80U;  // first bit is 1, low 7 bits are len
+    uint64_t seven_bytes = 0U;               // real data
     if ((SystemByteOrder() == ByteOrder::PAIMON_LITTLE_ENDIAN)) {
         for (int32_t i = 0; i < len; i++) {
-            seven_bytes |= ((0x00000000000000FFL & bytes[i]) << (i * 8L));
+            seven_bytes |= (static_cast<uint64_t>(bytes[i]) & 0xFFU) << (i * 8U);
         }
     } else {
         for (int32_t i = 0; i < len; i++) {
-            seven_bytes |= ((0x00000000000000FFL & bytes[i]) << ((6 - i) * 8L));
+            seven_bytes |= (static_cast<uint64_t>(bytes[i]) & 0xFFU) << ((6 - i) * 8U);
         }
     }
-    const int64_t offset_and_size =
-        (first_byte << 56) |  // NOLINT(clang-analyzer-core.UndefinedBinaryOperatorResult)
-        seven_bytes;
-    segment->PutValue<int64_t>(field_offset, offset_and_size);
+    const uint64_t offset_and_size = (first_byte << 56) | seven_bytes;
+    segment->PutValue<int64_t>(field_offset, static_cast<int64_t>(offset_and_size));
 }
 
 }  // namespace paimon
--- a/src/paimon/common/data/abstract_binary_writer.cpp
+++ b/src/paimon/common/data/abstract_binary_writer.cpp
@@ -1,0 +1,222 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "paimon/common/data/abstract_binary_writer.h"
+
+#include <cassert>
+#include <memory>
+#include <optional>
+
+#include "paimon/common/data/binary_array.h"
+#include "paimon/common/data/binary_map.h"
+#include "paimon/common/data/binary_row.h"
+#include "paimon/common/data/binary_section.h"
+#include "paimon/common/data/binary_string.h"
+#include "paimon/common/memory/memory_segment_utils.h"
+#include "paimon/data/decimal.h"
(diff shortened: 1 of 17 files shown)


leaves12138 (changes requested):
Thanks for the binary row implementation. This PR currently does not compile against its base branch because it includes and uses `BinaryArray`/`BinaryMap`, but `src/paimon/common/data/binary_array.h` and `src/paimon/common/data/binary_map.h` are not present in the PR or in `main`. Please either add those dependencies in this PR, rebase after the PR that introduces them is merged, or remove the nested array/map support from this change.

src/paimon/common/data/abstract_binary_writer.cpp
```diff
@@ -0,0 +1,222 @@
...
+        for (int32_t i = 0; i < len; i++) {
+            seven_bytes |= ((0x00000000000000FFL & bytes[i]) << (i * 8L));
+        }
+    } else {
+        for (int32_t i = 0; i < len; i++) {
+            seven_bytes |= ((0x00000000000000FFL & bytes[i]) << ((6 - i) * 8L));
+        }
+    }
+    const int64_t offset_and_size =
+        (first_byte << 56) |  // NOLINT(clang-analyzer-core.UndefinedBinaryOperatorResult)
```
leaves12138: This inline-string/binary encoding still has undefined behavior. `first_byte` is an `int64_t`; for inline values `len | 0x80` sets bit 7, so shifting it left by 56 tries to produce the signed sign bit value (`0x80 << 56`), which is outside the representable range of `int64_t`. Please build this field with `uint64_t` (similar to `BinarySection::HIGHEST_FIRST_BIT`) and cast back only when storing it.

leaves12138 (changes requested):
Thanks for the update. The previous missing BinaryArray/BinaryMap issue is fixed, but I found one remaining correctness blocker in the inline bytes writer path.

src/paimon/common/data/binary_row.cpp
```diff
@@ -0,0 +1,311 @@
...
+#include "paimon/common/data/binary_data_read_utils.h"
+#include "paimon/common/memory/memory_segment.h"
+#include "paimon/common/memory/memory_segment_utils.h"
+#include "paimon/io/byte_order.h"
+#include "paimon/memory/bytes.h"
+#include "paimon/memory/memory_pool.h"
+
+namespace paimon {
+const int64_t BinaryRow::FIRST_BYTE_ZERO =
+    (SystemByteOrder() == ByteOrder::PAIMON_LITTLE_ENDIAN) ? (~0xFFL) : (~(0xFFL << 56L));
```
leaves12138: This still has undefined behavior on the big-endian branch:  is signed, and  cannot be represented as a signed . Please build this mask with an unsigned value first, e.g. , or keep the mask as  until the final cast.

leaves12138 (changes requested):
Thanks for the update. The previous inline bytes writer issue is fixed, but there is still one signed left-shift overflow issue in the newly added BinaryRow code.

src/paimon/common/data/binary_row_writer.cpp
```diff
@@ -0,0 +1,181 @@
...
+                        writer->SetNullAt(field_idx);
+                    }
+                    return;
+                }
+                return writer->WriteTimestamp(
+                    field_idx, DataDefine::GetVariantValue<Timestamp>(field), precision);
+            };
+            return field_setter;
+        }
+        case arrow::Type::type::DECIMAL: {
```
leaves12138, line 146: This branch still will not handle `arrow::decimal128(...)`. In Arrow C++, decimal128 reports `arrow::Type::DECIMAL128` (the rest of this codebase also switches on `DECIMAL128`, e.g. `DataDefine::VariantValueToLiteral` and `BinaryArrayWriter::GetElementSize` tests). As written, `CreateFieldSetter(arrow::decimal128(...))` falls through to the unsupported-type error, so the decimal cases added in `binary_row_writer_test.cpp` cannot work in a real build. Please switch this to `DECIMAL128` (and add DECIMAL256 handling only if intended).

src/paimon/common/data/binary_row_writer.cpp
```diff
@@ -0,0 +1,181 @@
...
+            auto* decimal_type =
+                arrow::internal::checked_cast<arrow::Decimal128Type*>(field_type.get());
+            assert(decimal_type);
+            auto precision = decimal_type->precision();
+            field_setter = [field_idx, precision](const VariantType& field,
```
leaves12138: For non-compact decimals, the setter captures only precision and then calls `WriteDecimal` with a `Decimal` value whose scale may differ from the Arrow field scale. `WriteDecimal` currently only asserts the precision and never checks or normalizes scale, but the reader reconstructs decimals with the schema scale. That means a value like `Decimal(5, 3, 123)` written to a `decimal128(5, 2)` field will be read back as `1.23` instead of the original `0.123` (or being rejected). Please capture the Arrow scale here and either validate `value.Scale() == scale` or rescale/reject before writing, so the stored unscaled bytes match the schema scale.

leaves12138 (changes requested):
Thanks for the update. I re-reviewed the current PR head end-to-end and found two remaining blockers that should be fixed before merging.

leaves12138:
Re-reviewed the current head (`94b3929`). The two blocking issues I found are already listed in my latest review comments: the decimal setter switches on the wrong Arrow decimal enum, and non-compact decimal writes do not validate/normalize scale. I did not find additional blockers in this pass.

src/paimon/common/data/binary_row_writer.cpp
leaves12138: Correction: this comment was inaccurate. Arrow C++ defines `arrow::Type::DECIMAL` as a compatibility alias of `arrow::Type::DECIMAL128`, so `arrow::decimal128(...)` is handled by this `case arrow::Type::DECIMAL` branch. Please ignore this comment; no enum change is needed here.

leaves12138 (approved):
I re-reviewed the latest head (329611d1cc4ef1556347fcd8d23345860e1aaf2e). The previous inline encoding signed-shift issues have been addressed, and my earlier comment about Arrow DECIMAL vs DECIMAL128 was incorrect and has been corrected separately. I did not find any remaining blocking correctness issues in the current binary row/array reader and writer changes.\n\nI also ran a lightweight whitespace check with git diff --check successfully. I could not run a full local build in this environment because the checkout does not include usable CMake build files and Arrow headers are not installed here.

PR #24, opened 2026-05-26
feat: migrate common/data/columnar module

lxy-9602:
### Purpose

No Linked issue.

Migrate `common/data/columnar/` module. This module provides columnar data access layer built on top of Apache Arrow arrays, including:

- **ColumnarArray**: Access vector column data for array types
- **ColumnarMap**: Access vector column data for map types  
- **ColumnarRow**: Row view over arrow::Array with per-row field access
- **ColumnarRowRef**: Lightweight row view sharing batch-level context to reduce per-row overhead
- **ColumnarUtils**: Utility functions for reading Arrow arrays (including dictionary-encoded arrays)
- **ColumnarBatchContext**: Shared batch context struct for ColumnarRowRef

Also migrates the dependency `DateTimeUtils` (timestamp conversion utilities).

### Tests

- `columnar_utils_test.cpp` — Tests GetView/GetBytes for plain and dictionary-encoded string arrays
- `columnar_array_test.cpp` — Tests all primitive types, complex/nested types (struct, list, map, decimal, timestamp), and null handling
- `columnar_row_test.cpp` — Tests ColumnarRow and ColumnarRowRef for simple types, complex/nested types, dictionary types, null handling, data lifecycle, and binary access

### DIFF
--- a/src/paimon/common/data/columnar/columnar_array.cpp
+++ b/src/paimon/common/data/columnar/columnar_array.cpp
@@ -47,8 +47,11 @@ Decimal ColumnarArray::GetDecimal(int32_t pos, int32_t precision, int32_t scale)
     auto array = arrow::internal::checked_cast<const ArrayType*>(array_);
     assert(array);
     arrow::Decimal128 decimal(array->GetValue(offset_ + pos));
-    return Decimal(precision, scale,
-                   static_cast<Decimal::int128_t>(decimal.high_bits()) << 64 | decimal.low_bits());
+    return Decimal(
+        precision, scale,
+        static_cast<Decimal::int128_t>(
+            static_cast<Decimal::uint128_t>(static_cast<uint64_t>(decimal.high_bits())) << 64 |
+            decimal.low_bits()));
 }
 
 Timestamp ColumnarArray::GetTimestamp(int32_t pos, int32_t precision) const {
--- a/src/paimon/common/data/columnar/columnar_array.cpp
+++ b/src/paimon/common/data/columnar/columnar_array.cpp
@@ -71,7 +71,7 @@ std::shared_ptr<InternalArray> ColumnarArray::GetArray(int32_t pos) const {
     assert(list_array);
     int32_t offset = list_array->value_offset(offset_ + pos);
     int32_t length = list_array->value_length(offset_ + pos);
-    return std::make_shared<ColumnarArray>(list_array->values(), pool_, offset, length);
+    return std::make_shared<ColumnarArray>(list_array->values().get(), pool_, offset, length);
 }
 
 std::shared_ptr<InternalMap> ColumnarArray::GetMap(int32_t pos) const {
--- a/src/paimon/common/data/columnar/columnar_array.cpp
+++ b/src/paimon/common/data/columnar/columnar_array.cpp
@@ -1,0 +1,157 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include "paimon/common/data/columnar/columnar_array.h"
+
+#include <utility>
+
+#include "arrow/api.h"
+#include "arrow/array/array_decimal.h"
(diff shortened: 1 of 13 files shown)


src/paimon/common/data/columnar/columnar_row.cpp
```diff
@@ -0,0 +1,81 @@
...
+    DateTimeUtils::TimeType time_type = DateTimeUtils::GetTimeTypeFromArrowType(timestamp_type);
+    auto [milli, nano] = DateTimeUtils::TimestampConverter(
+        data, time_type, DateTimeUtils::TimeType::MILLISECOND, DateTimeUtils::TimeType::NANOSECOND);
+    return Timestamp(milli, nano);
+}
+
+std::shared_ptr<InternalRow> ColumnarRow::GetRow(int32_t pos, int32_t num_fields) const {
+    auto struct_array = arrow::internal::checked_cast<const arrow::StructArray*>(array_vec_[pos]);
+    assert(struct_array);
+    return std::make_shared<ColumnarRow>(struct_array->fields(), pool_, row_id_);
```
leaves12138, line 69: This nested row still drops the ownership chain. The returned ColumnarRow stores raw child Array pointers, but it is constructed without a holder. If the parent ColumnarRow is the only object owning the top-level StructArray, a nested row returned from GetRow can outlive the parent and then its raw child pointers dangle. The current DataLifeCycle test only uses result_row while the parent row is alive, so it does not catch this. Please keep an owning holder here, for example by passing a shared_ptr/sliced StructArray or by returning ColumnarRowRef with a ColumnarBatchContext that owns struct_array->fields().

src/paimon/common/data/columnar/columnar_array.cpp
```diff
@@ -0,0 +1,157 @@
...
+    return Status::OK();
+}
+
+Decimal ColumnarArray::GetDecimal(int32_t pos, int32_t precision, int32_t scale) const {
+    using ArrayType = typename arrow::TypeTraits<arrow::Decimal128Type>::ArrayType;
+    auto array = arrow::internal::checked_cast<const ArrayType*>(array_);
+    assert(array);
+    arrow::Decimal128 decimal(array->GetValue(offset_ + pos));
+    return Decimal(precision, scale,
+                   static_cast<Decimal::int128_t>(decimal.high_bits()) << 64 | decimal.low_bits());
```
leaves12138: Please avoid reconstructing the signed int128 value by left-shifting a signed high word. For negative Decimal128 values, decimal.high_bits() is negative, and left-shifting a negative signed integer is undefined behavior in C++. This same expression is also used in ColumnarRow and ColumnarRowRef. Please combine the bits through an unsigned 128-bit value first, then cast to Decimal::int128_t, and add a negative decimal test.

leaves12138 (changes requested):
Thanks for the update. The non-owning ColumnarArray constructor is clearer now, but I found two remaining correctness blockers in nested row lifetime and decimal conversion.

src/paimon/common/data/columnar/columnar_utils.h
```diff
@@ -0,0 +1,106 @@
...
+            const auto* typed_array =
+                arrow::internal::checked_cast<const arrow::DictionaryArray*>(array);
+            assert(typed_array);
+            auto dict_type =
+                arrow::internal::checked_pointer_cast<arrow::DictionaryType>(array->type());
+            assert(dict_type);
+            auto value_type_id = dict_type->value_type()->id();
+            auto index_type_id = dict_type->index_type()->id();
+            int64_t dict_index = -1;
+            if (index_type_id == arrow::Type::type::INT32) {
```
leaves12138: This only decodes dictionary indices for `INT32` and `INT64`, but Arrow dictionary indices are not limited to these two widths (for example `int8`/`int16` dictionaries are valid and commonly used for small dictionaries). For those arrays `dict_index` stays `-1`; in release builds the assert is compiled out and the code then calls `dictionary->GetView(-1)`, which can read before the offsets buffer or crash. Please handle all supported integer index types (or return a proper error instead of relying on `assert`).

leaves12138 (changes requested):
I did another full pass and found one additional correctness issue beyond the two existing blockers.

leaves12138 (dismissed):
Thanks for the update. The decimal conversion and dictionary index-width issues look fixed now, but the nested row lifetime blocker is still present in the latest commit. In src/paimon/common/data/columnar/columnar_row.cpp, ColumnarRow::GetRow still returns a nested ColumnarRow built from struct_array->fields() without an owning holder. The new comment documents the caller-side lifetime requirement, but it does not fix the unsafe case introduced by the owning ColumnarRow constructor: if the parent row is the only object keeping the top-level StructArray alive, a nested row returned here can outlive that parent and then its raw child-array pointers dangle. The current TestDataLifeCycle still accesses result_row only while row is alive, so it does not exercise the failure. Please keep the ownership chain for escaped nested rows, e.g. return a row/ref that owns the child arrays via a shared context, or otherwise holds a shared owner/sliced struct array.

leaves12138 (approved):
Thanks for clarifying the intended ownership contract. Given that ColumnarRow::GetRow is intentionally a non-owning hot-path view and the caller is required to keep the parent data alive, the remaining lifetime point should not block this PR. The previous decimal conversion and dictionary index-width issues are fixed in the latest commit. Looks good to me.

PR #25, opened 2026-05-27
chore: add fix_includes.py for include cleanup workflow

lucasfang:
### Purpose

Linked issue: close #TODO

Introduce `build_support/fix_includes.py` to support include cleanup and ordering workflow in build tooling.
This change adds one utility script and does not modify runtime code paths.

### Tests

1. python3 -m py_compile build_support/fix_includes.py (pass)
2. Verified migrated file line count: 2498
3. Confirmed file is tracked as a single new addition in this change set

### API and Format

No impact on public API under include/ and no storage/protocol format change.

### Documentation

No end-user feature introduced. This is a build/developer tooling migration.

### Generative AI tooling

Generated-by: GitHub Copilot (GPT-5.3-Codex)

### DIFF
--- a/build_support/fix_includes.py
+++ b/build_support/fix_includes.py
@@ -1,0 +1,2498 @@
+#!/usr/bin/env python3
+
+##===--- fix_includes.py - rewrite source files based on iwyu output ------===##
+#
+#                     The LLVM Compiler Infrastructure
+#
+# This file is distributed under the University of Illinois Open Source
+# License. See LICENSE.TXT for details.
+#
+##===----------------------------------------------------------------------===##
+
+from __future__ import print_function
+
+"""Update files with the 'correct' #include and forward-declare lines.
+
+Given the output of include_what_you_use on stdin -- when run at the
+(default) --v=1 verbosity level or higher -- modify the files
+mentioned in the output, removing their old #include lines and
+replacing them with the lines given by the include_what_you_use
+script.
+
+This script runs in four stages.  In the first, it groups physical
+lines together to form 'move spans'.  A 'move span' is the atomic unit
+for moving or deleting code.  A move span is either a) an #include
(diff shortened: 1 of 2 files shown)


build_support/fix_includes.py
```diff
@@ -0,0 +1,2498 @@
...
+    flags.comments = True
+
+  if flags.sort_only:
+    if not files_to_modify:
+      sys.exit('FATAL ERROR: -s flag requires a list of filenames')
+    SortIncludesInFiles(files_to_modify, flags)
+  else:
+    ProcessIWYUOutput(sys.stdin, files_to_modify, flags, cwd=os.getcwd())
+
+  return 0
```
leaves12138, line 2494: This currently always returns 0, even in `--dry_run` mode when files would be modified. The upstream `fix_includes.py` returns `min(process_ret, 100)` for dry runs, and the help text above also documents that behavior. With this change, CI/pre-commit workflows cannot use `fix_includes.py --dry_run` to fail when includes need cleanup, so include drift would be silently accepted. Please keep the upstream dry-run exit-code behavior (or update the workflow/docs if a different contract is intended).

leaves12138 (changes requested):
Thanks for adding the include-cleanup helper. I found one workflow blocker compared with the upstream IWYU script behavior.

leaves12138 (approved):
Acknowledged that this file is intended to remain a verbatim IWYU copy. Please disregard my previous dry-run exit-code comment as a blocker; preserving the upstream script unchanged is reasonable. I have no further blocking comments.

leaves12138 (approved):
Re-reviewed the latest update. The LICENSE now lists build_support/fix_includes.py under the include-what-you-use / UIUC-NCSA notice, and preserving the imported script unchanged is fine. I have no further comments.

PR #28, opened 2026-05-28
feat: add GenericRow, RecordBatch, MemorySize and TimeDuration

lszskye:
### Purpose

GenericRow: A generic implementation of InternalRow backed by an array of VariantType, supporting all Paimon data types.

RecordBatch: Encapsulates a batch of columnar data with partition info, bucket id, and per-row RowKind (INSERT / UPDATE_BEFORE / UPDATE_AFTER / DELETE).

MemorySize: Parses human-readable memory size strings (e.g., "1024mb", "2g") into byte values.
Supports units: bytes, kibibytes, mebibytes, gibibytes, tebibytes, with case-insensitive matching.

TimeDuration: Parses human-readable time duration strings (e.g., "500ms", "10s", "2h") into millisecond values.
Supports units: nanoseconds, microseconds, milliseconds, seconds, minutes, hours, days.

### Tests
GenericRowTest
RecordBatchTest
MemorySizeTest
TimeDurationTest

### DIFF
--- a/include/paimon/record_batch.h
+++ b/include/paimon/record_batch.h
@@ -1,0 +1,144 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "paimon/result.h"
+#include "paimon/visibility.h"
+
+struct ArrowArray;
+
+namespace paimon {
+/// `RecordBatch` encapsulates a batch of data with the same schema, supporting different types such
+/// as `INSERT`, `UPDATE_BEFORE`, `UPDATE_AFTER`, and `DELETE`. It is typically used in streaming
+/// write or batch processing scenarios, with underlying data stored in the Apache Arrow format.
+/// @note Do not use this class directly, use `RecordBatchBuilder` to build a `RecordBatch` which
+/// has input validation.
+class PAIMON_EXPORT RecordBatch {
+ public:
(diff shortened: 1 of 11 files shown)


leaves12138 (dismissed):
Thanks for the PR. I found several issues that should be fixed before this can be merged.

1. `RecordBatchBuilder::Finish()` leaves the builder holding a moved-from `ArrowArray` pointer until the scope guard runs.
   The `RecordBatch` constructor calls `ArrowArrayMove(impl_->data_, data_)`, which releases the source ArrowArray. After returning from `std::make_unique<RecordBatch>(...)`, the scope guard calls `impl_->Reset()`, which calls `ArrowArrayRelease(impl_->data_)` again. This can double-release or at least invoke the release path on a moved-from C Data Interface struct. The builder should transfer ownership by clearing `impl_->data_` after the move, or construct the `RecordBatch` in a way that consumes and nulls the pointer before `Reset()` runs.

2. When `row_kinds_` is not explicitly set, `Finish()` stores an empty vector in the `RecordBatch`, even though the public API says the default row kind is `INSERT`. Downstream code calling `GetRowKind()` will see no row kinds at all and cannot distinguish this from a zero-row batch. Please materialize `data->length` `INSERT` entries in `Finish()` when `row_kinds_` is empty.

3. The bucket default is inconsistent with the public contract. The header says that if the bucket is not set, the default value is `-1`, but the implementation uses `std::numeric_limits<int32_t>::min()` as the stored sentinel. As a result, `GetBucket()` returns an unexpected value for the default case, and a user explicitly calling `SetBucket(-1)` will make `HasSpecifiedBucket()` return true. Please use the documented `-1` default/sentinel or update the API semantics consistently.

4. `GenericRow::operator==` unconditionally dereferences `kind_` and `other.kind_`. Since `SetRowKind()` accepts a raw pointer and there is no null check, a default-constructed row can be made to crash by calling `SetRowKind(nullptr)` and then comparing. Please either reject null row kinds or make equality robust to null pointers.

leaves12138 (changes requested):
Thanks for the clarification. I rechecked the previously disputed RecordBatchBuilder/ArrowArray release behavior and the empty `row_kinds_` convention, and I agree those two points should not be blockers. I have removed them from the review.

I still think the following issues should be fixed before merge:

1. The bucket default is inconsistent with the public contract. The header says that if the bucket is not set, the default value is `-1`, but the implementation uses `std::numeric_limits<int32_t>::min()` as the stored sentinel. As a result, `GetBucket()` returns an unexpected value for the default case, and a user explicitly calling `SetBucket(-1)` will make `HasSpecifiedBucket()` return true. Please use the documented `-1` default/sentinel or update the API semantics consistently.

2. `GenericRow::operator==` unconditionally dereferences `kind_` and `other.kind_`. Since `SetRowKind()` accepts a raw pointer and there is no null check, a row can be made to crash by calling `SetRowKind(nullptr)` and then comparing. Please either reject null row kinds or make equality robust to null pointers.

leaves12138 (approved):
Thanks for the update. I rechecked the two previously blocking points: the bucket sentinel/API contract is now documented consistently, and GenericRow equality is robust when row kind pointers are null. I do not see remaining blockers.

PR #29, opened 2026-05-28
Configure squash-only merge button

leaves12138:
This updates `.asf.yaml` so Apache GitHub settings expose only the squash merge button for pull requests. Merge commits and rebase merges are both disabled; squash remains enabled.


JingsongLi (approved):
+1

PR #21, opened 2026-05-26
feat: add GenericLruCache, ConcurrentHashMap, MurmurHash, and Precond…

dalingmeng:
### Purpose

No Linked issue.

Introduce additional common utility modules to the apache/paimon-cpp codebase, including a generic LRU cache, concurrent hash map, murmur hash utilities, and precondition checks.

### Tests

generic_lru_cache_test.cpp — LRU eviction, TTL expiration, concurrent access
concurrent_hash_map_test.cpp — Insert, lookup, delete, and multi-threaded correctness
murmurhash_utils_test.cpp — Hash computation correctness and consistency
preconditions_test.cpp — CheckState and CheckArgument validation

### API and Format

No API or format changes.

### Documentation

No documentation changes.

PR #30, opened 2026-05-29
feat: migrate common/data/columnar module

leaves12138:
Replacement for #24 because the original branch is dirty after #21/#22/#23 were merged and this maintainer token cannot push to the contributor fork.

The net change is the same columnar module from #24, applied on top of current `main`; the `date_time_utils.h` conflict was resolved by keeping the newer timezone helpers from #23 and retaining the timestamp helper APIs needed by the columnar code.

Closes #24.


leaves12138:
Closing this replacement PR because the original PR #24 will be updated manually by the author to resolve the conflicts.

PR #31, opened 2026-05-29
feat: add binary row serializers and blob implementation

lszskye:
### Purpose

Introduce serialization infrastructure for Paimon's binary row format, including `BinaryRowSerializer`, `BinarySerializerUtils`, `RowCompactedSerializer`, and the `Blob` implementation.

#### `BinaryRowSerializer`
- Serializes/deserializes `BinaryRow`

#### `BinarySerializerUtils`
- Utility class for converting `InternalRow`, `InternalArray`, `InternalMap` into their binary counterparts (`BinaryRow`, `BinaryArray`, `BinaryMap`)

#### `RowCompactedSerializer`
- Compact serialization format for `InternalRow`, using bitset-based null tracking and variable-length integer encoding
- Provides `SerializeToBytes()` / `Deserialize()` for row-level serialization

#### `Blob`
- Implementation of `Blob` class for managing blob descriptors and blob data

### Tests

- `BinaryRowSerializerTest`
- `BinarySerializerUtilsTest`
- `RowCompactedSerializerTest`

### DIFF
--- a/src/paimon/common/data/blob.cpp
+++ b/src/paimon/common/data/blob.cpp
@@ -38,8 +38,8 @@ class MemoryPool;
 
 class Blob::Impl {
  public:
-    Impl(std::unique_ptr<BlobDescriptor>&& descriptor, const std::string& uri)
-        : descriptor_(std::move(descriptor)), uri_(uri) {}
+    explicit Impl(std::unique_ptr<BlobDescriptor>&& descriptor)
+        : descriptor_(std::move(descriptor)) {}
 
     PAIMON_UNIQUE_PTR<Bytes> SerializeDescriptor(const std::shared_ptr<MemoryPool>& pool) const {
         return descriptor_->Serialize(pool);
@@ -50,12 +50,11 @@ class Blob::Impl {
     }
 
     const std::string& Uri() const {
-        return uri_;
+        return descriptor_->Uri();
     }
 
  private:
     std::unique_ptr<BlobDescriptor> descriptor_;
-    std::string uri_;
 };
 
 Result<std::unique_ptr<Blob>> Blob::FromPath(const std::string& path) {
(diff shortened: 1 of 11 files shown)


src/paimon/common/data/serializer/row_compacted_serializer.h
```diff
@@ -0,0 +1,202 @@
...
+#pragma once
+#include <functional>
+
+#include "arrow/api.h"
+#include "paimon/common/data/binary_row.h"
+#include "paimon/common/data/binary_writer.h"
+#include "paimon/common/memory/memory_segment.h"
+#include "paimon/common/memory/memory_segment_utils.h"
+#include "paimon/common/memory/memory_slice.h"
+#include "paimon/common/utils/var_length_int_utils.h"
```
leaves12138, line 29: This header is not present in the PR head or the current base branch, so including `paimon/common/utils/var_length_int_utils.h` makes this code fail to compile. Please add the missing utility header/implementation or switch to an existing varint helper.

src/paimon/common/data/blob.cpp
```diff
@@ -0,0 +1,119 @@
...
+Result<std::unique_ptr<Blob>> Blob::FromPath(const std::string& path) {
+    return FromPath(path, /*offset=*/0, /*length=*/-1);
+}
+
+Result<std::unique_ptr<Blob>> Blob::FromPath(const std::string& path, int64_t offset,
+                                             int64_t length) {
+    PAIMON_ASSIGN_OR_RAISE(std::string normalized_path, PathUtil::NormalizePath(path));
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<BlobDescriptor> descriptor,
+                           BlobDescriptor::Create(normalized_path, offset, length));
+    auto impl = std::make_unique<Blob::Impl>(std::move(descriptor), descriptor->Uri());
```
leaves12138: This passes `std::move(descriptor)` and `descriptor->Uri()` in the same function call. The argument evaluation order is unspecified, so `descriptor->Uri()` may run after the unique_ptr has been moved from and become null. Please capture the URI before moving `descriptor`; the same pattern below in `FromDescriptor` needs the same fix.

leaves12138 (changes requested):
Thanks for the PR. I found two blockers that should be fixed before merge: one compile-time missing include and one moved-from unique_ptr dereference risk in Blob construction.

leaves12138 (approved):
Re-reviewed the latest head. The Blob construction no longer uses the descriptor after moving it, so the unspecified argument evaluation issue is resolved. I am not treating the var_length_int_utils dependency as a blocker per the planned follow-up. I did not find other blockers in this pass.

PR #32, opened 2026-05-29
feat: add projected_array, projected_row, field comparator, partition…

lszskye:
### Purpose
Introduce core utility classes for row/array projection, field comparison, partition computation, and internal row helper functions.

#### `ProjectedRow`
- An `InternalRow` implementation that provides a projected (column-selected and reordered) view of an underlying `InternalRow`

#### `ProjectedArray`
- An `InternalArray` implementation that provides a projected view of an underlying `InternalArray`

#### `FieldsComparator`
- Compares two `InternalRow` instances field-by-field based on configurable sort fields and ordering
- Supports all Paimon data types
- Implements Java-compatible floating-point ordering: `-infinity < -0.0 < +0.0 < +infinity < NaN == NaN`

#### `BinaryRowPartitionComputer`
- Converts between partition string maps and `BinaryRow` representations
- Supports all partition column types via `PartitionConverter`

#### `InternalRowUtils`
- Utility functions for converting between `InternalArray` and `std::vector<std::string>` / `std::vector<std::optional<std::string>>`

### Tests
- `BinaryRowPartitionComputerTest`
- `FieldsComparatorTest`
- `InternalRowUtilsTest`
- `ProjectedRowTest`
- `ProjectedArrayTest`

### DIFF
--- a/src/paimon/common/utils/binary_row_partition_computer.cpp
+++ b/src/paimon/common/utils/binary_row_partition_computer.cpp
@@ -1,0 +1,164 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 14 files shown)


leaves12138 (changes requested):
I found one correctness blocker in `FieldsComparator`: the `FLOAT` and `DOUBLE` comparators use plain `==` and `<`, even though `CompareFloatingPoint()` is already defined for Java-compatible ordering. This makes comparisons with NaN non-antisymmetric (for example, both `1.0 <=> NaN` and `NaN <=> 1.0` return `1`), and it also treats `-0.0` and `+0.0` as equal despite the intended ordering documented in `CompareFloatingPoint`. Please route both float and double comparisons through `CompareFloatingPoint()` and add tests for NaN / signed-zero cases. I did not find other blockers in this pass.

leaves12138 (approved):
Re-reviewed with the clarified intended semantics. The previous blocker I raised was incorrect: FieldsComparator is intended to use PK comparator semantics where -0.0 and +0.0 compare equal and NaN is treated as equal to +infinity, while CompareFloatingPoint is for range index / SST key ordering. With that distinction, I do not find blockers in this pass.

PR #38, opened 2026-06-02
feat: add data_converter_utils, field_type_utils, file_type and decim…

lszskye:
### Purpose
Introduce utility classes for data type conversion, field type introspection, file type classification, and decimal operations.

### Changes
`DataConverterUtils`: Provides bidirectional converters between string values and `BinaryRow` fields 

`FieldTypeUtils`: Utility for `FieldType`, e.g., integer type checking, scale comparison

 `FileType`: Enum class defining file types: `kMeta`, `kData`, `kBucketIndex`, `kGlobalIndex`, `kFileIndex`. 

`DecimalUtils`: validates Arrow decimal type constraints

### Tests
- `DataConverterUtilsTest`
- `FieldTypeUtilsTest`
- `FileTypeTest`
- `DecimalUtilsTest`

### DIFF
--- a/src/paimon/common/utils/data_converter_utils.h
+++ b/src/paimon/common/utils/data_converter_utils.h
@@ -116,26 +116,6 @@ class DataConverterUtils {
                     return Status::OK();
                 };
                 break;
-            case arrow::Type::FLOAT:
-                converter = [](const std::string& value_str, int32_t field_idx,
-                               BinaryRowWriter* writer) {
-                    auto value = StringUtils::StringToValue<float>(value_str);
-                    RETURN_INVALID_WITH_FIELD_INFO(value, field_idx, value_str,
-                                                   arrow::internal::ToString(arrow::Type::FLOAT));
-                    writer->WriteFloat(field_idx, value.value());
-                    return Status::OK();
-                };
-                break;
-            case arrow::Type::DOUBLE:
-                converter = [](const std::string& value_str, int32_t field_idx,
-                               BinaryRowWriter* writer) {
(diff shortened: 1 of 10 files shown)


src/paimon/common/utils/data_converter_utils.h
```diff
@@ -0,0 +1,280 @@
...
+                                arrow::internal::ToString(type)));
+        }
+        return converter;
+    }
+
+    // support float and double
+    template <typename T>
+    static std::string FloatValueToString(const T& value, int32_t precision) {
+        std::stringstream oss;
+        if (value >= 1e-3 && value <= 1e7) {
```
leaves12138: This condition should use the magnitude rather than the signed value. As written, negative values such as `-233.0` or `-467.66472` take the scientific branch and are emitted as `-2.33E2` / `-4.6766472E2`, while positive values in the same range are emitted in fixed decimal form. That makes `CreateBinaryRowFieldToStringConverter` fail to round-trip negative FLOAT/DOUBLE partition values and can produce Java-incompatible partition path names. Please use `std::abs(value)` for the range check and add negative float/double coverage.

leaves12138 (changes requested):
Thanks for the migration. I found one blocker in DataConverterUtils: non-legacy partition string conversion for negative FLOAT/DOUBLE values uses scientific notation because the fixed-format branch only accepts positive values. This breaks round-trip/path compatibility for negative floating partition values.

leaves12138 (approved):
Re-reviewed PR #38. Given the current scope does not support FLOAT/DOUBLE as partition fields, the removal of that conversion path is acceptable. I did not find other blockers.

PR #39, opened 2026-06-02
feat: add IO stream infrastructure

lszskye:
#### `DataInputStream`
- Abstract input stream for reading primitive types (`int8`, `int16`, `int32`, `int64`, `float`, `double`) and byte arrays

#### `DataOutputStream`
- Abstract output stream for writing primitive types and byte arrays

#### `ByteArrayInputStream`
- `DataInputStream` backed by an in-memory byte array (`Bytes`), supporting sequential reads with position tracking

#### `BufferedInputStream`
- Buffered wrapper around `DataInputStream` with configurable buffer size, supporting `Seek()`, `Skip()`, and efficient sequential reads

#### `OffsetInputStream`
- `DataInputStream` that reads from a sub-range of an underlying stream, defined by offset and length, with boundary enforcement

#### `MemorySegmentOutputStream`
- `DataOutputStream` backed by `MemorySegment` with auto-growing capacity, producing `MemorySlice` output via `ToSlice()`

### Tests

- `BufferedInputStreamTest`
- `ByteArrayInputStreamTest`
- `DataInputOutputStreamTest`
- `MemorySegmentOutputStreamTest`
- `OffsetInputStreamTest`

### DIFF
--- a/src/paimon/common/io/byte_array_input_stream.cpp
+++ b/src/paimon/common/io/byte_array_input_stream.cpp
@@ -1,0 +1,107 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "paimon/io/byte_array_input_stream.h"
+
+#include <cassert>
+#include <cstring>
+#include <utility>
+
+#include "fmt/format.h"
+
+namespace paimon {
+ByteArrayInputStream::ByteArrayInputStream(const char* buffer, uint64_t length)
+    : buffer_(buffer), length_(length), position_(0) {
+    assert(buffer_);
+}
+
+const char* ByteArrayInputStream::GetRawData() const {
+    return buffer_ + position_;
+}
+
+Status ByteArrayInputStream::Seek(int64_t offset, SeekOrigin origin) {
+    switch (origin) {
+        case SeekOrigin::FS_SEEK_SET: {
+            position_ = offset;
+            break;
+        }
+        case SeekOrigin::FS_SEEK_CUR: {
+            position_ += offset;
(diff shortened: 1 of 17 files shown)


src/paimon/common/io/byte_array_input_stream.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const char* ByteArrayInputStream::GetRawData() const {
+    return buffer_ + position_;
+}
+
+Status ByteArrayInputStream::Seek(int64_t offset, SeekOrigin origin) {
+    switch (origin) {
+        case SeekOrigin::FS_SEEK_SET: {
+            position_ = offset;
```
leaves12138, line 41: Please compute the target position in a local variable and assign `position_` only after validation succeeds. Currently a failed seek (for example, seeking past EOF) returns an error but leaves `position_` out of range, so subsequent `GetPos()`/`Read()` calls observe corrupted stream state. The same invariant should hold for `FS_SEEK_CUR` and `FS_SEEK_END`.

src/paimon/common/io/offset_input_stream.cpp
```diff
@@ -0,0 +1,137 @@
...
+}
+
+OffsetInputStream::OffsetInputStream(const std::shared_ptr<InputStream>& wrapped, int64_t length,
+                                     int64_t offset)
+    : wrapped_(wrapped), length_(length), offset_(offset) {}
+
+Status OffsetInputStream::Seek(int64_t offset, SeekOrigin origin) {
+    switch (origin) {
+        case SeekOrigin::FS_SEEK_SET: {
+            inner_position_ = offset;
```
leaves12138, line 64: Same issue here: `inner_position_` is mutated before both boundary validation and the wrapped seek succeed. A failed seek should not change the logical position. Please calculate a local target position first, validate it, seek the wrapped stream, and only then commit `inner_position_`.

src/paimon/common/io/offset_input_stream.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+Result<int64_t> OffsetInputStream::GetPos() const {
+    return inner_position_;
+}
+
+Result<uint64_t> OffsetInputStream::Length() const {
+    return length_;
+}
+
+Status OffsetInputStream::AssertBoundary(int32_t inner_pos) const {
```
leaves12138, line 128: This boundary check should not take `int32_t`. The stream API and `length_` use 64-bit offsets, so passing values like `inner_position_ + size` or `offset + size` through `int32_t` can truncate offsets above 2 GiB and either reject valid reads or allow out-of-range reads. Please use a 64-bit type consistently here.

leaves12138 (changes requested):
I found correctness blockers in the new stream implementations: failed seeks can leave stream positions corrupted, and OffsetInputStream boundary checks truncate 64-bit offsets. Please fix these before merging.

leaves12138 (approved):
Re-reviewed PR #39. The remaining concerns about failed seek position rollback and int64_t file system offsets are planned for a later unified file-system refactor, so they should not block this PR. I did not find other blockers.

PR #40, opened 2026-06-02
feat: add IO cache (LRUCache, CacheManager, CacheInputStream) and MemorySlice utilities

lxy-9602:
### Purpose

No Linked issue.

Introduce IO caching infrastructure and memory slice utilities:

**IO Cache (`src/paimon/common/io/`):**
- `Cache` — abstract cache interface (`cache/cache.h`)
- `CacheKey` — cache key with file path and offset identity (`cache/cache_key.h/cpp`)
- `LRUCache` — least-recently-used cache implementation (`cache/lru_cache.h/cpp`)
- `CacheManager` — manages cache lifecycle and eviction (`cache/cache_manager.h/cpp`)
- `CacheInputStream` — input stream backed by the page cache (`cache_input_stream.h`)

**Memory Slice (`src/paimon/common/memory/`):**
- `MemorySlice` — a view over a contiguous memory region (`memory_slice.h/cpp`)
- `MemorySliceInput` — sequential reader over a MemorySlice (`memory_slice_input.h`)
- `MemorySliceOutput` — growable writer producing a MemorySlice (`memory_slice_output.h/cpp`)

### Tests

- `cache_input_stream_test.cpp` — cached read behavior and cache hits/misses
- `lru_cache_test.cpp` — LRU eviction order and capacity limits
- `memory_slice_test.cpp` — slice construction, subslicing, comparison

### Generative AI tooling
Migrate-by: Aone Copilot (Claude)

### DIFF
--- a/src/paimon/common/memory/memory_slice_output.cpp
+++ b/src/paimon/common/memory/memory_slice_output.cpp
@@ -1,0 +1,118 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "paimon/common/memory/memory_slice_output.h"
+
+#include "fmt/format.h"
+#include "paimon/common/utils/math.h"
+#include "paimon/common/utils/var_length_int_utils.h"
+namespace paimon {
+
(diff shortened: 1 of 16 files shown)


src/paimon/common/memory/memory_slice_output.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+bool MemorySliceOutput::NeedSwap() const {
+    return SystemByteOrder() != byte_order_;
+}
+
+void MemorySliceOutput::EnsureSize(int32_t size) {
+    if (size <= segment_.Size()) {
+        return;
+    }
```
leaves12138, line 95: `size` is already the required total capacity passed by callers (`size_ + bytes_to_write`). Adding the current segment size again makes growth much larger than needed; for example, appending one byte to a full power-of-two buffer grows from N to 4N. More importantly, if `MemorySliceOutput(0, pool)` is used, `capacity` remains 0 and `capacity <<= 1` loops forever. Please use `min_capacity = size` and handle `capacity == 0` explicitly (or reject/normalize zero initial capacity), and add coverage for the zero-capacity case.

leaves12138 (changes requested):
Thanks for the migration. I found one blocker in MemorySliceOutput growth: EnsureSize treats the already-required total size as an increment, which causes excessive growth and can hang forever for a zero-sized initial buffer.

leaves12138 (approved):
I re-checked MemorySliceOutput against Java Paimon's MemorySliceOutput.ensureSize(). The C++ implementation intentionally follows the same growth policy, and current callers initialize it with a non-zero capacity. I did another pass over the cache and memory slice code and did not find remaining blockers.

PR #41, opened 2026-06-02
feat(fs): introduce local file system

zjw1111:
### Purpose

Linked issue: None

Introduce the local file system implementation migrated from the Alibaba Paimon C++ repository.

This PR adds:
- local file primitives and file status wrappers
- `LocalFileSystem` input/output stream support
- local file system factory registration
- local file unit coverage

The requested migration scope was `src/paimon/fs/local` excluding `CMakeLists.txt`. No additional dependency files were required.

License handling:
- Replaced Alibaba-owned source headers with ASF headers.
- Checked source and target `LICENSE` / `NOTICE`; no third-party declaration updates were required for these files.

External contributor handling:
- Re-ran blame-based contributor analysis for the final file set.
- No external contributor thresholds were hit, so no `Co-authored-by` trailer or PR thank-you comment is needed.

### API and Format

This adds local file system implementation classes under `src/paimon/fs/local`. It does not change public storage format or protocol.

### Documentation

No documentation changes.

### Generative AI tooling

Migrated-by: OpenAI Codex

### DIFF
--- a/src/paimon/fs/local/local_file_system.cpp
+++ b/src/paimon/fs/local/local_file_system.cpp
@@ -1,0 +1,327 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "paimon/fs/local/local_file_system.h"
+
+#include <fcntl.h>
+
+#include <cassert>
+#include <cerrno>
+#include <cstdio>
+#include <cstring>
+#include <utility>
+
+#include "fmt/format.h"
+#include "paimon/common/utils/path_util.h"
+#include "paimon/common/utils/string_utils.h"
+#include "paimon/fs/local/local_file_status.h"
+
+namespace paimon {
+
+Result<LocalFile> LocalFileSystem::ToFile(const std::string& path_string) const {
+    // local file system does not support path_string with scheme, e.g., "file:/tmp" will be
+    // rewritten to "/tmp"
+    PAIMON_ASSIGN_OR_RAISE(Path path, PathUtil::ToPath(path_string));
+    if (!path.scheme.empty() && StringUtils::ToLowerCase(path.scheme) != "file") {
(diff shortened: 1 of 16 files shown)


src/paimon/fs/local/local_file_system.cpp
```diff
@@ -0,0 +1,327 @@
...
+        return Status::NotExist(fmt::format("File '{}' not exists", path));
+    }
+    PAIMON_ASSIGN_OR_RAISE(LocalFile file, ToFile(path));
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<LocalInputStream> in, LocalInputStream::Create(file));
+    return in;
+}
+
+Result<std::unique_ptr<OutputStream>> LocalFileSystem::Create(const std::string& path,
+                                                              bool overwrite) const {
+    PAIMON_ASSIGN_OR_RAISE(bool is_exist, Exists(path));
```
leaves12138, line 53: This calls `Exists(path)` before creating the parent directories. For a new file under a non-existing parent (for example `/tmp/a/b/file`), `ToFile(path).Exists()` returns an IOError because `access()` fails with `ENOENT` on the missing parent path, so `Create` returns before lines 68-71 can create the parent directory and open the output stream. The `Create` path should treat missing targets as non-existing, then create parents before opening the file; otherwise `FileSystem::Create` only works when the parent directory already exists.

leaves12138 (changes requested):
I found a correctness blocker in local file creation: creating a path whose parent does not exist currently fails before it reaches the recursive Mkdirs call. Please fix this before merging.

src/paimon/fs/local/local_file_system.cpp
```diff
@@ -0,0 +1,327 @@
...
+    PAIMON_ASSIGN_OR_RAISE(bool is_file, src_file.IsFile());
+    std::string new_file_name = dst;
+
+    if (is_file && new_file_name[new_file_name.length() - 1] == '/') {
+        return Status::Invalid(err_msg, "src file is not a dir");
+    }
+    PAIMON_ASSIGN_OR_RAISE(LocalFile dst_file, ToFile(dst));
+    auto parent = dst_file.GetParentFile();
+    PAIMON_RETURN_NOT_OK(Mkdirs(parent.GetAbsolutePath()));
+    if (::rename(src.c_str(), dst.c_str()) != 0) {
```
leaves12138: This should use the normalized local paths, not the original `src`/`dst` strings. Earlier in this method, `Exists(src)`, `Exists(dst)`, `ToFile(src)`, and `ToFile(dst)` all strip the optional `file:` scheme via `ToFile`, but the actual syscall still passes the raw URI strings to `::rename`. So `Rename("file:/tmp/a", "file:/tmp/b")` validates `/tmp/a` and creates `/tmp`, then tries to rename a literal path named `file:/tmp/a`, which fails. Please call `::rename(src_file.GetAbsolutePath().c_str(), dst_file.GetAbsolutePath().c_str())` (or equivalent normalized paths).

leaves12138 (changes requested):
Thanks for pointing out my previous review mistake: access(..., F_OK) returns ENOENT when a parent directory is missing, so Create can continue to Mkdirs(parent). I re-reviewed the PR and found a different blocker in Rename: it validates normalized local paths but calls ::rename with the original URI strings.

leaves12138 (approved):
Reviewed the latest update. The previous Rename blocker is resolved: Rename now operates on normalized local paths, and I did not find any new blocking issue in this revision.

PR #55, opened 2026-06-08
feat(schema): add schema validation utilities

lucasfang:
### Purpose

Linked issue: No linked issue

This change adds schema validation support for table schemas and Arrow schemas.

Included changes:

- Table schema validation:
  - Adds schema validation helpers for checking table schema definitions.
  - Adds validation coverage for fields, partition keys, primary keys, options, and schema constraints.
- Arrow schema validation:
  - Adds Arrow schema validator utilities for checking compatibility with table schema definitions.
  - Adds validation coverage for supported Arrow field types, nested structures, and incompatible schema cases.

### Tests

Not run. Local compile, CMake, and gtest environment checks are not part of this PR description.

- `SchemaValidationTest`
- `ArrowSchemaValidatorTest`

### API and Format

### Documentation

### Generative AI tooling

Migrate-by: Aone Copilot (GPT-5.5)

### DIFF
--- a/src/paimon/core/schema/arrow_schema_validator.cpp
+++ b/src/paimon/core/schema/arrow_schema_validator.cpp
@@ -1,0 +1,266 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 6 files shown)


leaves12138 (changes requested):
I found a blocking issue in the validation path. `SchemaValidation::ValidateSequenceGroup` has a reachable `assert(false)` when a sequence-group option key references a field that is not present in the table schema, for example `fields.f0,f3.sequence-group`. This is user-provided table configuration, and the added test expects an `Invalid` status for this case, but debug/test builds will abort before returning that status.

Please remove the assertion and return the `Invalid` status directly (or otherwise handle the bad configuration without aborting). I did not find another blocking issue in this pass.

leaves12138 (approved):
Re-reviewed the latest revision. The previous blocking issue is fixed: `SchemaValidation::ValidateSequenceGroup` now returns an `Invalid` status for missing sequence-group fields instead of aborting through `assert(false)`. I did not find any remaining blocking issue within this staged migration scope. Approving.

PR #57, opened 2026-06-08
feat(core): introduce snapshot cleanup operations

zjw1111:
### Purpose

Linked issue: close #xxx

Introduce snapshot expiration and orphan file cleanup operations from the Alibaba Paimon C++ repository.

Migrated files:
- `include/paimon/orphan_files_cleaner.h`
- `src/paimon/core/operation/expire_snapshots.h`
- `src/paimon/core/operation/expire_snapshots.cpp`
- `src/paimon/core/operation/expire_snapshots_test.cpp`
- `src/paimon/core/operation/orphan_files_cleaner.cpp`
- `src/paimon/core/operation/orphan_files_cleaner_impl.h`
- `src/paimon/core/operation/orphan_files_cleaner_impl.cpp`
- `src/paimon/core/operation/orphan_files_cleaner_test.cpp`
- `src/paimon/core/operation/metrics/clean_metrics.h`

The dependency closure is intentionally not included in this PR.

### Tests

- `git diff --cached --check`
- `git diff --check`

### API and Format

This change adds the public `orphan_files_cleaner.h` API. It does not change storage format or protocol.

### Documentation

No documentation changes.

### Generative AI tooling

Migrated-by: OpenAI Codex

### DIFF
--- a/include/paimon/orphan_files_cleaner.h
+++ b/include/paimon/orphan_files_cleaner.h
@@ -1,0 +1,197 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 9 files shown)


zjw1111:
Thanks @chongchongxiao for your previous contributions to paimon-cpp. This migration PR carries that work forward into Apache Paimon C++.

leaves12138 (changes requested):
Thanks for migrating this feature. I think we need a couple of correctness fixes before merging:

1. `snapshot.time-retained` is checked against the current candidate snapshot instead of its next snapshot. Snapshot `N` should only be expired after snapshot `N + 1` has been retained long enough. With the current loop, if snapshot `5` is still within the retention window, `ExpireUntil(..., 5)` can still delete snapshot `4`, even though its successor is not old enough yet. Please align this with the Java implementation by checking `snapshot_id + 1` and add a regression test for the boundary.

2. `CleanEmptyDirectories` schedules async deletion tasks that capture the range-for variable by reference (`[this, &empty_directory]`). Those tasks may execute after the loop variable has advanced or gone out of scope, so this is a use-after-scope risk and may delete the wrong directory. Please capture the path by value, e.g. `[this, empty_directory]`, and add coverage if possible.

PR #59, opened 2026-06-09
feat: add core options and utility helpers

lucasfang:
### Purpose

Linked issue: No linked issue

This change adds core options and utility support used by table metadata, partition path handling, field mapping, offsets, durations, and commit increments.

Included changes:

- Public utility API:
  - Adds `include/paimon/utils/special_field_ids.h` for shared special field identifiers.

- Core options:
  - Adds option enums and configuration helpers for changelog producer, compression, expiration, external path strategy, lookup compact mode, lookup strategy, merge engine, sort engine, and sort order.

- Core utilities:
  - Adds path factory helpers.
  - Adds file utility helpers.
  - Adds partition path parsing and generation utilities.
  - Adds field mapping utilities.
  - Adds offset row support.
  - Adds duration, commit increment, and special field id utility implementations.

- Test coverage:
  - Adds tests for lookup strategy, file utilities, partition path utilities, field mapping, and offset row behavior.

### Tests

Not run.

- `src/paimon/core/options/lookup_strategy_test.cpp`
- `src/paimon/core/utils/file_utils_test.cpp`
- `src/paimon/core/utils/partition_path_utils_test.cpp`
- `src/paimon/core/utils/field_mapping_test.cpp`
- `src/paimon/core/utils/offset_row_test.cpp`

### API and Format

This change adds public API in:

- `include/paimon/utils/special_field_ids.h`

No storage format or protocol changes.

### Documentation

### Generative AI tooling

Migrate-by: Aone Copilot (GPT-5.5)

### DIFF
--- a/include/paimon/utils/special_field_ids.h
+++ b/include/paimon/utils/special_field_ids.h
@@ -1,0 +1,47 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <limits>
+
+namespace paimon {
+
+/// A utility class for accessing special field IDs used in metadata.
+class SpecialFieldIds {
+ protected:
+    /// System defined constant for field id boundary. Value: INT32_MAX - 10000
+    static const int32_t CPP_FIELD_ID_END = std::numeric_limits<int32_t>::max() - 10000;
+
+ public:
+    /// Special field ID reserved for sequence number. Value: INT32_MAX - 1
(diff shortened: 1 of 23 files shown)


leaves12138 (changes requested):
Thanks for migrating these options and helpers. I found one correctness issue that should be fixed before merging.

`FieldMappingBuilder::CreateDataFilters` can incorrectly push down one side of an `OR` predicate after schema evolution. The method first splits only top-level `AND`s, then `ReconstructPredicateWithDataFields` returns `nullopt` for a whole predicate subtree when any child cannot be mapped. That behavior is correct for `OR` because partially pushing down `(A OR B)` as just `A` would filter out rows that only satisfy `B`. However, after reconstruction it calls `PredicateUtils::ExcludePredicateWithFields`, which removes any predicate that contains a partition field. For a mixed `OR`, for example `(partition_key = 1 OR data_col = 2)`, reconstruction succeeds for the whole `OR`, but `ExcludePredicateWithFields` then drops the entire predicate from the non-partition filter; the data reader will not apply `data_col = 2`, so files/rows that should be pruned can be read. More importantly, the same pattern in `CreatePartitionInfo` with `CreatePickedFieldFilter` can transform mixed `OR` predicates into a partition-only predicate, which is not logically equivalent and can incorrectly prune partitions.

Please handle mixed-field `OR` predicates conservatively: only push a compound predicate to partition/non-partition filters when the whole subtree can be represented on that side without changing semantics, otherwise skip pushing that subtree down. A regression test like `(part_col = 1 OR non_part_col = 2)` would be useful for both partition and non-partition filters.

leaves12138 (approved):
Thanks for the update. I re-reviewed the latest diff at `9375915ea45aa3b6dc2b0ea9897d582170e702ed`.

The previous blocking issue around `FieldMapping` predicate pushdown is no longer present in this PR because those files were removed from the current diff. I also re-checked the remaining option and utility helpers (`PartitionPathUtils`, `FileUtils`, `OffsetRow`, `LookupStrategy`, special field IDs, etc.) and did not find any remaining blocking correctness issue.

GitHub currently reports no checks for this branch, so please make sure the relevant C++ build/tests are run before merging.

PR #60, opened 2026-06-09
feat: add core table metadata utilities

lucasfang:
### Purpose

Linked issue: No linked issue

This change adds core utility support for table metadata path construction, snapshot/tag/branch management, manifest metadata reading, object serialization, and primary-key table helpers.

Included changes:

- File store and index path utilities:
  - Adds `FileStorePathFactory` and file store path factory cache utilities.
  - Adds index file path factory helpers.
  - Adds test coverage for file store path factory, cache, and index path factory behavior.
- Snapshot, tag, and branch utilities:
  - Adds snapshot manager support for snapshot discovery, lookup, commit handling, and expiration-related helpers.
  - Adds tag manager and branch manager helpers.
  - Adds test coverage for snapshot, tag, and branch manager behavior.
- Manifest and object utilities:
  - Adds manifest metadata reader utilities.
  - Adds object file, object serializer, and versioned object serializer helpers.
  - Adds batch writer support.
  - Adds test coverage for manifest metadata reader behavior.
- Primary-key table utilities:
  - Adds primary-key table helper functions.
  - Adds test coverage for primary-key table utility behavior.

### Tests

Not run. Local compile, CMake, and gtest environment checks are not part of this PR description.

- `file_store_path_factory_test.cpp`
- `file_store_path_factory_cache_test.cpp`
- `snapshot_manager_test.cpp`
- `tag_manager_test.cpp`
- `branch_manager_test.cpp`
- `manifest_meta_reader_test.cpp`
- `index_file_path_factories_test.cpp`
- `primary_key_table_utils_test.cpp`

### API and Format

### Documentation

### Generative AI tooling

Migrate-by: Aone Copilot (GPT5.5)

### DIFF
--- a/src/paimon/core/utils/batch_writer.h
+++ b/src/paimon/core/utils/batch_writer.h
@@ -1,0 +1,76 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+#include <memory>
+
+#include "paimon/result.h"
+
+struct ArrowArray;
+
+namespace paimon {
+
+class CommitIncrement;
+class RecordBatch;
+
+/// The BatchWriter is responsible for writing data and handling in-progress files used to
+/// write yet un-staged data. The incremental files ready to commit is returned to the system by the
+/// `PrepareCommit(bool)`.
+class BatchWriter {
+ public:
+    BatchWriter() = default;
+    virtual ~BatchWriter() = default;
+
+    /// Return current in-memory data usage in bytes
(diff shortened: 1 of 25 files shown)


leaves12138 (changes requested):
Thanks for adding these metadata utilities. I found one blocking lifecycle issue that should be fixed before merging.

`FileStorePathFactory` inherits from `std::enable_shared_from_this<FileStorePathFactory>`, and several public methods call `shared_from_this()` (`CreateManifestFileFactory`, `CreateManifestListFactory`, `CreateIndexManifestFileFactory`, and `CreateGlobalIndexFileFactory`). However, `FileStorePathFactory::Create` returns `Result<std::unique_ptr<FileStorePathFactory>>`, so callers can and do own the factory only through `std::unique_ptr` (for example the helper in `file_store_path_factory_test.cpp`). Calling any of those factory methods on such an object will throw `std::bad_weak_ptr` at runtime because the object was never owned by a `std::shared_ptr` control block.

This is also easy to hit indirectly: `CreateIndexFileFactory()` calls `CreateGlobalIndexFileFactory()` when `index_file_in_data_file_dir_` is false. Please make the ownership model consistent before merging, e.g. return `Result<std::shared_ptr<FileStorePathFactory>>` from `Create` and update callers/tests accordingly, or avoid `shared_from_this()` and have the nested factories use a safe ownership/lifetime strategy. A regression test should call one of these public factory methods on the normal `Create(...)` result so this does not regress.

leaves12138 (approved):
Thanks for the update. I re-reviewed the latest PR head (`cceb2e93035ea0d22c4ed3a4dc17474192bce6d9`). The previous `FileStorePathFactory` ownership issue is fixed: `Create(...)` now returns a `std::shared_ptr<FileStorePathFactory>`, so the public methods that use `shared_from_this()` are no longer invoked on a `unique_ptr`-owned object.

Treating the remaining prerequisite header dependencies as expected migration-state drift, I did not find another blocking correctness issue in these metadata utility components. The migrated implementation matches the existing C++ source snapshot aside from Apache license header updates and the shared ownership fix.

PR #64, opened 2026-06-09
feat: add data evolution reader infrastructure

lszskye:
#### `DataEvolutionArray`
- An `InternalArray` implementation composed of multiple `BinaryArray` instances. Uses `array_offsets` and `field_offsets` mapping to redirect element access to the correct underlying array and position. Elements mapped to negative offsets are treated as null, enabling schema evolution scenarios where fields may not exist in all source arrays.

#### `DataEvolutionRow`
- An `InternalRow` implementation composed of multiple `BinaryRow` instances. Uses `row_offsets` and `field_offsets` mapping to redirect field access to the correct underlying row and position. Lazily retrieves `RowKind` from the first row. Fields mapped to negative offsets return null, supporting schema evolution where columns may be absent in some source rows.

#### `DataEvolutionFileReader`
- A `BatchReader` that assembles multiple inner readers into a single unified reader. Each output field is sourced from a specific inner reader at a specific offset, controlled by `reader_offsets` and `field_offsets` arrays. Handles batch alignment across readers with different schemas, caches partial batches, and creates null arrays for non-existent fields. Only supports `NextBatchWithBitmap()`.

### Tests

- `DataEvolutionArrayTest`
- `DataEvolutionRowTest`
- `DataEvolutionFileReaderTest`


leaves12138 (changes requested):
Thanks for migrating the data evolution reader infrastructure. I found a blocking path/layout issue that should be fixed before merging.

This PR adds all seven new files at the repository root:

```text
data_evolution_array.h
data_evolution_array_test.cpp
data_evolution_file_reader.cpp
data_evolution_file_reader.h
data_evolution_file_reader_test.cpp
data_evolution_row.h
data_evolution_row_test.cpp
```

However, the code and tests include these headers as `paimon/common/reader/...`, and the existing C++ source snapshot places the same components under `src/paimon/common/reader/`. In the current PR layout, a clean build will not find headers such as `paimon/common/reader/data_evolution_array.h`, and these sources/tests are also outside the expected source tree.

Please move these files under `src/paimon/common/reader/` (preserving the `paimon/common/reader/...` include paths), or otherwise update the include/build layout consistently if a different location is intended. After moving them, please re-run the relevant C++ build/tests from a clean checkout.

leaves12138 (approved):
Thanks for the update. I re-reviewed PR #64 at head `238f90cece8a8dc092592eb7dc69e931c6c6b6c7`.

The previous path/layout blocker is fixed: the data evolution reader files are now under `src/paimon/common/reader/`, matching the `paimon/common/reader/...` include paths and the existing C++ source snapshot.

Treating the remaining prerequisite header dependencies as expected migration-state drift, I did not find another blocking correctness issue. The migrated implementation matches the existing C++ source snapshot aside from Apache license header updates.

PR #74, opened 2026-06-10
feat: add BTree global index core components

lszskye:
#### `BtreeDefs`
- Constants struct defining BTree index configuration keys and default values: compression algorithm, compression level, block size, cache size, high-priority pool ratio, and read buffer size. Also defines the `"btree"` identifier.

#### `BTreeFileFooter`
- Footer structure for BTree index files containing bloom filter handle, index block handle, and null bitmap handle. Supports versioned serialization/deserialization via `Read()`/`Write()` with a magic number (`0x50425449`) for validation. Fixed encoding length of 52 bytes.

#### `BTreeFileMetaSelector`
- Implements `FunctionVisitor<std::vector<GlobalIndexIOMeta>>` to select candidate BTree index files based on filter predicates. 

#### `BTreeIndexMeta`
- Per-file metadata holding serialized first key, last key, and a `has_nulls` flag. Keys can be null when the entire file contains only null values (`OnlyNulls()`). Provides `Serialize()`/`Deserialize()` for binary persistence with a 9-byte fixed header (two 4-byte key lengths + 1-byte null flag).

#### `KeySerializer`
- Static utility for serializing `Literal` values to `Bytes` and deserializing `MemorySlice` back to `Literal`, supporting all Paimon data types. Provides `CreateComparator()` to build a `MemorySlice::SliceComparator` for binary key comparison based on Arrow data type.

### Tests

- `BTreeFileFooterTest`
- `BTreeFileMetaSelectorTest`
- `BTreeIndexMetaTest`
- `KeySerializerTest`

### DIFF
--- a/src/paimon/common/global_index/btree/btree_defs.h
+++ b/src/paimon/common/global_index/btree/btree_defs.h
@@ -1,0 +1,63 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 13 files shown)


leaves12138 (approved):
Thanks for migrating the BTree global index core components. I reviewed PR #74 at head `5130412cd9e90563f79f7836060b6f817a845636`.

Treating the remaining prerequisite dependency state as expected migration-period drift, I did not find a blocking correctness issue in the migrated BTree definitions, file footer, index metadata, file meta selector, key serializer, and related tests. The migrated files match the existing C++ source snapshot aside from Apache license header updates.

PR #79, opened 2026-06-11
feat: add type casting infrastructure for schema evolution

lszskye:
#### `CastExecutor`
- Abstract interface for casting data between different Paimon types. Supports both single-value casting (Literal-level) and batch casting (Arrow Array-level).

#### `CastExecutorFactory`
- Singleton factory that manages a registry of `CastExecutor` instances indexed by `(source type, target type)` pairs. Resolves the appropriate executor for any supported type conversion.

#### `CastedRow`
- An `InternalRow` implementation that wraps an underlying row and applies type casting on-the-fly during field access. Reads data according to source field types and casts to target field types using per-field `CastExecutor` instances. Used in schema evolution scenarios where the stored data type differs from the read schema.

#### `CastingUtils`
- Static utility for performing Arrow Compute-based type casts on arrays and scalars. Includes integer overflow detection for casted literals and templated scalar casting with configurable `CastOptions`.

#### `NumericPrimitiveCastExecutor`
- `CastExecutor` implementation for numeric primitive type conversions (TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE). Uses a dispatch map indexed by `(source FieldType, target FieldType)` pairs for literal casting, and delegates to Arrow Compute for array-level batch casting.

### Tests

- `CastExecutorFactoryTest`
- `CastedRowTest`
- `CastingUtilsTest`

### DIFF
--- a/src/paimon/core/casting/cast_executor.h
+++ b/src/paimon/core/casting/cast_executor.h
@@ -1,0 +1,45 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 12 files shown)


zjw1111 (approved):
+1

JingsongLi (approved):
+1

PR #89, opened 2026-06-17
feat(btree): add btree global index reader/writer/factory

lszskye:
### Summary

Implement BTree-based global index reader with lazy file filtering, complementing the existing BTree global index writer. This enables efficient predicate evaluation over BTree global index files with metadata-based file pruning to minimize I/O.

### New Classes

**`BTreeGlobalIndexer`** — Top-level indexer implementation for BTree global index. Instead of building a B-Tree directly in memory, it forms a logical B-Tree via multi-level metadata over SST files, significantly reducing memory pressure during index reads. Responsible for creating corresponding Writer and Reader instances.

**`BTreeGlobalIndexReader`** — Single-file reader for BTree global index. Evaluates filter predicates (Equal, Range, In, StartsWith, etc.) against an SST file structure where each key maps to a list of row IDs. Supports null bitmap handling and min/max key pruning for fast short-circuiting.

**`BTreeGlobalIndexWriter`** — Writer for BTree global index files. Writes keys in monotonically increasing order into SST files, merging entries with identical keys into compact row ID lists. Null keys are tracked via a separate RoaringBitmap, serialized and appended to the file end on close.

**`BTreeGlobalIndexerFactory`** — Factory class for BTree global indexers. Implements the `GlobalIndexerFactory` interface and creates `BTreeGlobalIndexer` instances from configuration options.

**`LazyFilteredBTreeReader`** — Lazy-loading, file-filtering BTree global index reader for multi-file scenarios. Uses `BTreeFileMetaSelector` to prune candidate files based on predicate conditions before lazily instantiating per-file readers on demand. Supports reader caching and parallel execution.

### New Tests

- `btree_compatibility_test.cpp`
- `lazy_filtered_btree_reader_test.cpp`

### DIFF
--- a/src/paimon/common/global_index/btree/btree_compatibility_test.cpp
+++ b/src/paimon/common/global_index/btree/btree_compatibility_test.cpp
@@ -1,0 +1,931 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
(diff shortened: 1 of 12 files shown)


lxy-9602:
+1

JingsongLi (approved):

PR #99, opened 2026-06-18
feat: introduce table scan, split and plan

lszskye:
### Summary

Add the `table/source` module which provides the public API and core implementation for table scan, split generation, and plan creation. This module serves as the entry point for batch and streaming read operations, supporting split serialization/deserialization compatible with Java Paimon for cross-language interoperability.

#### Public API (`include/paimon/table/source/`)

**`Split`** — Base class for input splits used by batch computation engines. Supports binary serialization and deserialization compatible with the Java version, enabling cross-process and cross-language split transmission. A `Split` can be either a `DataSplit` (for direct data file reads) or an `IndexedSplit` (for reads leveraging global indexes).

**`DataSplit`** — Extends `Split` for direct data file reading scenarios. Contains `SimpleDataFileMeta` describing each file's path, size, row count, sequence numbers, schema, level, creation time, and optional delete row count. Provides a file list accessor for append table reads.

**`Plan`** — Interface representing the result of a `TableScan`. Exposes the generated splits and the associated snapshot ID (or `nullopt` for empty tables).

**`TableScan`** — Scanner interface that reads table metadata and produces a `Plan`. Created from a `ScanContext`, it serves as the primary entry point for initiating table scan operations in both batch and streaming modes.

**`TableRead`** — Given a `Split` or a list of `Split`s, creates `BatchReader` instances for reading data. Manages memory allocation through a shared `MemoryPool`. 

**`StartupMode`** — Specifies the startup mode. Supports `Default`, `LatestFull`, `Latest`, `FromSnapshot`, `FromSnapshotFull`, and `FromTimestamp` modes, each with different semantics for batch vs. streaming sources. Provides string conversion and parsing.

#### Internal Implementation (`src/paimon/core/table/source/`)

**`AbstractTableScan`** — Abstract base class above `FileStoreScan` that provides input split generation logic. Implements the `CreateStartingScanner` method which routes to different `StartingScanner` implementations based on the configured `StartupMode`, handling snapshot lookup, timestamp-based resolution, and tag-based scanning.

**`DataSplitImpl`** — Concrete implementation of `DataSplit` with full serialization support. Tracks partition, bucket, file metadata, deletion files, streaming flag, and raw-convertibility. Includes a `Builder` pattern for construction and supports multiple `DataFileMeta` serializer versions (v9, v10, v12, legacy).

**`SplitGenerator`** — Generates split groups from `DataFileMeta` lists. Produces `SplitGroup`s that distinguish between raw-convertible groups (directly readable without merge) and non-raw-convertible groups (requiring merge-tree processing). Provides separate entry points for batch and streaming split generation.

**`DeletionFile`** — Represents a deletion vector index file associated with a data file. Contains path, offset, length, and optional cardinality (number of deleted rows). Supports versioned serialization/deserialization (v3 and v4+) with list-level serialize/deserialize helpers.

**`ScanMode`** — Enum specifying which part of a snapshot to scan: `ALL` (complete data files) or `DELTA` (only newly changed files).

**`PlanImpl`** — Concrete implementation of the `Plan` interface. Holds a snapshot ID and a vector of splits, and provides a static `EmptyPlan()` factory method for empty results.

### New Tests

- `deletion_file_test.cpp`
- `split_generator_test.cpp`
- `startup_mode_test.cpp`
- `table_scan_test.cpp`

### DIFF
--- a/include/paimon/table/source/data_split.h
+++ b/include/paimon/table/source/data_split.h
@@ -1,0 +1,84 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "paimon/data/timestamp.h"
+#include "paimon/memory/memory_pool.h"
+#include "paimon/result.h"
+#include "paimon/table/source/split.h"
+#include "paimon/visibility.h"
+
+namespace paimon {
+class MemoryPool;
+
+/// Input data split for reading operation. Needed by most batch computation engines.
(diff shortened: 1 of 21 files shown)


lxy-9602:
+1

JingsongLi (approved):

PR #100, opened 2026-06-18
feat: introduce type casting executors for schema evolution

lszskye:
### Summary

Add a set of `CastExecutor` implementations under `src/paimon/core/casting/` to support type casting between boolean, numeric, decimal, and string types. These executors are essential for schema evolution scenarios where column types change across table versions, enabling predicate pushdown and data reading to work correctly across different schemas.

Each executor supports two casting modes:
- **Literal casting**: Converts a single `Literal` value to the target type (used during predicate rewriting).
- **Array casting**: Converts an entire Arrow `Array` to the target type (used during batch data reading).

### New Classes

**`BooleanToDecimalCastExecutor`** — Casts boolean values to decimal types. Converts `true` to `Decimal128(1)` and `false` to `Decimal128(0)` with the specified target precision and scale.

**`BooleanToNumericCastExecutor`** — Casts boolean values to numeric primitive types (int8, int16, int32, int64, float, double). Uses a dispatch map to route casting logic per target `FieldType`, mapping `true` to `1` and `false` to `0`.

**`BooleanToStringCastExecutor`** — Casts boolean values to string type. Converts `true` to `"true"` and `false` to `"false"`.

**`DecimalToDecimalCastExecutor`** — Casts decimal values between different precision/scale decimal types. Handles rescaling by adjusting the underlying `Decimal128` representation to match the target precision and scale, with overflow detection.

**`DecimalToNumericPrimitiveCastExecutor`** — Casts decimal values to numeric primitive types. Truncates the decimal value to the target integer or floating-point type by dividing out the scale factor.

**`NumericPrimitiveToDecimalCastExecutor`** — Casts numeric primitive values (int8 through double) to decimal types. Employs templated dispatch to handle each source type, scaling the input value to the target decimal precision and scale.

**`NumericToBooleanCastExecutor`** — Casts numeric values to boolean type. Uses a dispatch map per source `FieldType`, converting zero values to `false` and non-zero values to `true`.

**`NumericToStringCastExecutor`** — Casts numeric values to string type. Uses a dispatch map per source `FieldType`, converting each numeric value to its string representation.

**`StringToBooleanCastExecutor`** — Casts string values to boolean type. Recognizes `"true"`/`"TRUE"` as `true` and `"false"`/`"FALSE"` as `false`, returning an error for unrecognized string values.

**`StringToDecimalCastExecutor`** — Casts string values to decimal types. Parses the string into an `arrow::Decimal128` value with the specified target precision and scale, with support for null propagation when parsing fails.

**`StringToNumericPrimitiveCastExecutor`** — Casts string values to numeric primitive types (int8 through double). Uses a templated dispatch map per target `FieldType`, parsing the string representation into the corresponding numeric value.

### DIFF
--- a/src/paimon/core/casting/boolean_to_decimal_cast_executor.cpp
+++ b/src/paimon/core/casting/boolean_to_decimal_cast_executor.cpp
@@ -1,0 +1,98 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "paimon/core/casting/boolean_to_decimal_cast_executor.h"
+
+#include <cassert>
+#include <cstdint>
+#include <optional>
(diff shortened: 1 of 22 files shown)


lucasfang:
+1

JingsongLi (approved):

PR #101, opened 2026-06-18
feat: introduce type casting executors for Binary, Date, and Timestam…

lszskye:
### Summary

Add a set of `CastExecutor` implementations under `src/paimon/core/casting/` to support type casting between binary, date, timestamp, string, and numeric types. These executors extend the casting framework to cover temporal and binary type conversions required for schema evolution scenarios.

Each executor supports two casting modes:
- **Literal casting**: Converts a single `Literal` value to the target type (used during predicate rewriting).
- **Array casting**: Converts an entire Arrow `Array` to the target type (used during batch data reading).

### New Classes

**`BinaryToStringCastExecutor`** — Casts binary (byte array) values to string type by interpreting the raw bytes as a UTF-8 encoded string.

**`DateToStringCastExecutor`** — Casts date values (days since epoch) to their ISO-8601 string representation (e.g., `"2024-01-15"`).

**`DateToTimestampCastExecutor`** — Casts date values to timestamp types with configurable time unit (seconds, milliseconds, microseconds, nanoseconds). Sets the time component to midnight (00:00:00). Note: C++ Paimon supports a narrower date range than Java Paimon when the target unit is nanoseconds, limited by int64 bounds.

**`NumericPrimitiveToTimestampCastExecutor`** — Casts numeric primitive values (interpreted as seconds since epoch) to timestamp types with the specified precision. Note: When the target type is nanosecond precision, the valid numeric range is narrower than Java Paimon due to int64 overflow constraints.

**`StringToBinaryCastExecutor`** — Casts string values to binary type by encoding the string content as raw bytes.

**`StringToDateCastExecutor`** — Casts string values in date format (e.g., `"2024-01-15"`) to date type (days since epoch).

**`StringToTimestampCastExecutor`** — Casts string values in timestamp format to timestamp types. Supports ISO-8601 formats including `"1970-01-01T00:00:00"`. Note: Differs from Java Paimon in that it does not support passing numeric values as timestamp strings, but additionally supports the `T`-separated format.

**`TimestampToDateCastExecutor`** — Casts timestamp values to date type by truncating the time component and retaining only the date part (days since epoch).

**`TimestampToNumericPrimitiveCastExecutor`** — Casts timestamp values to numeric primitive types by extracting the epoch seconds. When the source timestamp has sub-second precision, the fractional part is truncated.

**`TimestampToStringCastExecutor`** — Casts timestamp values to their ISO-8601 string representation with appropriate precision (e.g., `"2024-01-15 12:30:45.123"`).

**`TimestampToTimestampCastExecutor`** — Converts between `TIMESTAMP` and `TIMESTAMP_WITH_LOCAL_TIME_ZONE` types, handling precision changes (e.g., milliseconds to nanoseconds) by scaling the underlying int64 value accordingly.

### DIFF
--- a/src/paimon/core/casting/binary_to_string_cast_executor.cpp
+++ b/src/paimon/core/casting/binary_to_string_cast_executor.cpp
@@ -1,0 +1,59 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "paimon/core/casting/binary_to_string_cast_executor.h"
+
+#include <cassert>
+#include <string>
+#include <utility>
+
+#include "arrow/compute/cast.h"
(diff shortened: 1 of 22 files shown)


lucasfang:
+1

JingsongLi (approved):

PR #111, opened 2026-06-23
feat: implement TableRead and FallbackDataSplit

lszskye:
### Summary

Add the implementation of `TableRead` and introduce `FallbackDataSplit` to support reading data from Paimon tables, including fallback branch reading capabilities.

### New Classes

**`TableRead` (table_read.cpp)** — The core entry point for reading data from a Paimon table. `TableRead::Create` constructs the appropriate table reader based on the `ReadContext`:
- Resolves the table schema from the schema manager or from a user-provided specific schema.
- Merges user options with table schema options.
- Dispatches to `AppendOnlyTableRead` for append-only tables or `KeyValueTableRead` for primary-key tables.
- Supports fallback branch reading: when `scan-fallback-branch` is configured, creates a `FallbackTableRead` that wraps both the primary and fallback branch readers.
- Supports system table reading by detecting and delegating to the appropriate `SystemTable` implementation.
- `CreateReader` concatenates readers from multiple splits into a single `ConcatBatchReader`.

**`FallbackDataSplit`** — A decorator over `DataSplit` that marks whether a split originates from the fallback branch. Wraps an existing `DataSplit` and exposes `IsFallback()` to indicate its origin, enabling the read layer to apply branch-specific read logic.

### Tests

- `data_split_test.cpp`
- `fallback_data_split_test.cpp`
- `table_read_test.cpp`

### DIFF
--- a/src/paimon/core/table/source/data_split_test.cpp
+++ b/src/paimon/core/table/source/data_split_test.cpp
@@ -1,0 +1,1124 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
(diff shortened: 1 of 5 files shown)


lucasfang:
+1

JingsongLi (approved):

PR #137, opened 2026-06-30
feat: introduce FieldMapping and FieldMappingBuilder

lszskye:
### Purpose
Introduce `FieldMapping` and `FieldMappingBuilder` for resolving the mapping between a read schema and a data file schema, supporting partition pruning, schema evolution, and predicate push-down.

**`FieldMapping`** — A plain data structure that holds the resolved mapping result for a single data file. It contains three parts:
- `partition_info`: mapping info for fields that are partition keys, including their index in the read schema, index in the partition row, and a partition-level filter predicate.
- `non_partition_info`: mapping info for non-partition fields that exist in both the read schema and the data schema, including field-level cast executors for type evolution and a data-file-level filter predicate with field indices rewritten to match the data schema.
- `non_exist_field_info`: mapping info for fields requested by the read schema but absent in the data file (added via schema evolution), so the reader can fill them with nulls.

**`FieldMappingBuilder`** — A builder created once per query from the read schema, partition keys, and query predicate. For each data file, `CreateFieldMapping(data_fields)` produces a `FieldMapping` by:
1. Matching fields between the read schema and the data schema by field ID (schema evolution–safe).
2. Separating matched fields into partition and non-partition groups.
3. Building `CastExecutor` chains for type-evolved fields (atomic types only; complex types are rejected).
4. Reconstructing the query predicate against the data schema field indices and data types, including literal casting for integer widening and overflow detection; predicates on non-existent or type-incompatible fields are dropped.
5. Extracting a partition-level filter from the reconstructed predicate for partition pruning.

### Tests

- `TestBasicFieldMapping` — basic read/data schema mapping without partition or predicate
- `TestFieldMappingWithPartition` — partition field separation and index mapping
- `TestFieldMappingWithPredicate` — predicate rewriting for partition fields
- `TestFieldMappingWithPartitionAndPredicateInPartitionField` — predicate push-down into partition filter
- `TestFieldMappingWithPartitionAndPredicateInNonPartitionField` — predicate push-down into data file filter
- `TestSchemaEvolution` — field add / delete / rename / type casting without predicate
- `TestSchemaEvolutionWithPredicate` — schema evolution combined with predicate rewriting and overflow detection
- `TestSchemaEvolutionWithPredicate2` — schema evolution with inconsistent field order in read schema
- `TestCompoundPredicateWithoutPushDown` — OR compound predicate containing non-pushdownable sub-predicates is correctly dropped

### DIFF
--- a/src/paimon/core/utils/field_mapping.cpp
+++ b/src/paimon/core/utils/field_mapping.cpp
@@ -1,0 +1,349 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#include "paimon/core/utils/field_mapping.h"
+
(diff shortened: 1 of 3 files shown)


lxy-9602:
+1

JingsongLi (approved):

PR #163, opened 2026-08-01
feat: support real-time append writes and reads with pluggable memory indexers

lxy-9602:
### Purpose

Linked issue: #158

This PR introduces an opt-in, process-local real-time write and union-read framework for
fixed-bucket append tables.

Applications attach the same `RealtimeContext` to their write and scan contexts. Each
partition-bucket is backed by a pluggable `MemIndexer`, so newly written rows become queryable
before the next snapshot commit. During prepare commit, Paimon seals the current segment, opens a
new building segment for subsequent writes, and flushes the sealed segment through the existing
rolling writer. Paimon therefore continues to own data-file formats, file rolling, file indexes,
statistics, and commit-message generation.

The framework assigns an internal contiguous offset range to each written batch. These offsets are
segment and snapshot progress metadata. `PrepareCommitWithProgress` returns each commit message
together with its partition, bucket, and inclusive offset range. `CommitWithProgress` orders and
validates these ranges against the latest committed prefix, then atomically publishes the data files
and updated progress.

For reads, Paimon captures immutable `MemReadView` objects and combines them with the selected disk
snapshot through `RealtimeSplit`. The append reader concatenates disk readers with memory readers
whose offsets are newer than that snapshot's committed partition-bucket boundary. Refreshing a
committed snapshot advances the shared context and allows fully covered sealed segments to be
reclaimed, while existing query plans continue to pin their original memory views.

The main changes are:

- Add pluggable `MemIndexer`, `MemIndexerFactory`, `RealtimeSegmentHandle`, and
  `MemReadView` interfaces.
- Provide a default Arrow-backed memory indexer.
- Add `RealtimeContext` to own partition-bucket indexers and share them between writers and
  readers.
- Add `PrepareCommitWithProgress` and `CommitWithProgress`.
- Seal memory segments during prepare commit and flush them through existing Paimon file writers.
- Persist versioned per-partition-bucket committed progress in `metadata/<uuid>.offsets` files
  referenced by snapshot properties.
- Restore the next internal offset from the latest committed snapshot.
- Add real-time table scans, `RealtimeSplit`, and append disk-memory union reads.
- Capture immutable memory views so a query remains stable across concurrent writes and refresh.
- Reclaim sealed memory after committed progress advances.
- Support projection, partition and bucket filters, framework-level exact predicate filtering, and
  optional plugin predicate pushdown.
- Preserve existing rolling, file-format, file-index, statistics, and physical-field handling.

The current implementation supports streaming writes and latest-snapshot batch scans for
fixed-bucket append tables. Primary-key tables, deletion vectors, data evolution, streaming scans,
scan-limit pushdown, and global-index splits are not included in this PR.

### Tests

Added unit coverage for:

- Arrow memory-indexer write, seal, query-boundary, reclamation, and close behavior;
- real-time context indexer reuse, immutable read views, offset restoration, and monotonic committed
  progress;
- versioned offset JSON serialization, validation, offset-file I/O, and snapshot-property merging;
- unordered commit-progress sorting and contiguous-prefix validation;
- real-time writer range tracking and API validation;
- predicate field binding and writer memory-manager behavior.

Added 14 integration tests covering:

- append write, commit, and disk read;
- rolling files while preserving continuous commit progress;
- ordering commit messages prepared out of order;
- memory-only reads before prepare commit;
- disk and building-memory union reads;
- projection and exact predicate filtering across disk and memory;
- Parquet disk predicate pushdown without memory filtering;
- committed snapshot refresh and memory reclamation;
- immutable query plans across refresh;
- repeated write, commit, read, and refresh cycles;
- concurrent write, prepare commit, read, commit, and refresh;
- disk-memory combinations across multiple partitions;
- independent progress restoration across multiple buckets;
- internal offset restoration from a committed snapshot.

### API and Format

This PR adds the following public API concepts:

- `RealtimeContext`, `RealtimePartitionBucket`, and `RealtimeOffsetMap`;
- `MemIndexer`, `MemIndexerFactory`, `RealtimeSegmentHandle`, and `MemReadView`;
- `RealtimeWriteBatch`, `MemQueryContext`, and `RealtimeCommitProgress`;
- `WriteContextBuilder::WithRealtimeContext`;
- `ScanContextBuilder::WithRealtimeContext`;
- `FileStoreWrite::PrepareCommitWithProgress`;
- `FileStoreWrite::RefreshCommittedSnapshot`;
- `FileStoreCommit::CommitWithProgress`.

The feature is enabled only when a `RealtimeContext` is supplied. Existing write, commit, and
disk-only read paths remain unchanged otherwise.

This PR adds a versioned offset metadata file referenced by the `realtime.offsets` snapshot property. The file stores the largest committed internal offset for each logical partition and bucket; data files and progress metadata are published by the same snapshot commit.

Real-time commits currently fail directly on snapshot conflicts instead of retrying with stale
progress. Failure recovery and idempotent commit retry can be addressed separately.

### Documentation

This is a new opt-in feature. The public interfaces contain API documentation, and the design is
tracked in #158.

### Generative AI tooling


src/paimon/core/operation/abstract_file_store_write.cpp
```diff
@@ -261,7 +266,74 @@ Result<std::vector<std::shared_ptr<CommitMessage>>> AbstractFileStoreWrite::Prep
...
+        for (const auto& [partition, buckets] : writers_) {
+            for (const auto& [bucket, container] : buckets) {
+                writer_snapshots.push_back(
+                    WriterSnapshot{partition, bucket, container.total_buckets, container.writer});
+            }
+        }
+    }
+
+    std::vector<RealtimeCommitProgress> result;
+    for (const WriterSnapshot& snapshot : writer_snapshots) {
```
HaHaJeff, line 301: How should realtime writes recover if `PrepareCommitWithProgress` fails for
one bucket?

Terminating all bucket writers would make the failure scope too large for a
realtime workload. Could we preserve the prepared state of successful buckets
and retry or recreate only the failed bucket writer?

lxy-9602: Thanks for raising this concern. Bucket-level failure isolation is indeed valuable for long-running real-time workloads.

We checked the existing Paimon write path, including the Java Spark integration. Currently, `FileStoreWrite.prepareCommit` does not provide partial-success semantics across buckets. If preparing any bucket fails, the exception is propagated through `TableWrite` to the Spark data writer, causing the corresponding Spark task attempt to fail. Commit messages from buckets prepared earlier in the same call are not returned as independently recoverable results.

More generally, Paimon’s current read and write APIs do not define a contract for partial failure and partial recovery within one operation. For the first phase, we would prefer to keep the real-time implementation consistent with this existing behavior rather than introduce a separate recovery model only for `PrepareCommitWithProgress`.

For your use case, one practical approach is to use one `FileStoreWrite` instance per bucket. This keeps the failure scope local to that bucket: a failed bucket writer can be recreated and replayed independently, while writers for other buckets remain unaffected. A higher-level coordinator can still collect their commit messages and commit them under the desired snapshot boundary.

Bucket-level prepared-state preservation and retry could be considered as a future enhancement, but it would require a broader partial-recovery contract across Paimon’s read and write paths.

HaHaJeff: Thanks for the clarification. This approach works for us.

We can use one `FileStoreWrite` per partition/bucket, recover a failed writer
independently, and let a higher-level coordinator collect the progress and
commit it under the same snapshot boundary.

include/paimon/write_context.h
```diff
@@ -40,6 +40,9 @@ class MemoryPool;
 /// @see WriteContextBuilder
 class PAIMON_EXPORT WriteContext {
  public:
+    /// Constructs a write context with an optional real-time write context.
```
zjw1111: need this comments?

test/inte/realtime_write_inte_test.cpp
```diff
@@ -0,0 +1,1152 @@
...
+        ASSERT_OK_AND_ASSIGN(std::vector<Row> read_after_refresh, ReadRows(realtime_context));
+        ASSERT_EQ(read_before_refresh, read_after_refresh);
+        ASSERT_OK_AND_ASSIGN(uint64_t memory_usage_after_refresh,
+                             GetRealtimeMemoryUsage(realtime_context));
+        ASSERT_EQ(0, memory_usage_after_refresh);
+    }
+    ASSERT_OK(writer->Close());
+}
+
+TEST_F(RealtimeWriteInteTest, TestConcurrentWritePrepareCommitReadAndRefresh) {
```
zjw1111, line 815: a little finding: this case is so long

src/paimon/core/operation/commit/realtime_commit_properties.cpp
```diff
@@ -0,0 +1,255 @@
...
+        return offset_;
+    }
+
+ private:
+    std::map<std::string, std::string> partition_;
+    int32_t bucket_ = -1;
+    int64_t offset_ = -1;
+};
+
+class OffsetsJson {
```
zjw1111: override `Jsonizable`?

zjw1111 (commented):
Re-checked the previously reported issues against the current head `5e65d206`. Five of them are still present and confirmed at this head; the inline comments below describe the problems only.

By impact: the indexer lifetime issue (`realtime_append_only_writer.cpp`) and the duplicate-offset issue (`realtime_context.cpp`) both affect correctness and both come from the same missing ownership model for `MemIndexer` — the lifetime one is reachable through a plain `Close()` call. The absolute-path and `ArrowSchema*` contract issues concern persisted metadata format and public API surface respectively. The missing snapshot id is an API gap.

For the record, the following previously reported problems are fixed at this head: the double-applied offset in `CommitBatchReader`, the commit-time window between continuity validation and offset merging, and the inconsistent state left by a partially failed `AdvanceCommittedProgress`.

src/paimon/core/realtime/realtime_context.cpp
```diff
@@ -0,0 +1,194 @@
...
+        int64_t initial_offset = 0;
+        auto offset_iter = committed_offsets_.find(key);
+        if (offset_iter != committed_offsets_.end()) {
+            if (offset_iter->second == std::numeric_limits<int64_t>::max()) {
+                if (write_schema) {
+                    ArrowSchemaRelease(write_schema.get());
+                }
+                return Status::Invalid("real-time offset has reached INT64_MAX");
+            }
+            initial_offset = offset_iter->second + 1;
```
zjw1111, line 60: The initial offset handed to a writer is derived only from the committed offsets. It ignores segments that a reused indexer has already sealed but not yet committed.

`SealForCommit()` resets `building_range_` (`arrow_mem_indexer.cpp:296`), and `Write()` only checks continuity against `building_range_` (`arrow_mem_indexer.cpp:269-270`), so there is no continuity check across a seal boundary. `MemIndexer` also exposes no accessor for the highest offset it currently holds, and the context does not track it.

Consequence: when a new writer is created against a context that still holds a sealed-but-uncommitted segment for the same partition-bucket — the cross-writer-instance context reuse that `realtime_context.h:87-91` describes — the writer restarts at `committed + 1` and the overlapping range is accepted silently. The indexer then holds both the sealed `[x..y]` segment and a new building range starting at `x`; `AcquireReadView()` merges the two, so memory reads return duplicated rows. At commit time the second range fails the continuity check in `realtime_commit_properties.cpp:215-220`, far from the root cause; if the first commit message was abandoned, the already-flushed data files are left uncommitted.

For scope: this is not reachable inside a single `FileStoreWrite` instance, since realtime writers are never evicted (the legacy `PrepareCommit` is rejected at `abstract_file_store_write.cpp:150-153` and the memory manager is a no-op). Two live writers sharing one context are also caught by the `building_range_` check. The seal boundary is the only unguarded case.

src/paimon/core/realtime/realtime_append_only_writer.cpp
```diff
@@ -0,0 +1,214 @@
...
+}
+
+Status RealtimeAppendOnlyWriter::Sync() {
+    return file_writer_->Sync();
+}
+
+Status RealtimeAppendOnlyWriter::Close() {
+    {
+        std::lock_guard<std::mutex> lock(mem_indexer_mutex_);
+        PAIMON_RETURN_NOT_OK(mem_indexer_->Close());
```
zjw1111: The writer closes a `MemIndexer` that `RealtimeContext` owns and shares with scans and with later writers.

`FileStoreWrite::Close()` closes every writer and clears the writer map (`abstract_file_store_write.cpp:337-348`) — a normal lifecycle call, not an error path. `ArrowMemIndexer::Close()` sets `closed_` and drops all batches (`arrow_mem_indexer.cpp:377-385`), after which `Write`, `SealForCommit`, `AcquireReadView` and `AdvanceCommittedOffset` all fail with `mem indexer is closed` (`:266`, `:287`, `:314`, `:353`). The indexer stays in `RealtimeContext::indexers_`, which is never erased, and `GetOrCreateMemIndexer` returns the cached entry without checking `closed_` (`realtime_context.cpp:62-68`).

So after a writer is closed, the shared context is permanently degraded for that partition-bucket: scans through the same context fail in `AcquireReadViews`, `RefreshCommittedSnapshot` fails inside `AdvanceCommittedOffset`, and a writer newly created from the same context can never write, with no way to recover.

This is uncovered: every case in `test/inte/realtime_write_inte_test.cpp` calls `writer->Close()` as its last statement, and both restore tests create a fresh `RealtimeContext` for the second writer, so "close a writer, then keep using the same context" is never exercised.

include/paimon/file_store_commit.h
```diff
@@ -71,6 +72,21 @@ class PAIMON_EXPORT FileStoreCommit {
...
+    /// Entries for each partition-bucket must form a contiguous range beginning after the offset
+    /// recorded by the latest committed snapshot. Input entries may be unordered; this method
+    /// orders them by partition, bucket, and offset before validating continuity. The resulting
+    /// snapshot atomically publishes the data files and the updated offset map.
+    ///
+    /// @param realtime_commits Commit messages and inclusive offset ranges to commit.
+    /// @param commit_identifier Identifier of the streaming commit operation.
+    /// @param watermark Optional event-time watermark.
+    /// @return Status indicating the success or failure of the commit operation.
+    virtual Status CommitWithProgress(const std::vector<RealtimeCommitProgress>& realtime_commits,
```
zjw1111: `CommitWithProgress` returns only `Status`, but `FileStoreWrite::RefreshCommittedSnapshot(int64_t snapshot_id)` needs the id of the snapshot this call produced. The API offers no way to obtain it, so the caller has to query `LatestSnapshot()` again and cannot tell whether the result is its own snapshot — `test/inte/realtime_write_inte_test.cpp` does exactly this. There is no reliable link between a commit and the refresh that is supposed to follow it.

In the current implementation the offsets map is carried forward by every commit path (`realtime_commit_properties.cpp:191-201`), so refreshing with a later snapshot is usually harmless. `RollbackTo` is the exception: the new snapshot is built from the target snapshot's properties (`file_store_commit_impl.cpp:345-354`), so the offsets map can move backwards, and `AdvanceCommittedProgress` then rejects every later refresh with `committed snapshot cannot move backwards`.

src/paimon/core/operation/commit/realtime_commit_properties.cpp
```diff
@@ -0,0 +1,255 @@
...
+    }
+    if (offsets_directory.empty()) {
+        return Status::Invalid("real-time offsets directory is empty");
+    }
+    std::string uuid;
+    if (!UUID::Generate(&uuid)) {
+        return Status::Invalid("fail to generate uuid for real-time offsets file");
+    }
+    PAIMON_RETURN_NOT_OK(file_system->Mkdirs(offsets_directory));
+    std::string path = PathUtil::JoinPath(offsets_directory, uuid + ".offsets");
```
zjw1111, line 246: The offsets file reference written into snapshot properties is an absolute path, and `ReadOffsets` reads it back directly (`:174`). This couples snapshot metadata to the current table location: after the table directory is relocated, restored elsewhere, or the mount point changes, old snapshots can no longer resolve their offsets file.

Every other snapshot metadata reference in this repo stores a bare file name and resolves it through `FileStorePathFactory` — see `ToManifestListPath` in `src/paimon/core/utils/file_store_path_factory.h:134`. This is the only reference that does not follow that convention, and it is persisted format, so the compatibility cost grows once it ships.

include/paimon/realtime/mem_indexer.h
```diff
@@ -0,0 +1,165 @@
...
+class PAIMON_EXPORT MemIndexerFactory {
+ public:
+    virtual ~MemIndexerFactory() = default;
+
+    /// Creates an indexer configured with the supplied schema, options, and memory pool.
+    /// @param write_schema Complete table write schema.
+    /// @param options Effective table options available to the indexer.
+    /// @param memory_pool Memory pool provided by the write context.
+    virtual Result<std::shared_ptr<MemIndexer>> Create(
+        ::ArrowSchema* write_schema, const std::map<std::string, std::string>& options,
```
zjw1111: The ownership and lifetime contract of `write_schema` is unspecified: whether ownership transfers, whether the implementation must consume or release it, whether it may retain the pointer, and who releases it when `Create` fails.

The two sides currently disagree in a way that only works by accident. The default factory consumes the schema through `arrow::ImportSchema` (`src/paimon/core/realtime/arrow_mem_indexer_factory.cpp`), while `RealtimeContext` calls `ArrowSchemaRelease` unconditionally on every path, including the success path (`src/paimon/core/realtime/realtime_context.cpp:55-57`, `:64-66`, `:71-73`). That is only safe because `ArrowSchemaRelease` is a no-op once `release` is null, so the effective contract is "released unless already released" and an implementation cannot tell whether the schema is still live after `Create` returns. An implementation that retains the raw pointer gets a use-after-release, since the context releases it as soon as `Create` returns.

One level up the ownership is explicit — `RealtimeContext::GetOrCreateMemIndexer` takes `std::unique_ptr<ArrowSchema>` — and that information is discarded at the plugin boundary.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #166, opened 2026-08-03
fix(parquet): reuse RowGroupPageIndexReader in FileReaderWrapper layer

zhf999:
### Purpose

Arrow caches the file-level `PageIndexReader` (`ParquetFileReader::GetPageIndexReader()`), but
`PageIndexReader::RowGroup(i)` builds a brand-new `RowGroupPageIndexReaderImpl` on every call, and
the OffsetIndex/ColumnIndex byte buffer (`offset_index_buffer_`) is cached **inside that instance**.
So every new `RowGroup(i)` call re-issues a real `ReadAt` for the whole page-index region of the row
group. These reads are not covered by the data-page pre-buffer/range cache, so they are genuine
small I/Os against the underlying file.

Previously each stage of a page-filtered read created its own row-group reader, so for one
partially-matched row group the page index region was read up to three times:

1. `FileReaderWrapper::CollectPreBufferRanges()` -> `ComputePageRanges()`
2. `FileReaderWrapper::NextPageFiltered()` -> `ComputePageRanges()`
3. `PageFilteredRowGroupReader::ReadFilteredRowGroup()` (for the per-field `OffsetIndex` lookups)

This PR makes the row-group page index reader an explicit, reusable input:

- `FileReaderWrapper` gains `GetRowGroupPageIndexReader(row_group_index)` plus a
  `row_group_page_index_readers_` map that memoizes one reader per row group (including the
  "no page index" negative result), so all stages share the same page-index buffers.
- `PageFilteredRowGroupReader::ComputePageRanges()` and
  `PageFilteredRowGroupReader::ReadFilteredRowGroup()` now take the
  `std::shared_ptr<::parquet::RowGroupPageIndexReader>` from the caller instead of deriving it
  internally from `ParquetFileReader::GetPageIndexReader()`.

Behavior is unchanged: files without a page index still pass a null reader and fall back to the
existing whole-column-chunk path; read results and computed ranges are identical. The only
difference is that the OffsetIndex region of each row group is read and deserialized once per file
reader instead of once per stage.

### Tests

No.

### API and Format

No changes.

### Documentation

No changes.

### Generative AI tooling

Generated-by: No.

### DIFF
--- a/src/paimon/format/parquet/column_index_filter.h
+++ b/src/paimon/format/parquet/column_index_filter.h
@@ -64,9 +64,8 @@ class ColumnIndexFilter {
     /// @return RowRanges that may contain matching rows.
     static Result<RowRanges> CalculateRowRanges(
         const std::shared_ptr<Predicate>& predicate,
-        const std::shared_ptr<::parquet::PageIndexReader>& page_index_reader,
-        const std::map<std::string, int32_t>& column_name_to_index, int32_t row_group_index,
-        int64_t row_group_row_count);
+        const std::shared_ptr<::parquet::RowGroupPageIndexReader>& rg_page_index_reader,
+        const std::map<std::string, int32_t>& column_name_to_index, int64_t row_group_row_count);
 
  private:
     /// Visit a predicate and calculate row ranges.
--- a/src/paimon/format/parquet/column_index_filter.h
+++ b/src/paimon/format/parquet/column_index_filter.h
@@ -57,9 +57,8 @@ class ColumnIndexFilter {
 
     /// Calculate row ranges based on predicate and column indices.
     /// @param predicate The predicate to evaluate.
-    /// @param page_index_reader The page index reader for the file.
+    /// @param rg_page_index_reader The page index reader of target row group for the file.
     /// @param column_name_to_index Map from column name to column index.
-    /// @param row_group_index The row group index to filter.
     /// @param row_group_row_count The number of rows in the row group.
     /// @return RowRanges that may contain matching rows.
     static Result<RowRanges> CalculateRowRanges(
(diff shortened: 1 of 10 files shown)


src/paimon/format/parquet/page_filtered_row_group_reader.h
```diff
@@ -56,20 +56,24 @@ class PageFilteredRowGroupReader {
...
     /// @param arrow_file_reader The Arrow FileReader for ColumnReader tree creation
     /// @return A RecordBatchReader streaming the filtered rows.
     static Result<std::unique_ptr<arrow::RecordBatchReader>> ReadFilteredRowGroup(
         const TargetRowGroup& target_row_group, const std::vector<int32_t>& column_indices,
         const ::arrow::io::CacheOptions& cache_options, bool pre_buffered,
         const std::vector<::arrow::io::ReadRange>& page_ranges, int64_t max_chunksize,
-        std::shared_ptr<::arrow::MemoryPool> pool, ::parquet::arrow::FileReader* arrow_file_reader);
+        std::shared_ptr<::arrow::MemoryPool> pool,
+        std::shared_ptr<::parquet::RowGroupPageIndexReader> row_group_page_index_reader,
+        ::parquet::arrow::FileReader* arrow_file_reader);
```
lxy-9602: I’m a bit curious whether `row_group_page_index_reader` and `arrow_file_reader` are both intended to be output parameters. If so, should we make them consistent and use `shared_ptr` for both before putting them into the pool?

src/paimon/format/parquet/column_index_filter.h
lxy-9602, line 61: Please adjust comments.

src/paimon/format/parquet/page_filtered_row_group_reader.h
zhf999: Since `arrow_file_reader` is a `unique_ptr` hold by FileReaderWrapper, it cannot be transfer into `shared_ptr`, we use raw pointer here.

The order of those parameter has been resorted, output parameter `pool` and `arrow_file_reader` have been just at the end of this function.

src/paimon/format/parquet/column_index_filter.h
zhf999: Done.

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM. @zhf999, please fix the CI failure.

SteNicholas (approved):

PR #172, opened 2026-08-03
fix(parquet): allow reading nested list/map columns whose leaf types differ only in representation

zhf999:
### Purpose

Reading a Parquet column of type `list<struct<... timestamp(us, <tz>)>>` (or a `map` whose value
field is not literally named `value`) failed with:

```
Parquet does not support partial projection inside list/map: src ... vs target ...
```

even though no partial projection was requested.

`ParquetFileBatchReader::CollectLeafIndices()` guards against pruning fields inside LIST/MAP by
requiring `read_type->Equals(file_type)`. That guard is too strict, because the two types being
compared come from different layers:

- `file_type` comes from `FileReaderWrapper::GetSchema()`, i.e. the raw Arrow schema derived from
  the Parquet metadata.
- `read_type` is the Paimon-facing read schema. Even the default self-read path in
  `ParquetFileBatchReader::Create()` feeds back `GetFileSchema()`, which has already been rewritten
  by `ParquetTimestampConverter::AdjustTimezone()`.

So `Equals()` rejects pure representation differences that the reader is designed to reconcile
later in `NextBatch()` via `ParquetTimestampConverter::NeedCastArrayForTimestamp()` /
`CastArrayForTimestamp()`:

1. **Timezone**: the Parquet reader always reports LTZ timestamps as `UTC`, while Paimon exposes
   them in the local timezone.
2. **Unit**: the Parquet writer has no SECOND timestamp, so a SECOND column is stored as MILLI and
   cast back to SECOND on read.
3. **Child field names / nullability**: a map value field named `attrs` in the file versus the
   `value` produced by `arrow::map()`, a `list` element field name, or a differing nullable flag.

This PR makes the guard compare only the *projection shape*:

- New file-local helper `HasSameNestedProjectionShape()` in
  `src/paimon/format/parquet/parquet_file_batch_reader.cpp`:
  - STRUCT: same field count, same field names, recursively same shape (nullability ignored).
  - LIST: compares `value_type()` only, ignoring the element field name and nullability.
  - MAP: compares `key_type()` and `item_type()` only, ignoring the key/value field names.
  - TIMESTAMP leaves: accepted when the units are equal, or when the file is MILLI and the read
    type is SECOND — exactly the conversions `ParquetTimestampConverter` supports. Timezone
    differences are always accepted and left to the cast path.
  - All other leaves keep the strict `Equals()` check, and a nested-versus-atomic mismatch is
    rejected.
- `CollectLeafIndices()` splits the shared LIST/MAP branch and now recurses through
  `ListType::value_type()` and `MapType::key_type()`/`item_type()` instead of iterating
  `file_type->field(i)`. This also fixes a latent leaf-index bug: `arrow::MapType::num_fields()`
  is 1 (the `entries` struct), so the old code fell into the STRUCT branch and matched map children
  **by name**; when the file's value field name differed from the read schema's, it took the
  `SkipLeafIndices()` path and silently produced a wrong leaf index list.

Fail-fast behavior for genuinely unsupported cases is preserved: pruning a field from a struct
inside a list/map, incompatible leaf types, and unsupported timestamp unit conversions all still
return `Status::Invalid`.

### Tests

New case in target `paimon-parquet-format-test`:

- `ParquetFileBatchReaderTest.TestNestedListTimestampTimezoneAndMapFieldName`
  (`src/paimon/format/parquet/parquet_file_batch_reader_test.cpp`) writes
  `list<struct<key: utf8, attrs: map<utf8, utf8 ("attrs" value field)>, updated_at: timestamp(us,
  Asia/Shanghai)>>` under a `TimezoneGuard`, including empty maps, null maps and null timestamps,
  then asserts:
  - the full round trip now succeeds and matches the expected array (previously `Status::Invalid`);
  - pruning `updated_at` from the nested struct still fails with
    `"Parquet does not support partial projection inside list/map"`;
  - reading `updated_at` as `utf8` still fails;
  - reading `updated_at` as `timestamp(ns, ...)` (an unsupported unit conversion) still fails.

### API and Format

No changes.

### Documentation

No changes.

### Generative AI tooling

No.

### DIFF
--- a/src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
+++ b/src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
@@ -738,6 +738,64 @@ TEST_F(ParquetFileBatchReaderTest, TestNestedListTimestampTimezoneAndMapFieldNam
                         "Parquet does not support partial projection inside list/map");
 }
 
+TEST_F(ParquetFileBatchReaderTest, TestNestedTimestampSecondReadFromMilliFile) {
+    const std::string timezone = "Asia/Shanghai";
+    paimon::test::TimezoneGuard timezone_guard(timezone);
+
+    // Parquet has no second-precision timestamp, so the writer stores second timestamps as
+    // milliseconds. Reading them back with a second-precision schema must cast milli to second,
+    // including for timestamp leaves nested inside list/struct/map.
+    auto event_type = arrow::struct_({
+        arrow::field("name", arrow::utf8()),
+        arrow::field("ts_sec", arrow::timestamp(arrow::TimeUnit::SECOND)),
+        arrow::field("ts_tz_sec", arrow::timestamp(arrow::TimeUnit::SECOND, timezone)),
+    });
+    auto schema = arrow::schema({
+        arrow::field("events", arrow::list(arrow::field("element", event_type))),
+        arrow::field("marks", arrow::map(arrow::utf8(), arrow::timestamp(arrow::TimeUnit::SECOND))),
+    });
+
+    const std::string data_json = R"([
+        [[ ["e-1", "2026-07-16 12:00:01", "2026-07-16 12:00:02"] ],
+         [["begin", "2026-07-16 12:00:03"]]],
+        [[ ["e-2", "2026-07-16 12:00:04", null],
+           ["e-3", null, "2026-07-16 12:00:05"] ], []],
+        [[null], null]
+    ])";
+    auto write_array = std::dynamic_pointer_cast<arrow::StructArray>(
+        arrow::ipc::internal::json::ArrayFromJSON(arrow::struct_(schema->fields()), data_json)
+            .ValueOrDie());
+    WriteArray(file_path_, write_array, schema, /*write_batch_size=*/write_array->length(),
+               /*enable_dictionary=*/false, /*max_row_group_length=*/write_array->length());
+
+    auto parquet_batch_reader =
+        PrepareParquetFileBatchReader(file_path_, schema, /*predicate=*/nullptr,
+                                      /*selection_bitmap=*/std::nullopt, /*batch_size=*/2);
+
+    // The nested second timestamps are physically stored as milliseconds in the file.
+    ASSERT_OK_AND_ASSIGN(auto c_file_schema, parquet_batch_reader->GetFileSchema());
+    auto file_schema = arrow::ImportSchema(c_file_schema.get()).ValueOr(nullptr);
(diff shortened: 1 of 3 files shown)


src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
```diff
@@ -646,6 +646,98 @@ TEST_F(ParquetFileBatchReaderTest, TestReadSchemaWithMapSelectedKeysMetadata) {
         << "expected: " << expected_array->ToString() << "\nactual: " << result_array->ToString();
 }
 
+TEST_F(ParquetFileBatchReaderTest, TestNestedListTimestampTimezoneAndMapFieldName) {
+    const std::string timezone = "Asia/Shanghai";
```
lxy-9602, line 650: Could we extend the test coverage for both newly supported nested timestamp projection cases?

First, please add a focused `ParquetFileBatchReaderTest` for a nested MILLI -> SECOND conversion, for example with `list<struct<..., timestamp(SECOND)>>`. The existing test only covers the nested MICRO timezone-only case and does not exercise the unit-conversion branch allowed by this guard.

In addition, please add E2E write-and-read cases for both scenarios in `write_and_read_inte_test.cpp`:

1. nested MILLI -> SECOND unit conversion;
2. nested MICRO timestamp with a timezone-only difference.

This would verify both the reader-level conversion logic and the complete Parquet write/read path.

test/inte/write_and_read_inte_test.cpp
```diff
@@ -806,6 +806,161 @@ TEST_P(WriteAndReadInteTest, TestPkTimestampType) {
...
+    ])";
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatch> batch,
+                         TestHelper::MakeRecordBatch(arrow::struct_(fields), data,
+                                                     /*partition_map=*/{}, /*bucket=*/0, {}));
+    ASSERT_OK(helper->WriteAndCommit(std::move(batch), /*commit_identifier=*/0,
+                                     /*expected_commit_messages=*/std::nullopt));
+
+    if (file_format == "parquet") {
+        // Verify the precision downgrade really happened, so the read below is exercising an
+        // actual milli to second conversion instead of a no-op.
```
lxy-9602: Please rm file verify in E2E tests.

zhf999: Done.

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #176, opened 2026-08-04
fix: support nullable map key schemas and reject null key data

lxy-9602:
### Purpose

No Linked issue.

This PR restores nullable MAP-key schema compatibility that was removed by [alibaba/paimon-cpp#226](https://github.com/alibaba/paimon-cpp/pull/226).

The changes:

- Accept nullable MAP key declarations when parsing existing Paimon schemas.
- Normalize parsed MAP key fields to non-nullable Arrow fields because Apache Arrow does not support nullable MAP keys.
- Remove the general schema validation requiring MAP keys to be explicitly declared as `NOT NULL`.
- Continue rejecting nullable MAP keys for the shared-shredding storage layout.
- Validate Arrow arrays at the ORC and Parquet format-reader boundaries so data containing an actual null MAP key returns `Status::Invalid` instead of triggering an Arrow abort.
- Restore nullable MAP key declarations in existing test schemas.

### Tests

- Parsing a nullable MAP key declaration and normalizing it for Arrow.
- Creating table schemas containing nullable MAP key declarations.
- Rejecting nullable MAP keys when shared-shredding is enabled.
- Reading the nullable-MAP-key ORC fixture with `read.batch.size=1`:
  - The first two rows are read and verified successfully.
  - The third row returns `Map array keys array should have no nulls`.
  

### API and Format

Schema compatibility is relaxed so existing schemas may declare nullable MAP keys. An error is returned only when the physical data actually contains a null MAP key. Shared-shredding continues to require non-nullable MAP keys.

### Documentation
Updated the data type and schema documentation to describe nullable MAP key compatibility and read-time validation behavior.

### Generative AI tooling


duanyyyyyyy:
LGTM

lszskye (approved):
+1

SteNicholas (approved):
LGTM.

PR #181, opened 2026-08-04
feat(build): support Linux aarch64 builds

### LINKED ISSUE #171: [Feature] Support aarch64/arm64 architecture
### Search before asking

### Motivation

Paimon C++ currently only builds and is validated on x86_64. aarch64/arm64 has become a mainstream server and desktop platform (Apple Silicon, AWS Graviton, Alibaba Cloud Yitian, Ampere, etc.), and users deploying Paimon readers/writers on these platforms cannot use the library today.

Concretely, the following gaps block first-class aarch64/arm64 support:

1. **No architecture detection in the build system.** `cmake_modules/SetupCxxFlags.cmake` consumes `PAIMON_CPU_FLAG` (with `x86`, `ppc`, and `armv8` branches), but nothing in the repository ever sets it from `CMAKE_SYSTEM_PROCESSOR`, so the `armv8` tuning path (`PAIMON_ARMV8_ARCH_FLAG`, CRC feature detection) is effectively dead code and the x86 SIMD defines (`PAIMON_HAVE_SSE4_2`/`AVX2`/`AVX512`) are not guarded by an actual architecture check.
2. **Third-party dependencies with x86_64-only artifacts.** In `cmake_modules/ThirdpartyToolchain.cmake`, the lumina dependency hardcodes `artifacts/cpp/linux-x86_64/install-root/...` paths, so the build cannot link it on aarch64. (JindoSDK already resolves `aarch64` artifacts for both Linux and macOS, so it is not a blocker.) The tantivy-based global index is built from Rust via Corrosion and needs verification that the toolchain targets aarch64 correctly.
3. **No aarch64 CI coverage.** All workflows under `.github/workflows/` run on x86_64 GitHub-hosted runners (`ubuntu-22.04`/`ubuntu-24.04`), so architecture-specific regressions are invisible. There is at least one known class of portability issues already noted in the codebase: `char` is unsigned by default on aarch64 (see the workaround comment in `cmake_modules/BuildUtils.cmake`), which can silently change behavior of code assuming signed `char`.
4. **No arm64 release artifacts.** The packaging/release tooling only produces x86_64 outputs, and the documentation does not state supported platforms.

### Solution

Proposed incremental plan:

1. Derive `PAIMON_CPU_FLAG` from `CMAKE_SYSTEM_PROCESSOR` (`x86` / `armv8` / `ppc`) in the CMake setup, so the existing armv8 branch in `SetupCxxFlags.cmake` becomes reachable and x86 SIMD flags are only applied on x86.
2. Audit dependencies in `ThirdpartyToolchain.cmake`: select aarch64 artifacts where available (as already done for JindoSDK), build from source or make the dependency optional where no aarch64 artifact exists (e.g. lumina), and verify the Corrosion/Rust tantivy build on aarch64.
3. Add an aarch64 CI job using GitHub's arm64 hosted runners (`ubuntu-24.04-arm`) that runs the standard build and unit test suite, and fix any test failures uncovered (char signedness, alignment, intrinsics).
4. Extend the release/packaging scripts to produce linux-aarch64 (and macos-arm64) artifacts, and document the supported platform matrix in `docs/source/building.rst`.

### Anything else?

_No response_

SteNicholas:
### Purpose

Linked issue: close #171

Paimon C++ only built and was validated on x86_64: the architecture-specific build logic keyed on `PAIMON_CPU_FLAG` and `PAIMON_ARMV8_ARCH`, neither of which was ever defined, so the armv8 tuning branch was dead code and Arm64 got no `-march=` flag at all. This change resolves the target architecture once and keys every architecture-specific decision on it. Running the suite on `ubuntu-24.04-arm` then exposed two classes of target-dependent behavior — plain-`char` signedness, which the AArch64 Linux ABI decides differently, and undefined float-to-integer conversion — plus a gap in how a failure is reported.

Build system and CI:

- **`cmake_modules/TargetArchitecture.cmake`** derives `PAIMON_TARGET_PROCESSOR` and `PAIMON_TARGET_CPU_FAMILY` from `CMAKE_SYSTEM_PROCESSOR`, with no compiler probe so `cmake -P` can test it. An unrecognized processor gets no tuning flags rather than a configure error.
- **`PAIMON_AARCH64_MARCH`** selects the Arm64 `-march=` value, defaulting to `armv8-a`; an explicitly empty value passes no `-march` flag at all. A fallback covers `add_subdirectory()` consumers, for which the option does not exist.
- **The dead SIMD build logic is removed.** The SSE4.2 CRC32C kernel computed Castagnoli, not the zlib CRC-32 persisted in the SST block trailer and the B-tree global index, and it hung off the never-defined `PAIMON_SIMD_LEVEL`, so no build ever selected it — no on-disk value changes. The ppc AltiVec probe and the never-read `PAIMON_HAVE_NEON` / `PAIMON_HAVE_ARMV8_CRC` / `PAIMON_HAVE_ARMV8_CRYPTO` definitions go with it.
- **Lumina** is prebuilt for `linux-x86_64` only: configuring it elsewhere now fails with an explicit error, and CI disables it off x86_64.
- **`build_and_package.sh`** names the artifact after the platform, e.g. `output/paimon-cpp-linux-x86_64.tar.gz`. `--platform` overrides the label alone (cross compilation is not wired up end to end); `--print-name` prints the resolved name and exits.
- **CI** runs the same matrix on both architectures — Debug and Release, GCC and Clang, ASan/UBSan and TSan — with clang-tidy the one exception since it analyses the source. Job names now carry the architecture (`gcc-debug-x86_64`, `gcc-debug-aarch64`, …), a `script-tests` job is added, and `.asf.yaml` is renamed and extended in lockstep so all jobs stay required checks. Release verification gains an aarch64 entry in its own workflow.

Defects the aarch64 run exposed:

- **`BinaryString::NumBytesForFirstByte` classified a UTF-8 leading byte by its sign.** With `char` unsigned on aarch64, every byte was taken for single-byte ASCII, so `NumChars`, `Substring` and `IndexOf` counted bytes instead of characters — `NumChars("Paimon中文社区")` returned 18 rather than 10. The byte is now classified as `uint8_t`, branch for branch equivalent to the signed form on x86-64, so x86-64 results are unchanged.
- **`ToUpperCase` and `ToLowerCase` passed bytes to `toupper` / `tolower` as plain `char`** — undefined behavior for bytes at or above 0x80 where `char` is signed, x86-64 included. They now go through `unsigned char`, on the fast path and in the `CppTo*Case` fallbacks, and a stray pre-loop `tolower` write in `ToUpperCase` is removed. Intended results do not change.
- **Min and max aggregation compared TINYINT variants as plain `char`**, so on aarch64 `min(-20, 10)` returned 10. They compare the signed value now, as does `VariantValueToString`, which printed 236 for -20.
- **The sanitizer build did not link on aarch64**: UBSan's overflow check on 128-bit multiplications calls `__muloti4`, which libgcc does not provide and Clang inlines on x86-64 but not on aarch64. The Clang UBSan build now links compiler-rt's builtins archive.
- **Converting an out-of-range or non-finite float to an integer was undefined behavior**, in `NumericPrimitiveCastExecutor` and in the Arrow kernel it delegated to. `JavaFloatingToIntegerCast` now defines it as Java does — `NaN` becomes 0, out-of-range saturates at the int32 bounds (int64 for BIGINT), then narrows to the target width by keeping the low bits — on both the literal and the array path, so stats and data stay consistent. **This changes results on x86-64**, where those conversions previously produced whatever the hardware did.
- **`asan_symbolize.py` truncated the failure report**: it decoded stdin with `surrogateescape` but re-opened stdout strictly, so a byte that is not valid UTF-8 broke the pipe — which is why the first aarch64 run reported only a single failure. Both streams now use `surrogateescape`.

### Tests

New script-mode tests, run by the `script-tests` CI job without a toolchain or a configured build tree:

- `ci/scripts/test_cmake_modules.sh` — processor-string mapping in `TargetArchitecture.cmake`, and the `PAIMON_AARCH64_MARCH` default, override and `add_subdirectory()` fallback.
- `ci/scripts/test_packaging_args.sh` — `build_and_package.sh` argument parsing, the default platform label, `--platform` validation, `--print-name` and the resulting archive name.
- `ci/scripts/test_asan_symbolize.sh` — invalid UTF-8 bytes round-trip unchanged and symbolization continues past them; each case was verified to fail against the unfixed script.

Updated unit tests (target `unittest`):

- `crc32c_test.cpp` pins the checksum to zlib CRC-32 reference values rather than to whichever kernel was compiled in.
- `cast_executor_test.cpp` covers all eight float/double to integer pairs with the same twelve inputs and expected Java results on both the literal and the array path, including `MAX`/`LOWEST`/`±INFINITY`/`NaN`, plus the empty-array, sliced-offset and null edges of the conversion that replaced Arrow's kernel.
- `field_min_max_agg_test.cpp` and `data_define_test.cpp` cover the negative TINYINT; two utils tests now read it back through `int8_t`.
- `binary_string_test.cpp` enables the four-byte UTF-8 case as `u8"\U0001F919"` (it was disabled as an ill-formed surrogate pair), adds the same code point to `TestSubstring` and `TestIndexOf`, and adds invalid UTF-8 bytes passing through case conversion unchanged. The existing multi-byte tests were the ones failing on aarch64; they are unchanged and pass there now.

The `gcc-debug-aarch64` and `gcc-release-aarch64` jobs run the full unit and integration suite on `ubuntu-24.04-arm`, the end-to-end verification that the library builds and passes on aarch64 in both configurations.

### API and Format

No change to the public API under `include/`, and no change to the storage format: the removed CRC32C kernel was never selected by any build, and `BinaryString::NumBytesForFirstByte` keeps its `char` parameter and its x86-64 results, returning those same values on aarch64 now.

One behavior change: casting a float or double whose truncated value does not fit an integer type now follows Java Paimon on every architecture. It was undefined behavior before, so x86-64 results for those inputs change — `CAST(MAX_FLOAT AS TINYINT)` yields -1 instead of 0, `CAST(300.9 AS TINYINT)` yields 44, and `CAST(NaN AS INT)` yields 0 instead of `INT_MIN`. Values that already fit are unaffected.

Build-facing changes for downstream users: a new `PAIMON_AARCH64_MARCH` option, the released archive is now named `paimon-cpp-<platform>.tar.gz` instead of `paimon-cpp.tar.gz` (with `--platform` / `--print-name` added to `build_and_package.sh`), and configuring with lumina enabled off `linux-x86_64` now fails explicitly instead of failing later at link time.

### Documentation

Yes. `docs/source/building.rst` gains a supported platform matrix and documents `PAIMON_AARCH64_MARCH` and the platform-labelled packaging. `docs/source/user_guide/read.rst` updates the type change support matrix: float/double to integer casts are now well defined and Java-consistent on every supported architecture. `docs/code-style.md` gains rules against relying on the signedness of plain `char` and on undefined float-to-integer conversions, pointing at `JavaFloatingToIntegerCast` as the policy Paimon applies.

### Generative AI tooling

Generated-by: Claude Opus 5 (1M context)

### DIFF
--- a/build_support/asan_symbolize.py
+++ b/build_support/asan_symbolize.py
@@ -8,6 +8,7 @@
 #
 #===------------------------------------------------------------------------===#
 import bisect
+import io
 import os
 import re
 import subprocess
@@ -333,8 +334,17 @@ class SymbolizationLoop(object):
     if sys.version_info[0] == 2:
       sys.stdout = os.fdopen(sys.stdout.fileno(), 'w', 0)
     else:
-      # Unbuffered output is not supported in Python 3
-      sys.stdout = os.fdopen(sys.stdout.fileno(), 'w')
+      # Test output is not guaranteed to be valid UTF-8, so pin both streams to the same
+      # encoding and round such bytes through surrogateescape, which passes them out
+      # unchanged instead of raising and truncating the rest of the test log. Both the
+      # encoding and the error handler have to be set on both streams: taking either from
+      # the locale or from PYTHONIOENCODING would let them disagree and re-encode the bytes.
+      # The buffers are rewrapped rather than reconfigured, which needs Python 3.7, and
+      # line buffering stands in for the unbuffered output Python 3 does not support.
+      sys.stdin = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8',
+                                   errors='surrogateescape')
+      sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding='utf-8',
+                                    errors='surrogateescape', line_buffering=True)
 
     while True:
       line = sys.stdin.readline()
--- a/cmake_modules/SetupCxxFlags.cmake
+++ b/cmake_modules/SetupCxxFlags.cmake
@@ -22,6 +22,28 @@ include(CheckCXXCompilerFlag)
 
 message(STATUS "System processor: ${CMAKE_SYSTEM_PROCESSOR}")
 
+include("${CMAKE_CURRENT_LIST_DIR}/TargetArchitecture.cmake")
+message(STATUS "Target processor: ${PAIMON_TARGET_PROCESSOR} "
+               "(CPU family: ${PAIMON_TARGET_CPU_FAMILY})")
+
+# Probe the architecture specific tuning flags. x86 deliberately gets none of its
+# own, so there is nothing to probe there: the baseline stays whatever the
+# toolchain defaults to, and no persisted value depends on it; see
+# src/paimon/common/utils/crc32c.h.
+if(PAIMON_TARGET_CPU_FAMILY STREQUAL "aarch64" AND NOT "${PAIMON_AARCH64_MARCH}" STREQUAL
+                                                   "")
+    set(PAIMON_AARCH64_MARCH_FLAG "-march=${PAIMON_AARCH64_MARCH}")
+    # check_cxx_compiler_flag() caches under a fixed name, so a reconfigure in the
+    # same build directory would otherwise reuse the old PAIMON_AARCH64_MARCH verdict.
+    if(NOT "${PAIMON_AARCH64_MARCH_FLAG}" STREQUAL "${PAIMON_PROBED_AARCH64_MARCH_FLAG}")
+        unset(CXX_SUPPORTS_AARCH64_MARCH CACHE)
+        set(PAIMON_PROBED_AARCH64_MARCH_FLAG
+            "${PAIMON_AARCH64_MARCH_FLAG}"
+            CACHE INTERNAL "Arm64 arch flag that CXX_SUPPORTS_AARCH64_MARCH refers to")
+    endif()
+    check_cxx_compiler_flag("${PAIMON_AARCH64_MARCH_FLAG}" CXX_SUPPORTS_AARCH64_MARCH)
+endif()
+
 # Support C11
 if(NOT DEFINED CMAKE_C_STANDARD)
     set(CMAKE_C_STANDARD 11)
@@ -225,58 +247,22 @@ if(BUILD_WARNING_FLAGS)
 endif(BUILD_WARNING_FLAGS)
 
 # Only enable additional instruction sets if they are supported
-if(PAIMON_CPU_FLAG STREQUAL "x86")
-    if(PAIMON_SIMD_LEVEL STREQUAL "AVX512")
-        if(NOT CXX_SUPPORTS_AVX512)
-            message(FATAL_ERROR "AVX512 required but compiler doesn't support it.")
-        endif()
-        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_AVX512_FLAG}")
-        add_definitions(-DPAIMON_HAVE_AVX512 -DPAIMON_HAVE_AVX2 -DPAIMON_HAVE_BMI2
-                        -DPAIMON_HAVE_SSE4_2)
-    elseif(PAIMON_SIMD_LEVEL STREQUAL "AVX2")
-        if(NOT CXX_SUPPORTS_AVX2)
-            message(FATAL_ERROR "AVX2 required but compiler doesn't support it.")
-        endif()
-        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_AVX2_FLAG}")
-        add_definitions(-DPAIMON_HAVE_AVX2 -DPAIMON_HAVE_BMI2 -DPAIMON_HAVE_SSE4_2)
-    elseif(PAIMON_SIMD_LEVEL STREQUAL "SSE4_2")
-        if(NOT CXX_SUPPORTS_SSE4_2)
-            message(FATAL_ERROR "SSE4.2 required but compiler doesn't support it.")
-        endif()
-        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_SSE4_2_FLAG}")
-        add_definitions(-DPAIMON_HAVE_SSE4_2)
-    endif()
-endif()
-
-if(PAIMON_CPU_FLAG STREQUAL "ppc")
-    if(CXX_SUPPORTS_ALTIVEC AND PAIMON_ALTIVEC)
-        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_ALTIVEC_FLAG}")
-    endif()
-endif()
-
-if(PAIMON_CPU_FLAG STREQUAL "armv8")
-    if(NOT CXX_SUPPORTS_ARMV8_ARCH)
-        message(FATAL_ERROR "Unsupported arch flag: ${PAIMON_ARMV8_ARCH_FLAG}.")
-    endif()
-    if(PAIMON_ARMV8_ARCH_FLAG MATCHES "native")
-        message(FATAL_ERROR "native arch not allowed, please specify arch explicitly.")
-    endif()
-    set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_ARMV8_ARCH_FLAG}")
-
-    add_definitions(-DPAIMON_HAVE_NEON)
-
-    if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS
-                                                "5.4")
-        message(WARNING "Disable Armv8 CRC and Crypto as compiler doesn't support them well."
-        )
+if(PAIMON_TARGET_CPU_FAMILY STREQUAL "aarch64")
+    if("${PAIMON_AARCH64_MARCH}" STREQUAL "")
+        # An explicitly empty PAIMON_AARCH64_MARCH opts out of -march entirely, for
+        # toolchains that reject the flag, such as AppleClang targeting arm64.
+        message(STATUS "PAIMON_AARCH64_MARCH is empty: passing no -march flag")
     else()
-        if(PAIMON_ARMV8_ARCH_FLAG MATCHES "\\+crypto")
-            add_definitions(-DPAIMON_HAVE_ARMV8_CRYPTO)
+        if(NOT CXX_SUPPORTS_AARCH64_MARCH)
+            message(FATAL_ERROR "The compiler does not accept ${PAIMON_AARCH64_MARCH_FLAG}. "
+                                "Set -DPAIMON_AARCH64_MARCH to a value it accepts, or to an "
+                                "empty string to pass no -march flag.")
         endif()
-        # armv8.1+ implies crc support
-        if(PAIMON_ARMV8_ARCH_FLAG MATCHES "armv8\\.[1-9]|\\+crc")
-            add_definitions(-DPAIMON_HAVE_ARMV8_CRC)
+        if(PAIMON_AARCH64_MARCH_FLAG MATCHES "native")
+            message(FATAL_ERROR "native arch not allowed, please specify arch explicitly."
+            )
         endif()
+        set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} ${PAIMON_AARCH64_MARCH_FLAG}")
     endif()
 endif()
 
--- a/src/paimon/common/data/binary_string.cpp
+++ b/src/paimon/common/data/binary_string.cpp
@@ -21,6 +21,7 @@
 
 #include <algorithm>
 #include <cctype>
+#include <cstdint>
 
 #include "paimon/common/memory/memory_segment_utils.h"
 #include "paimon/memory/bytes.h"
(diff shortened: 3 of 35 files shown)


cmake_modules/SetupCxxFlags.cmake
```diff
@@ -22,6 +22,30 @@ include(CheckCXXCompilerFlag)
...
 
+include("${CMAKE_CURRENT_LIST_DIR}/TargetArchitecture.cmake")
+message(STATUS "Target processor: ${PAIMON_TARGET_PROCESSOR} "
+               "(CPU family: ${PAIMON_TARGET_CPU_FAMILY})")
+
+# Probe the architecture specific tuning flags. x86 deliberately gets none of its
+# own, so there is nothing to probe there: the baseline stays whatever the
+# toolchain defaults to, and no persisted value depends on it; see
+# src/paimon/common/utils/crc32c.h.
+if(PAIMON_TARGET_CPU_FAMILY STREQUAL "ppc")
```
zjw1111: Thanks for the thorough cleanup of the architecture detection here.

I wonder if the ppc handling is still worth keeping. paimon-cpp does not support building for ppc: the `docs/code-style.md` update in this very PR states that only x86_64 and aarch64 are supported, ppc is not built or tested in CI, and `build_lumina` raises a `FATAL_ERROR` for any platform outside `linux-x86_64`.

On top of that, the ppc branch cannot take effect even on a ppc host. `PAIMON_ALTIVEC` is never defined anywhere in the repository — there is no `define_option` for it — so the `if(CXX_SUPPORTS_ALTIVEC AND PAIMON_ALTIVEC)` guard further down is always false, and this new `-maltivec` probe only feeds a variable that is never read. That is the same reason the x86 `PAIMON_CPU_FLAG` / `PAIMON_SIMD_LEVEL` block was removed in this PR, so it seems a little inconsistent to keep ppc.

Would it be possible to drop the ppc pieces altogether in this PR — this probe block, the `PAIMON_TARGET_CPU_FAMILY STREQUAL "ppc"` block further down in this file, and the ppc mapping in `cmake_modules/TargetArchitecture.cmake`? Unknown processors already fall through to `unknown` without an error, so ppc hosts would keep configuring with toolchain defaults, exactly as they effectively do today. If you would rather keep the scaffolding for a future port, that is fine too — in that case adding the missing `PAIMON_ALTIVEC` option would at least make the branch reachable.

zjw1111 (commented):
Thanks for taking on the aarch64 support work — the architecture handling is much cleaner after this. One question about the ppc handling that this refactor carries over.

src/paimon/common/data/binary_string.cpp
```diff
@@ -75,16 +76,19 @@ std::string BinaryString::ToString() const {
 }
 
 int32_t BinaryString::NumBytesForFirstByte(char b) {
```
lszskye, line 78: This PR adds a new rule in docs/code-style.md: “do not rely on the signedness of plain char”, and fixes NumBytesForFirstByte accordingly. However, the case-conversion code in the same file still has this issue:
- :209 / :240: (*bytes)[0] = tolower(segment_and_offset.Value()); passes a plain char directly to tolower. On x86_64, char is signed, so bytes >= 0x80 become negative. Passing such values to tolower / toupper is UB: the standard requires the argument to be representable as unsigned char or be EOF. glibc happens to tolerate -128..-1, so it does not crash. Note that this line is executed before the ASCII check at :212, so non-ASCII input also goes through it.

- :216 / :247: toupper(static_cast<int32_t>(b)) does not remove the negative value; it only rewrites the same UB in a different form.

- :229 / :260: std::transform(str.begin(), str.end(), str.begin(), ::toupper) has the same issue, because dereferencing the iterator produces a char.

src/paimon/core/casting/numeric_primitive_cast_executor.h
```diff
@@ -46,14 +50,69 @@ class NumericPrimitiveCastExecutor : public CastExecutor {
                                                const std::shared_ptr<arrow::DataType>& target_type,
                                                arrow::MemoryPool* pool) const override;
 
+    /// Converts a floating point value to an integer type the way Java does, where a plain
```
lszskye, line 53: please modify `Type Change Support Matrix` in read.rst

build_support/asan_symbolize.py
```diff
@@ -333,8 +334,17 @@ def process_stdin(self):
...
-      # Unbuffered output is not supported in Python 3
-      sys.stdout = os.fdopen(sys.stdout.fileno(), 'w')
+      # Test output is not guaranteed to be valid UTF-8, so pin both streams to the same
+      # encoding and round such bytes through surrogateescape, which passes them out
+      # unchanged instead of raising and truncating the rest of the test log. Both the
+      # encoding and the error handler have to be set on both streams: taking either from
+      # the locale or from PYTHONIOENCODING would let them disagree and re-encode the bytes.
+      # The buffers are rewrapped rather than reconfigured, which needs Python 3.7, and
+      # line buffering stands in for the unbuffered output Python 3 does not support.
+      sys.stdin = io.TextIOWrapper(sys.stdin.buffer, encoding='utf-8',
```
zjw1111, line 344: maybe we need append a ASF license for this file, because this file is modified

cmake_modules/SetupCxxFlags.cmake
SteNicholas: Good point, done — the updated revision drops the ppc pieces altogether: the `-maltivec` probe and the ppc block in `SetupCxxFlags.cmake` are removed, and `TargetArchitecture.cmake` has no ppc mapping, so a ppc host resolves to the `unknown` family and keeps configuring with toolchain defaults, exactly as it effectively did before. `cmake_modules/tests/target_architecture_test.cmake` pins that ("ppc64le is not an error"), and the commit message records the rationale.

src/paimon/common/data/binary_string.cpp
SteNicholas: Fixed in the updated revision, all three spots:

- The pre-check `(*bytes)[0] = tolower(...)` lines are removed outright. As you note they ran before the ASCII check, and they used `tolower` even in `ToUpperCase`; the loop writes every byte including index 0, so they were also redundant.
- The loop now calls `toupper`/`tolower` through `static_cast<unsigned char>(b)`, so a byte at or above 0x80 reaches the ctype functions as a non-negative value.
- `CppToUpperCase`/`CppToLowerCase` pass a lambda taking `unsigned char` to `std::transform` instead of `::toupper`/`::tolower` directly.

`binary_string_test.cpp` now asserts that invalid UTF-8 bytes (`0x80`, `0xff`) pass through case conversion unchanged on both char-signedness ABIs.

src/paimon/core/casting/numeric_primitive_cast_executor.h
SteNicholas: Done — `read.rst` is updated in this revision: note 2️⃣ under the Type Change Support Matrix now records the Java-consistent semantics (`NaN` becomes 0, an out-of-range value saturates at the int32 bounds — int64 for `bigint` — and is then narrowed to the target width; float -> tinyint: `MAX_FLOAT -> -1`, `INFINITY -> -1`, `NaN -> 0`, `300.9 -> 44`), and the overflow section above the matrix calls out float/double -> integer as the one case that is now well defined and matches Java on every supported architecture.

build_support/asan_symbolize.py
SteNicholas: This file is vendored third-party code (LLVM compiler-rt's `asan_symbolize.py`, via Apache Arrow), so per ASF policy it keeps its upstream license header and must not gain an ASF one — local modifications do not change that: https://www.apache.org/legal/src-headers.html#3party. It is accounted for in `LICENSE` ("This product includes code from LLVM compiler-rt", pointing at `licenses/LICENSE-compiler-rt.txt`) and excluded from the RAT check via `.github/.rat-excludes`, so `rat-license-check` stays green as is.

lszskye (approved):
+1

zjw1111 (approved):
+1

JingsongLi (approved):
+1

zjw1111:
@u70b3 Sorry for not pinging you during the review stage of this PR — that was an oversight on my side.

This PR has already been merged. Whenever you have some time, could you take a look and check whether the current fix meets your use case? If anything is still missing or does not work as expected on your side, please let us know and we will follow up.

u70b3:
@zjw1111 No worries, and thanks for driving the aarch64 port!

I took it for a proper spin on my local aarch64 machine (a 128-core Kunpeng-920): replicated the full CI aarch64 matrix locally — gcc debug/release, clang debug/release, ASan+UBSan, and TSan. Everything builds cleanly and the entire test suite passes (40/40). The port works well for my use case. 🎉

While validating, I also ran a deeper ARM portability sweep and found a few suspected/confirmed issues that #181 didn't cover, so I went ahead and fixed them locally:

- a real double-checked-locking race in `Singleton<T>` — hidden by x86 TSO, but TSan-confirmed on aarch64 with torn reads observed under a 128-thread stress litmus
- a TSan-confirmed race on `IOHook`'s `mode_` (fault-injection/test-only impact)
- the plain-char INT8 pattern that survived #181 in `field_sum_agg.cpp` (latent, same bug class as the min/max fix)
- a strict-aliasing UB in `SerializationUtils::DeserializeBinaryRow` (latent)
- arch-divergent double→int conversions in `cache_manager.h` / `sst_file_writer.cpp` (latent, pathological configs only)

Fixes plus deterministic TSan regression tests are in #203 — details and the red→green evidence are in the PR description. Happy to split the PR or adjust anything if you'd prefer a different shape!

PR #182, opened 2026-08-04
fix(arrow): retain pools for stream buffers

mrdrivingduck:
I found this while debugging an S3 crash. I initially suspected the S3 filesystem, but the issue was in the Arrow input-stream adapter: a returned buffer can outlive its memory pool when a reader is closed while a caller still holds a buffer. This is more likely with network filesystems, where asynchronous reads may finish after the reader is closed.

This patch uses `shared_ptr`'s aliasing constructor to keep the pool alive with the returned buffer. It also adds deterministic coverage for `Read`, `ReadAt`, and `ReadAsync` using a small `InputStream` mock. Without the change, the tests crash; with it, they pass.

### DIFF
--- a/src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
+++ b/src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
@@ -43,6 +43,20 @@ arrow::Status ValidateArrowIoRange(int64_t value, const char* name) {
     return arrow::Status::OK();
 }
 
+struct BufferWithMemoryPool {
+    std::shared_ptr<arrow::MemoryPool> pool;
+    std::shared_ptr<arrow::Buffer> buffer;
+};
+
+std::shared_ptr<arrow::Buffer> KeepMemoryPoolAlive(std::shared_ptr<arrow::Buffer> buffer,
+                                                   const std::shared_ptr<arrow::MemoryPool>& pool) {
+    // TODO(lxy): Optimize the extra allocation introduced to retain the memory pool.
+    auto holder =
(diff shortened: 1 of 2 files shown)


src/paimon/common/utils/arrow/arrow_input_stream_adapter.cpp
```diff
@@ -43,6 +43,19 @@ arrow::Status ValidateArrowIoRange(int64_t value, const char* name) {
     return arrow::Status::OK();
 }
 
+struct BufferWithMemoryPool {
```
zjw1111, line 46: copied from https://github.com/alibaba/paimon-cpp/pull/464

Image

zjw1111: could you please modify the PR title to conventional commits style, and add a TODO assigned to xinyu, noting that performance will be optimized later.

mrdrivingduck: Done with PR title. TODO means an issue or simply cc from current PR?

zjw1111: add a line in code: 
```
// TODO(lxy): xxxxxx
```

mrdrivingduck: Done in `KeepMemoryPoolAlive`.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #185, opened 2026-08-05
fix(parquet): avoid seeking into the middle of a row group when prefetch and page index filtering are both enabled

zhf999:
### Purpose

Reading a Parquet file with prefetch enabled and `parquet.read.enable-page-index-filter` left at its
default (`true`) can fail with:

```
Invalid: seek to row failed. row number N should not be in the middle of readable range
```

The prefetch coordinator and the page-filtered read path disagree about what a reader position is:

- `PrefetchFileBatchReaderImpl` drives its sub readers with read ranges from
  `ParquetFileBatchReader::GenReadRanges()`, which are **row group aligned** and cover **all** row
  groups, independent of any pushed down predicate.
- A partially matched row group returns its first *selected* row, which is generally **inside** the
  row group rather than at its start, and the sub reader crosses into the next row group as soon as
  the current one is exhausted, while the coordinator still points at the previous read range.

`HandleReadResult()` then computes `slice_end == 0` ("fully out of range"), drops the batch and
records that mid-row-group row as `readers_pos_`, expecting to seek back to it later.
`FileReaderWrapper::SeekToRow()` only accepts row group boundaries, so the following
`EnsureReaderPosition()` fails and the read is aborted. The drop-and-re-read strategy comes from
ORC, whose `RowReader::seekToRow()` accepts an arbitrary row; it cannot work for Parquet.

Two independent defects are fixed:

**1. `PrefetchFileBatchReaderImpl::HandleReadResult()` no longer drops a batch that belongs to a
later read range.** It looks up the read range that actually contains the batch and re-keys the batch
to it, so no backward seek is needed at all. Rows that belong to *another* sub reader are still
dropped, which is safe because that reader produces them. The lookup is restricted to
`read_ranges_in_group_[reader_idx]`, i.e. the ranges assigned to the producing reader, so the
existing "one read range is served by exactly one queue" property is preserved and output ordering is
unchanged. The recursion can happen at most once, because the first row id lies inside the range
found and therefore cannot produce a zero slice end again.

**2. `FileReaderWrapper::NextPageFiltered()` now advances `next_row_to_read_`.** It used to leave the
cursor at the row group start for the whole duration of a partially matched row group, so
`GetNextRowToRead()` reported a position the reader had long passed. The cursor now points at the
next row that survives filtering (mapped through the row group's `RowRanges`), or at the row group
end once the filtered ranges are exhausted. This keeps the coordinator's read range in step with the
sub reader, so the "fully out of range" path above becomes a pure fallback instead of the common case.

Notes on behaviour:

- Output order is unchanged. Ordering is driven by the global `read_ranges_` cursor, and read ranges
  that yield no data are still retired by the existing `min_range` rule in `NextBatchWithBitmap()`.
- Read ranges skipped by pushdown are now retired as soon as the first real batch arrives, instead of
  after a drop plus re-seek round trip, which removes one redundant row group decode.
- ORC keeps its behaviour; it simply no longer re-reads a batch it has already produced.

Known follow-up, intentionally left out: `ParquetFileBatchReader::GenReadRanges()` still reports every
row group, including those already pruned by the predicate. Aligning it with the surviving target row
groups would avoid empty read ranges, but it also changes range dispatch and the adaptive prefetch
heuristic, so it deserves its own change.

### Tests

Two new cases, each verified to fail before the corresponding fix:

- `WriteAndReadInteTest.TestAppendWithParquetPageIndexFilterAndPrefetch` (`test/inte`) - to verify that pre-metioned problem is solved.
- `FileReaderWrapperTest.PageFilteredAdvancesNextRowToRead` (`paimon-parquet-format-test`) — to verify that `next_row_to_read_` is pushed forward.

### API and Format

No. 

### Documentation

No.

### Generative AI tooling

Generated-by: GLM 5.2

### DIFF
--- a/test/inte/write_and_read_inte_test.cpp
+++ b/test/inte/write_and_read_inte_test.cpp
@@ -1590,9 +1590,11 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilterAndPrefetch) {
     ASSERT_OK_AND_ASSIGN(auto result_plan, table_scan->CreatePlan());
     ASSERT_FALSE(result_plan->Splits().empty());
 
-    // A single prefetch sub reader owns every read range, so it is the one that has to cross
-    // row group boundaries. Row level filtering stays off: the expected rows below are exactly
-    // what page-index filtering selects.
+    // The row group aligned read ranges still cover RG0 although the predicate pruned that row
+    // group, and 2 sub readers take the ranges round robin. The reader0 owning RG0's range therefore
+    // finds no data for it and skips ahead into the next row group it owns, whose first selected row
+    // sits in the middle of that row group. Row level filtering stays off: the expected rows below
+    // are exactly what page-index filtering selects.
     ReadContextBuilder read_context_builder(table_path);
     read_context_builder.SetOptions(options)
         .SetPredicate(predicate)
--- a/test/inte/write_and_read_inte_test.cpp
+++ b/test/inte/write_and_read_inte_test.cpp
@@ -1591,10 +1591,10 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilterAndPrefetch) {
     ASSERT_FALSE(result_plan->Splits().empty());
 
     // The row group aligned read ranges still cover RG0 although the predicate pruned that row
-    // group, and 2 sub readers take the ranges round robin. The reader0 owning RG0's range therefore
-    // finds no data for it and skips ahead into the next row group it owns, whose first selected row
-    // sits in the middle of that row group. Row level filtering stays off: the expected rows below
-    // are exactly what page-index filtering selects.
+    // group, and 2 sub readers take the ranges round robin. The reader0 owning RG0's range
+    // therefore finds no data for it and skips ahead into the next row group it owns, whose first
+    // selected row sits in the middle of that row group. Row level filtering stays off: the
+    // expected rows below are exactly what page-index filtering selects.
     ReadContextBuilder read_context_builder(table_path);
     read_context_builder.SetOptions(options)
         .SetPredicate(predicate)
--- a/test/inte/write_and_read_inte_test.cpp
+++ b/test/inte/write_and_read_inte_test.cpp
@@ -1530,7 +1530,7 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilter) {
 }
 
 /// Reproduces the prefetch + parquet page-index filter failure: the predicate keeps only the last
-/// page of every row group, so the prefetch reader ends up seeking to a row in the middle of a row
+/// page of RG1, RG2 and RG3, so the prefetch reader ends up seeking to a row in the middle of a row
 /// group, which FileReaderWrapper::SeekToRow rejects.
 TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilterAndPrefetch) {
     auto [file_format, file_system] = GetParam();
@@ -1575,7 +1575,7 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilterAndPrefetch) {
     ASSERT_OK(helper->WriteAndCommit(std::move(batch), /*commit_identifier=*/0,
                                      /*expected_commit_messages=*/std::nullopt));
 
-    // Keep only the last row of every row group, so each row group is partially matched and its
+    // Keep only the last row of RG1, RG2 and RG3, so each row group is partially matched and its
     // first selected row is 3 rows behind the row group start.
     auto predicate = PredicateBuilder::In(/*field_index=*/0, /*field_name=*/"f0", FieldType::INT,
                                           {Literal(7), Literal(11), Literal(15)});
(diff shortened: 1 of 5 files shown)


test/inte/write_and_read_inte_test.cpp
```diff
@@ -1529,6 +1529,94 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilter) {
...
+        [8, "v8"],   [9, "v9"],   [10, "v10"], [11, "v11"],
+        [12, "v12"], [13, "v13"], [14, "v14"], [15, "v15"]
+    ])";
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatch> batch,
+                         TestHelper::MakeRecordBatch(arrow::struct_(fields), data,
+                                                     /*partition_map=*/{}, /*bucket=*/0, {}));
+    ASSERT_OK(helper->WriteAndCommit(std::move(batch), /*commit_identifier=*/0,
+                                     /*expected_commit_messages=*/std::nullopt));
+
+    // Keep only the last row of every row group, so each row group is partially matched and its
```
zjw1111: Small wording nit on this comment: the predicate `In{7, 11, 15}` matches the last row of RG1/RG2/RG3, but RG0 (rows 0–3) has no match and is fully pruned (excluded), so it is not "partially matched". Would it be possible to phrase this as "the last row of every *matched* row group" to avoid the impression that RG0 is also partially matched? This doesn't affect the test or the expected result `[7, 11, 15]`.

Separately (no change needed here): the new integration test only exercises `SetPrefetchMaxParallelNum(1)`, i.e. a single sub reader crossing row group boundaries. The re-keying + `min_range` retirement interaction under multiple concurrent reader queues isn't directly covered — it looks correct from the code, but a multi-reader regression case could be worth a follow-up if it's cheap to add.

zjw1111 (commented):
Thanks for the detailed write-up and for fixing this — the root-cause analysis (prefetch coordinator vs. page-filtered read path disagreeing on the reader position) is very clear, and I appreciate that both new tests are verified to fail before the corresponding fix.

I reviewed the two changes and did not find blocking issues:

- `HandleReadResult()` re-keying the batch to the read range that actually contains it looks correct. The "recurses at most once" reasoning holds: since `global_row_ids[0]` lies within `owner_range`, the recursive call still has `slice_begin == 0` but necessarily `slice_end >= 1`, so it cannot produce a zero slice end again. Restricting the lookup to `read_ranges_in_group_[reader_idx]` preserves the "one read range served by exactly one queue" property, and the `min_range` retirement in `NextBatchWithBitmap()` keeps output ordering unchanged. The `nullopt` fallback also stays safe because it only lands `readers_pos_` on a row-group-aligned range start.
- `NextPageFiltered()` advancing `next_row_to_read_` via `MapFilteredIndexToOriginalRow` correctly skips holes inside a row group's `RowRanges` and lands on the row group end once the filtered ranges are exhausted, matching the semantics used by `NextFullyMatched()` / `SeekToRow()`.

One minor, non-blocking follow-up on the multi-reader path is left as an inline note. The known `GenReadRanges()` follow-up you called out seems like a reasonable separate change.

lxy-9602 (approved):
+1

lucasfang:
+1

lucasfang (approved):

zjw1111 (commented):
Thanks for the updates — the comment wording now reads accurately, and bumping the prefetch parallelism to 2 nicely covers the multi-reader path I was wondering about (the surviving read ranges get dispatched across two sub readers, and a reader still crosses row group boundaries because its assigned ranges are non-contiguous). The core fix is unchanged and still looks good.

One small leftover inline note below.

test/inte/write_and_read_inte_test.cpp
```diff
@@ -1529,6 +1529,94 @@ TEST_P(WriteAndReadInteTest, TestAppendWithParquetPageIndexFilter) {
...
+    ScanContextBuilder scan_context_builder(table_path);
+    scan_context_builder.SetOptions(options)
+        .AddOption(Options::SCAN_MODE, StartupMode::LatestFull().ToString())
+        .SetPredicate(predicate);
+    ASSERT_OK_AND_ASSIGN(auto scan_context, scan_context_builder.Finish());
+    ASSERT_OK_AND_ASSIGN(auto table_scan, TableScan::Create(std::move(scan_context)));
+    ASSERT_OK_AND_ASSIGN(auto result_plan, table_scan->CreatePlan());
+    ASSERT_FALSE(result_plan->Splits().empty());
+
+    // A single prefetch sub reader owns every read range, so it is the one that has to cross
```
zjw1111: This comment is now stale after raising `SetPrefetchMaxParallelNum` to 2: the read ranges are no longer owned by a single sub reader — they are dispatched across the two readers. Would it be possible to update the wording, e.g. "The read ranges are dispatched across 2 sub readers; a reader still has to cross row group boundaries because its assigned ranges are non-contiguous"? Purely a comment/code consistency nit, no behavior impact.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #188, opened 2026-08-07
perf(reader): rebase Arrow offsets by slicing buffers (zero-copy) instead of copying values

zhf999:
### Purpose

`BatchReader` requires every returned array, including its nested children, to have a zero offset.
`ArrowUtils::NormalizeRecordBatchOffsets`  enforced that by calling `arrow::Concatenate({column})`,
which has no single-array fast path and therefore deep copies every buffer of the column.

This is on the hot path of page-filtered Parquet reads. `PageFilteredRowGroupReader` materializes a
filtered row group as an `arrow::Table` and streams it through `arrow::TableBatchReader`, which
returns `chunk->Slice(offset, chunksize)` for every batch after the first. As a result every batch
except the first was deep copied, column by column, adding a full extra copy of all scanned data.
Since nested columns now also go through this path, `Concatenate` recursively copied list offsets,
child values and per-level validity bitmaps as well.

Normalizing a slice does not require copying anything: a contiguous slice can be rebased to a zero
offset by slicing the underlying buffers.

- Fixed width: slice the value buffer.
- Boolean: rebase the value bitmap, which only needs a bit shift when the offset is not byte aligned.
- String / binary / large binary: rewrite the offsets so the first one is zero, and slice the value
  buffer to the referenced range.
- List / large list / map: rewrite the offsets, then recursively rebase the contiguous child range
  the offsets span.
- Struct: recursively rebase every child, sliced in step with the parent, because a struct slice
  keeps full length children.
- Validity bitmaps: sliced when byte aligned, and dropped entirely when the array has no nulls.

Layouts that cannot be rebased this way, currently dictionary and any type not listed above, keep
the previous behavior and fall back to `arrow::Concatenate`. 

### Tests

Added to `ArrowUtilsTest`:

- `TestNormalizeRecordBatchOffsetsCoversSupportedTypes`

  16 types crossed with 8 slices each. Every combination asserts `ValidateFull()`, equality with
  the original batch, and recursively that all offsets are zero.

- `TestNormalizeRecordBatchOffsetsSharesValueBuffers`

  Over the same 16 types, asserts that the buffer holding the values of each layout, addressed by a
  path of child indices plus a buffer index, is contained in the address range of the corresponding
  source buffer. This pins the optimization: an implementation that copies would allocate outside the source range and fail here.

- `TestNormalizeRecordBatchOffsetsFallsBackForDictionary`

  Checks that the copying fallback still satisfies the zero offset contract, and asserts that its
  index buffer is *not* contained in the source range. That negative case is what makes the
  containment assertions above meaningful rather than trivially true.

#### Datasets
Multiple performance testing were conducted on 1 dataset.
|Dataset| File size | num_rows | num_columns | num_rowgroups | page length (in rows)|
| --- | --- | --- | --- | --- | --- | 
| Dataset#1 | 6.7GB | 2M | 1,176 | 39 | 1024 |

#### Test Methodology
To validate bitmap behavior, we organize tests into six groups based on bitmap distribution patterns. Each group contains five distinct datasets, and the final result reported is the average of these five runs.

#### Test Cases
1.  Matching records are contiguous and only one row is queried.
2.  Matching records are contiguous and 10,000 rows are queried.
3.  Matching records are scattered in 5 segments, each segment containing 1 row.
4.  Matching records are scattered in 5 segments, each segment containing 100 rows.
5.  Matching records are scattered in 100 segments, each segment containing 1 row.
6.  Skip 100 rows, read 100 rows, skip 100 rows, read 100 rows util EOF.
7. Skip 10000 rows, read 10000 rows, skip 10000 rows, read 10000 rows util EOF.
8.  Scattered bitmap, skip 1 row, read 1 row, skip 1 row, read 1 row until EOF (i.e. `1010101010...`).
9. Scattered bitmap, skip 32 rows, read 1 row, skip 32 rows, skip 1 row until EOF.
10. Scattered bitmap, skip 99 rows, read 1 row, skip 99 rows, skip 1 row until EOF.
11.  Scattered bitmap, skip 9999 rows , read 1 row, skip 9999 rows, skip 1 row until EOF.

#### Test Results
On dataset1
| Branch | Case 1 | Case 2 | Case3 | Case 4 | Case 5 | Case 6 | Case 7 | Case 8 | Case 9 |  Case 10 | Case 11| Case 12|
| --- | --- | --- | --- | --- | --- | --- | --- | ---| --- | --- | --- | --- |
| Before | 0.38s | 0.84s | 1.47s |1.49s |16.07s | 115.14s | 82.03s | 140.68s | 87.00s | 74.39s | 21.84s | 124.83s |
| After | 0.34s | 0.77s | 1.31s | 1.33s | 15.27s | 100.06s | 71.48s | 122.73s | 82.02s | 67.98s | 20.40s | 124.37s |

### API and Format

No change. 

### Documentation

No change.

### Generative AI tooling

Generated-by:  GLM 5.2


src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -430,6 +430,192 @@ TEST(ArrowUtilsTest, TestNormalizeRecordBatchOffsets) {
     ASSERT_EQ(unchanged_batch.get(), normalized_batch.get());
 }
 
+namespace {
+
+/// A buffer that rebasing must expose as a view into the source.
+// This struct tells where a ArrayData stores value.
+struct SharedBuffer {
+    std::vector<int> child_path;
```
lxy-9602: int32_t

src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -430,6 +430,192 @@ TEST(ArrowUtilsTest, TestNormalizeRecordBatchOffsets) {
...
+         {{{0, 0}, 2}, {{0, 1}, 1}}},
+    };
+}
+
+const arrow::ArrayData& ResolvePath(const arrow::ArrayData& data,
+                                    const std::vector<int>& child_path) {
+    const arrow::ArrayData* node = &data;
+    for (int child_index : child_path) {
+        node = node->child_data[child_index].get();
+    }
```
lxy-9602, line 514: int32_t

src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -430,6 +430,192 @@ TEST(ArrowUtilsTest, TestNormalizeRecordBatchOffsets) {
...
+         "[true, null, false, true, true, null, false, false, true, null]",
+         {{{}, 1}}},
+        {arrow::int8(), "[0, 1, null, 3, 4, 5, null, 7, 8, 9]", {{{}, 1}}},
+        {arrow::int32(), "[0, 1, null, 3, 4, 5, null, 7, 8, 9]", {{{}, 1}}},
+        {arrow::int64(), "[0, 1, null, 3, 4, 5, null, 7, 8, 9]", {{{}, 1}}},
+        {arrow::float64(), "[0.5, 1.5, null, 3.5, 4.5, 5.5, null, 7.5, 8.5, 9.5]", {{{}, 1}}},
+        {arrow::date32(), "[0, 1, null, 3, 4, 5, null, 7, 8, 9]", {{{}, 1}}},
+        {arrow::decimal128(10, 2),
+         R"(["1.23", null, "3.45", "6.78", "0.01", null, "9.99", "8.88", "7.77", "6.66"])",
+         {{{}, 1}}},
```
lxy-9602, line 470: Please add tests for timestamp type.

src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -430,6 +430,192 @@ TEST(ArrowUtilsTest, TestNormalizeRecordBatchOffsets) {
...
+    CheckNormalizedSlice(array, /*offset=*/3, /*length=*/3);
+
+    // The fallback copies, which is also what makes the sharing checks above meaningful.
+    std::shared_ptr<arrow::RecordBatch> batch = MakeSliceBatch(array, /*offset=*/1, /*length=*/4);
+    ASSERT_OK_AND_ASSIGN(
+        std::shared_ptr<arrow::RecordBatch> normalized,
+        ArrowUtils::NormalizeRecordBatchOffsets(batch, arrow::default_memory_pool()));
+    ASSERT_FALSE(IsViewInto(*normalized->column_data(0)->buffers[1], *array->data()->buffers[1]));
+}
+
```
lxy-9602, line 706: Could we add a case where the parent offset is 0 but a child offset is non-zero?

The current cases all create slices from the top-level array, so they do not cover the input shape that motivates the recursive `HasNonZeroOffset` check. For example, construct a StructArray or ListArray directly from a sliced child, then verify equality and that all offsets are normalized to 0.

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #187, opened 2026-08-05
fix(commit): allow disabled manifest delete stats option

liujiayi771:
### Purpose

Allow the C++ commit path to accept `manifest.delete-file-drop-stats=false`. The disabled value matches the existing C++ behavior, which preserves statistics in DELETE manifest entries. The enabled value remains unsupported and rejected.

### Tests

- `./build-release/release/paimon-core-test --gtest_filter='FileStoreCommitImplTest.ValidateCommitOptionsAllowsDisabledManifestDeleteFileDropStats:FileStoreCommitImplTest.ValidateCommitOptionsRejectsUnsupportedOptions'`
- `./build-release/release/paimon-core-test --gtest_brief=1` (1579 tests passed)
- `pre-commit run --files src/paimon/core/operation/file_store_commit_impl.cpp src/paimon/core/operation/file_store_commit_impl_test.cpp`

### API and Format

No public API, storage format, or protocol changes.

### Documentation

No documentation changes. This fixes validation of an existing option.

### Generative AI tooling

Generated-by: Codex, GPT-5.

PR #190, opened 2026-08-09
feat(commit): support dropping delete file stats

liujiayi771:
### Purpose

This is a follow-up to #187 and adds full support for `manifest.delete-file-drop-stats=true` in the C++ commit path.

- Parse and expose the existing Paimon table option.
- Drop `value_stats` and `value_stats_cols` from final DELETE manifest entries.
- Apply the optimization after scan filtering and pruning for overwrite and append compaction paths.
- Preserve complete statistics for ADD entries, generic metadata scans, and writer restore.
- Keep writer restore statistics until the final entry kind is known, intentionally avoiding the metadata-only level-upgrade bug tracked in https://github.com/apache/paimon/issues/7026.

### Tests

- `cmake --build build --target paimon_static paimon_shared -j8`
- `cmake --build build --target paimon-core-test -j8`
- `./build/debug/paimon-core-test --gtest_brief=1` (1629 tests passed)
- `pre-commit run --files <changed files>`
- `git diff --check apache/main...HEAD`

### API and Format

Adds the public option key `kManifestDeleteFileDropStats`.

There is no manifest schema or protocol change. When enabled, DELETE entries store empty value statistics and an empty value-statistics column list, matching the existing Paimon format. ADD entries retain complete statistics for file skipping.

### Documentation

This implements an existing Paimon table option. No documentation files are included in this PR.

### Generative AI tooling

Generated-by: Codex (GPT-5) and Claude Code.

### DIFF
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -173,6 +173,10 @@ struct PAIMON_EXPORT Options {
     /// compaction of manifest, default value is 16MB.
     static const char MANIFEST_FULL_COMPACTION_FILE_SIZE[];
 
+    /// "manifest.delete-file-drop-stats" - Whether final DELETE manifest entries should omit
+    /// value statistics. Default is false.
+    static const char MANIFEST_DELETE_FILE_DROP_STATS[];
+
     /// "source.split.target-size" - Target size of a source split when scanning a bucket. Default
     /// value is 128MB.
     static const char SOURCE_SPLIT_TARGET_SIZE[];
--- a/include/paimon/defs.h
+++ b/include/paimon/defs.h
@@ -174,7 +174,7 @@ struct PAIMON_EXPORT Options {
     static const char MANIFEST_FULL_COMPACTION_FILE_SIZE[];
 
     /// "manifest.delete-file-drop-stats" - Whether final DELETE manifest entries should omit
-    /// value statistics. Default is false.
+    /// value statistics. Default is false only for compatibility with old readers.
     static const char MANIFEST_DELETE_FILE_DROP_STATS[];
 
     /// "source.split.target-size" - Target size of a source split when scanning a bucket. Default
--- a/src/paimon/core/operation/abstract_file_store_write.cpp
+++ b/src/paimon/core/operation/abstract_file_store_write.cpp
@@ -302,6 +302,10 @@ Result<std::shared_ptr<RestoreFiles>> AbstractFileStoreWrite::ScanExistingFileMe
     if (dv_maintainer_factory_) {
         index_file_handler = dv_maintainer_factory_->GetIndexFileHandler();
     }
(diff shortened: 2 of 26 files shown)


liujiayi771:
cc @lucasfang @lxy-9602

src/paimon/core/operation/commit/manifest_entry_changes.h
```diff
@@ -39,7 +39,11 @@ namespace paimon {
 /// Detailed changes from `CommitMessage`s.
 class ManifestEntryChanges {
  public:
-    explicit ManifestEntryChanges(int32_t default_num_bucket);
+    /// Create a change collector.
+    ///
+    /// @param default_num_bucket Bucket count used when a commit message omits it.
+    /// @param drop_delete_file_stats Whether DELETE data-file entries should omit value stats.
+    explicit ManifestEntryChanges(int32_t default_num_bucket, bool drop_delete_file_stats = false);
```
lxy-9602, line 47: Could we avoid using a default argument here? In production code, it’s better to pass the parameter explicitly.

src/paimon/core/operation/abstract_file_store_write.cpp
```diff
@@ -302,6 +302,10 @@ Result<std::shared_ptr<RestoreFiles>> AbstractFileStoreWrite::ScanExistingFileMe
     if (dv_maintainer_factory_) {
         index_file_handler = dv_maintainer_factory_->GetIndexFileHandler();
     }
+    // Paimon Java currently drops value stats during writer restore. This is a known bug: a
+    // restored file can become a compact-after ADD via metadata-only level upgrade and lose its
+    // stats (https://github.com/apache/paimon/issues/7026). C++ intentionally does not align with
+    // that behavior; stats are dropped later only when the final entry kind is DELETE.
```
lxy-9602, line 308: Nice catch! Could you please add a PK compaction end-to-end test in `pk_compaction_inte_test.cpp` to demonstrate that level upgrades do not lose their stats?

include/paimon/defs.h
```diff
@@ -173,6 +173,10 @@ struct PAIMON_EXPORT Options {
     /// compaction of manifest, default value is 16MB.
     static const char MANIFEST_FULL_COMPACTION_FILE_SIZE[];
 
+    /// "manifest.delete-file-drop-stats" - Whether final DELETE manifest entries should omit
+    /// value statistics. Default is false.
+    static const char MANIFEST_DELETE_FILE_DROP_STATS[];
```
lxy-9602, line 178: Also, could you complete the comment here? I noticed that the Java version also includes: “Default value is false only for compatibility with old readers.”

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #191, opened 2026-08-10
feat(read): support deletion vectors for data-evolution tables

### LINKED ISSUE #169: [Feature] Support deletion vectors for data-evolution tables
### Search before asking

### Motivation

Java Paimon supports deletion vectors on data-evolution tables since [`[core] Introducing DeletionVector mechanism for DataEvolution tables` (apache/paimon#8380)](https://github.com/apache/paimon/pull/8380): the `data-evolution.enabled` × `deletion-vectors.enabled` mutual exclusion was dropped from `SchemaValidation.validateRowTracking` (data evolution now only conflicts with `clustering.incremental`), and the scan, read and compaction paths learned to apply deletion files to data-evolution splits.

paimon-cpp still carries the old constraint, both in schema validation and in the read path:

- `SchemaValidation::ValidateRowTracking` rejects the combination outright (`src/paimon/core/schema/schema_validation.cpp:477-483`).
- The same function requires `data-evolution.enabled` for any table holding a BLOB column (`src/paimon/core/schema/schema_validation.cpp:501-504`), so for BLOB tables the two checks form an unsatisfiable pair — a BLOB table can never enable deletion vectors.
- Even with the schema check relaxed, reading would still fail: `DataEvolutionSplitRead::ApplyIndexAndDvReaderIfNeeded` returns `Invalid: DataEvolutionSplitRead do not support deletion vector` (`src/paimon/core/operation/data_evolution_split_read.cpp:334-336`), as documented on `src/paimon/core/operation/data_evolution_split_read.h:68`.

Creating a table with an INT column and a BLOB column plus `row-tracking.enabled=true`, `data-evolution.enabled=true` and `deletion-vectors.enabled=true` therefore fails at create time with:

```text
Invalid: Data evolution config must disabled with deletion-vectors.enabled
```

The current behavior is pinned by `SchemaValidationTest.TestRowTracking` (`src/paimon/core/schema/schema_validation_test.cpp:773-782`).

This blocks cross-language use of BLOB + deletion vector tables: a table created and written by a Java engine that includes apache/paimon#8380 can neither be created by nor read through paimon-cpp. It is also the follow-up already noted in #452 ("Java applies deletion vectors to placeholder gap readers; the C++ data-evolution blob read path does not wire deletion vectors yet (pre-existing)").

### Solution

Port the Java design of apache/paimon#8380:

1. **Schema validation** — drop the `data-evolution.enabled` × `deletion-vectors.enabled` check in `SchemaValidation::ValidateRowTracking` and update `SchemaValidationTest.TestRowTracking` accordingly, keeping the other data-evolution constraints (data evolution requires row tracking, a BLOB table must have other normal columns, a BLOB column cannot be a partition key).
2. **Scan and split plumbing** — populate the deletion files of data-evolution splits so every field bunch knows which deletion file applies to which file. `DataSplitImpl` already carries `DeletionFiles()`; `DataEvolutionFileStoreScan` and the data-evolution split generation need to fill and align them, mirroring the Java changes in `DataEvolutionFileStoreScan`, `DataSplit` and the new `DataEvolutionUtils`.
3. **Read path** — wire the existing `ApplyDeletionVectorBatchReader` (`src/paimon/core/deletionvectors/apply_deletion_vector_batch_reader.h`) into `DataEvolutionSplitRead::ApplyIndexAndDvReaderIfNeeded` instead of returning `Invalid`, composing it with the row-range selection and with the blob layers of `BlobFallbackBatchReader`, mirroring Java's `ApplyDeletionFileRecordIterator`, `BlobFallbackRecordReader` and the new `AllPlaceholdersRecordReader`.
4. **Write path** — data-evolution tables are unaware-bucket (`bucket = -1`) append tables, while paimon-cpp only has `BucketedDvMaintainer` (`src/paimon/core/deletionvectors/bucketed_dv_maintainer.h`); producing and merging deletion files for these tables needs the equivalent of Java's `AppendDeleteFileMaintainer`. This step is only required for deleting from C++ — reading a Java-written DV table needs steps 1–3 only.

Coverage: schema validation tests for the accepted combination, `DataEvolutionSplitRead` tests for a data-evolution split whose files carry deletion files (including a blob bunch with multiple sequence layers), and a cross-language check that Java and C++ agree on the rows surviving the deletion vector of a BLOB table with row tracking, data evolution and deletion vectors enabled.

### Anything else?

apache/paimon#8380 also updated data-evolution compaction (`DataEvolutionCompactCoordinator`, `DataEvolutionCompactTask`) to account for deletion files. paimon-cpp has no data-evolution compaction yet, so that part maps to a follow-up rather than to this issue.

SteNicholas:
### Purpose

Linked issue: close #169

Port the read side of Apache Paimon [`d2be7eace`](https://github.com/apache/paimon/commit/d2be7eacedf7c001ed1022856a62ade089e954c4) (apache/paimon#8380), so a data-evolution table can enable `deletion-vectors.enabled`.

A row range group's deletion vector is maintained against the group's **anchor file**, so its positions are anchor-relative rather than file-relative. Every reader of the group must therefore drop the same rows to keep the column merge positionally aligned: file readers apply the vector shifted by the file's offset inside the anchor row id range, and the blob fallback path drops the deleted row ids from its placeholder gap segments, which have no file reader to wrap.

Main changes:

- Add `DataEvolutionUtils::RetrieveAnchorFile`, matching Java's anchor rule: the oldest normal file of the group by `(max_sequence_number, file_name)`, skipping blob and vector-store files. The producer side (`AppendDeleteFileMaintainer`) follows the same rule, so the file a Java engine attaches the deletion vector to is the one this read path looks it up by.
- Drop the `SchemaValidation` rule forbidding `data-evolution.enabled` together with `deletion-vectors.enabled`, matching Java.
- Support committing a deletion vector index file for a table without buckets, which a data-evolution table always is. `IndexManifestFileHandler` combines such index files by file name, the way Java's `GlobalCombiner` does, rejecting a second vector for a data file or the removal of one that is not there. Without this the deletes another engine issues could not reach the table through this commit path at all, so lifting only the `SchemaValidation` rule would have left the combination unusable.
- Narrow, rather than drop, `ConflictDetection`'s refusal of deletion vectors in `BUCKET_UNAWARE` mode. It now refuses only a delta that **deletes data files**: such a table keeps its vectors in the index manifest, which the data file entries do not reference, and Java pairs the two first (`ConflictDetection#buildBaseEntriesWithDV`) so that dropping a data file conflicts with a concurrent commit rewriting that file's vector. That pairing is not ported, so the shape needing it fails fast instead of committing against a stale state. Adding files stays allowed, and so does another engine replacing a vector, which travels through the delta index entries this check does not inspect.
- Subtract deletion vector cardinality from `DataSplit::MergedRowCount`, and report the count unavailable when a deletion file carries no cardinality.
- Skip the limit push down when a non-partition filter or a row range index is present: a split's metadata row count is then only an upper bound of what the read returns, so pruning on it could return fewer rows than the limit.
- Apply the row ranges selection and the row range groups' deletion vectors to the blob view pre-read, so a reference held only by a dropped row is no longer resolved.
- Read a row range group made only of blob files, which a split can pack next to a group carrying a deletion vector, without an anchor. No writer can key a vector by such a group, so it reads undeleted rather than failing the whole split on the missing anchor. **Paimon Java throws here**; this is a deliberate divergence.

Scope: Paimon C++ **does not produce** these deletion vectors — the deletes are issued by another engine, and this change only lets that engine's index-only commit land and be read back. `deletion-vectors.bitmap64` and compaction of such a table remain unsupported.

Two behavior changes worth calling out for reviewers, since they reach append and primary key tables too:

- `ApplyPushDownLimit` used to consider only raw convertible splits, silently leaving a merged split out of the pruned plan. It now asks every split for its row count and abandons the push down as soon as one cannot be counted from metadata, which is the same conservative answer the raw convertible path already gave. That is what lets a data-evolution split, which is not raw convertible but does have an exact count, take part in the pruning.
- The new `CanPushDownLimit` guard also fixes a wrong result that predates deletion vectors: the push down never consulted the non-partition predicate, so a limited query carrying a value filter pruned splits on row counts the filter had not been applied to yet, and could return fewer rows than the limit asked for while more still existed.

### Tests

Unit tests:

| Case | Covers |
| --- | --- |
| `DataEvolutionUtilsTest.TestRetrieveAnchorFileSkipsBlobAndVectorStoreFiles` | anchor skips blob / vector-store files; `IsNormalFile` agrees |
| `DataEvolutionUtilsTest.TestRetrieveAnchorFileFailsWithoutNormalFile` | group with no normal file errors |
| `DataEvolutionUtilsTest.TestRetrieveAnchorFileTieBreaksWithFileName` | `(max_sequence_number, file_name)` tie-break |
| `DataEvolutionSplitReadTest.TestReadGroupDeletionVector` | anchor lookup; null factory, absent and empty vector all yield nullopt; a group with no anchor yields nullopt instead of failing |
| `DataEvolutionSplitReadTest.TestCreateGroupDvFactoryShiftsPositions` | per-file offset shift, window limit, read-only guards |
| `DataEvolutionSplitReadTest.TestCreateGroupDvFactoryRejectsFileOutsideAnchorRange` | file outside the anchor range is rejected |
| `DataEvolutionSplitReadTest.TestExcludeDeletedRowIds` | gap range arithmetic, including a fully deleted range |
| `DataSplitTest.TestDataEvolutionMergedRowCountSubtractsDeletionFileCardinality` | cardinality subtracted |
| `DataSplitTest.TestDataEvolutionMergedRowCountUnavailableWithoutCardinality` | missing cardinality makes the count unavailable |
| `DataSplitTest.TestSerializeDataEvolutionSplitWithDeletionFiles` | first row ids + deletion files survive a serialization round trip |
| `SchemaValidationTest` (2 cases updated) | the combination is now accepted, including on a blob table |
| `IndexManifestFileHandlerTest.GlobalDvCombinerReplacesIndexFileInBucketUnawareMode` | an updated vector replaces its index file, keeping the other group's |
| `IndexManifestFileHandlerTest.GlobalDvCombinerRejectsInconsistentDelta` | a second vector for one data file, a re-added index file and a deleted absent one are all rejected |
| `ConflictDetectionTest.TestDeletionVectorsAllowedWithBucketUnawareMode` (updated) | adding files commits; deleting a data file is refused |
| `ConflictDetectionTest.TestDeletionVectorsDeleteAllowedWithoutBucketUnawareMode` | the refusal is scoped to bucket-unaware tables |

Integration tests (`DataEvolutionTableTest`, parquet/orc/avro):

`TestReadWithDeletionVectors`, `TestReadWithDeletionVectorsAcrossReadBatches`, `TestReadWithDeletionVectorsOnPartOfRowRangeGroups`, `TestReadWithDeletionVectorsOnEveryRowRangeGroup` (two groups with their own vectors in one split, plus plan-level assertions on which data file carries which deletion file), `TestReadWithFullyDeletedRowRangeGroup`, `TestReadAfterUpdatingDeletionVectors`, `TestReadWithDeletionVectorsAfterAddingColumn`, `TestLimitPushDownWithHeavilyDeletedFirstRowRangeGroup`, `TestLimitPushDownDisabledByNonPartitionFilter`, `TestLimitPushDownKeptByPartitionFilter`, `TestLimitPushDownDisabledByRowRangeIndex`.

Integration tests (`BlobTableInteTest`):

`TestDataEvolutionBlobPartialUpdateWithDeletionVectors` (multi-layer blob fallback, placeholder gaps, blob-only projection), `TestBlobViewSkipsDanglingReferenceOfDeletedRow`, `TestBlobViewSkipsDanglingReferenceInEveryRowRangeGroup`, `TestBlobViewPreReadHonorsRowRangesNotPredicate`.

Deletion vectors are committed the way an external engine does, through the new `DeletionVectorTestHelper`: a deletion vector index file written with the production `DeletionVectorsIndexFile` writer and committed as an index-only commit message keyed by the group's anchor file.

### API and Format

No public API changes: the only edit under `include/` is a comment, an `@note` on `Options::DELETION_VECTORS_ENABLED` recording that Paimon C++ reads but does not write these vectors on a data-evolution table. No storage format or protocol changes.

`IndexManifestFileHandler::IndexManifestFileCombiner::Combine` now returns a `Result`, so the new combiner can report a rejected delta instead of writing an index manifest that contradicts itself; the class is private to the handler and the other two combiners keep their behavior. Review feedback likewise turned `DeletionVector::GetCardinality` into a `Result<int64_t>`: the position-shifted window view over a group's vector is the one implementation that can fail, and it now propagates that failure instead of asserting and returning a guess; the interface is internal, and its two production callers propagate the error. The deletion vector index format read here is the existing one; this change only makes it reachable for data-evolution tables. `DataSplitImpl::data_deletion_files_` gains a documented invariant (aligned one-to-one with the data files; only anchor files carry an entry) but its serialized form is unchanged, which `DataSplitTest.TestSerializeDataEvolutionSplitWithDeletionFiles` pins.

### Documentation

Yes.

- `docs/source/user_guide/read.rst` gains a section, *Deletion Vectors on Data-Evolution Tables*, covering how the anchor file locates a group's vector, how the shift and the blob fallback gaps apply it, and the limitations: only the default 32-bit vectors are readable, Paimon C++ does not write them, a commit dropping data files is refused, and such a table is never compacted.
- `docs/source/user_guide/compaction.rst` gains a note that such a table is never compacted here, matching the Paimon Java revision this was ported from; later Java releases do compact it and Paimon C++ has not caught up.

### Generative AI tooling

Generated-by: Claude Opus 5 (1M context)

### DIFF
--- a/src/paimon/core/operation/commit/conflict_detection_test.cpp
+++ b/src/paimon/core/operation/commit/conflict_detection_test.cpp
@@ -572,26 +572,10 @@ TEST_F(ConflictDetectionTest, TestBucketKeepSameCacheEviction) {
     ASSERT_EQ(1U, evicted_partition_buckets.size());
 }
 
-TEST_F(ConflictDetectionTest, TestDeletionVectorsNotSupportedWithBucketUnawareMode) {
-    ASSERT_OK_AND_ASSIGN(
-        std::shared_ptr<TableSchema> table_schema,
-        TableSchema::Create(/*schema_id=*/0, arrow::schema(fields_), /*partition_keys=*/{"f1"},
-                            /*primary_keys=*/{}, /*options=*/{}));
-    ASSERT_OK_AND_ASSIGN(CoreOptions core_options,
-                         CoreOptions::FromMap({{Options::BUCKET, "0"},
-                                               {Options::DELETION_VECTORS_ENABLED, "true"}}));
-    ConflictDetection detection(table_schema, core_options, /*snapshot_manager=*/nullptr,
-                                /*manifest_list=*/nullptr, /*manifest_file=*/nullptr,
-                                /*commit_scanner=*/nullptr, "test_user", "test_table",
-                                /*path_factory=*/nullptr);
-
-    ASSERT_NOK_WITH_MSG(CheckConflicts(detection, /*base_entries=*/{}, /*delta_entries=*/{},
-                                       Snapshot::CommitKind::Append()),
-                        "not yet support dv with BUCKET_UNAWARE mode");
-}
-
-TEST_F(ConflictDetectionTest,
-       TestDeletionVectorsNotSupportedWithResolvedBucketUnawareModeFromMinusOne) {
+TEST_F(ConflictDetectionTest, TestDeletionVectorsAllowedWithBucketUnawareMode) {
+    // an unaware bucket table may carry deletion vectors, told apart by index file name rather
+    // than by bucket. Adding files commits; dropping one needs the pairing this class does not
+    // build and is refused
     ASSERT_OK_AND_ASSIGN(
         std::shared_ptr<TableSchema> table_schema,
         TableSchema::Create(/*schema_id=*/0, arrow::schema(fields_), /*partition_keys=*/{"f1"},
@@ -604,9 +588,38 @@ TEST_F(ConflictDetectionTest,
                                 /*commit_scanner=*/nullptr, "test_user", "test_table",
                                 /*path_factory=*/nullptr);
 
-    ASSERT_NOK_WITH_MSG(CheckConflicts(detection, /*base_entries=*/{}, /*delta_entries=*/{},
-                                       Snapshot::CommitKind::Append()),
-                        "not yet support dv with BUCKET_UNAWARE mode");
+    ASSERT_OK(CheckConflicts(detection, /*base_entries=*/{}, /*delta_entries=*/{},
+                             Snapshot::CommitKind::Append()));
+
+    ASSERT_OK(CheckConflicts(detection, /*base_entries=*/{},
+                             {CreateManifestEntry("added.parquet", FileKind::Add())},
+                             Snapshot::CommitKind::Append()));
+
+    ASSERT_NOK_WITH_MSG(
+        CheckConflicts(detection, {CreateManifestEntry("dropped.parquet", FileKind::Add())},
+                       {CreateManifestEntry("dropped.parquet", FileKind::Delete())},
+                       Snapshot::CommitKind::Append()),
+        "Committing the deletion of data file dropped.parquet is not supported");
+}
+
+TEST_F(ConflictDetectionTest, TestDeletionVectorsDeleteAllowedWithoutBucketUnawareMode) {
+    // the guard above is specific to a table whose deletion vectors the entries cannot reference:
+    // a bucketed table pairs them by bucket, so deleting a data file stays allowed there
+    ASSERT_OK_AND_ASSIGN(
+        std::shared_ptr<TableSchema> table_schema,
+        TableSchema::Create(/*schema_id=*/0, arrow::schema(fields_), /*partition_keys=*/{"f1"},
+                            /*primary_keys=*/{}, /*options=*/{}));
+    ASSERT_OK_AND_ASSIGN(CoreOptions core_options,
+                         CoreOptions::FromMap({{Options::BUCKET, "2"},
+                                               {Options::DELETION_VECTORS_ENABLED, "true"}}));
+    ConflictDetection detection(table_schema, core_options, /*snapshot_manager=*/nullptr,
+                                /*manifest_list=*/nullptr, /*manifest_file=*/nullptr,
+                                /*commit_scanner=*/nullptr, "test_user", "test_table",
+                                /*path_factory=*/nullptr);
+
+    ASSERT_OK(CheckConflicts(detection, {CreateManifestEntry("dropped.parquet", FileKind::Add())},
+                             {CreateManifestEntry("dropped.parquet", FileKind::Delete())},
+                             Snapshot::CommitKind::Append()));
 }
 
 TEST_F(ConflictDetectionTest, TestDeletionVectorsAllowedWithResolvedDynamicBucketMode) {
--- a/src/paimon/core/operation/data_evolution_split_read.cpp
+++ b/src/paimon/core/operation/data_evolution_split_read.cpp
@@ -19,11 +19,13 @@
 #include "paimon/core/operation/data_evolution_split_read.h"
 
 #include <algorithm>
+#include <cassert>
 #include <limits>
 #include <map>
 #include <string>
 #include <string_view>
 #include <thread>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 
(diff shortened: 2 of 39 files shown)


src/paimon/core/operation/commit/conflict_detection_test.cpp
```diff
@@ -572,41 +572,58 @@ TEST_F(ConflictDetectionTest, TestBucketKeepSameCacheEviction) {
...
-    ConflictDetection detection(table_schema, core_options, /*snapshot_manager=*/nullptr,
-                                /*manifest_list=*/nullptr, /*manifest_file=*/nullptr,
-                                /*commit_scanner=*/nullptr, "test_user", "test_table",
-                                /*path_factory=*/nullptr);
+TEST_F(ConflictDetectionTest, TestDeletionVectorsAllowedWithBucketUnawareMode) {
+    // an unaware bucket table, however its bucket resolves, may carry deletion vectors, told
+    // apart by index file name rather than by bucket. Adding files commits; dropping one needs
+    // the pairing this class does not build and is refused
+    for (const char* bucket : {"0", "-1"}) {
+        ASSERT_OK_AND_ASSIGN(
```
lxy-9602: If I understand correctly, bucket cannot be set to 0 in practice, because table creation validation would fail. In that case, it seems `ResolveBucketMode` might not need to support the `bucket == 0` branch.

src/paimon/core/operation/data_evolution_split_read.cpp
```diff
@@ -48,9 +50,84 @@
...
+    /// The interface cannot report a failure, so a failed IsValid is answered with "everything
+    /// is deleted". That is the safe direction for a row count, which subtracts it, but it is a
+    /// guess: the assert keeps it from passing silently where asserts are on.
+    int64_t GetCardinality() const override {
+        Result<RoaringBitmap32> valid = inner_->IsValid(offset_, length_);
+        if (!valid.ok()) {
+            assert(false);
+            return length_;
+        }
+        return length_ - valid.value().Cardinality();
```
lxy-9602: Would it make sense to change the base class to return `Result` instead, so we can avoid `assert(false)`?

test/inte/data_evolution_table_test.cpp
```diff
@@ -2108,6 +2353,618 @@ TEST_P(DataEvolutionTableTest, TestWithRowIds) {
...
+
+    // First row range group: no f0 value lies in [100, 200], but the 300 keeps the group's stats
+    // range straddling the filter so the scan cannot prune it. It therefore reaches the plan
+    // reporting four rows, and contributes none of them to the result.
+    ASSERT_OK(
+        WriteAndCommitGroup(table_path, /*first_row_id=*/0, /*f0_values=*/{0, 1, 2, 300}).status());
+    // second row range group: every f0 value matches
+    ASSERT_OK(WriteAndCommitGroup(table_path, /*first_row_id=*/4,
+                                  /*f0_values=*/{100, 101, 102, 103})
+                  .status());
```
lxy-9602: It looks like we could replace `ASSERT_OK(Func.status())` with `ASSERT_OK(Func())` here as well. Could you please adjust it?

src/paimon/core/utils/data_evolution_utils_test.cpp
```diff
@@ -0,0 +1,85 @@
...
+    ASSERT_FALSE(DataEvolutionUtils::IsNormalFile(blob_file->file_name));
+    ASSERT_FALSE(DataEvolutionUtils::IsNormalFile(vector_store_file->file_name));
+}
+
+TEST(DataEvolutionUtilsTest, TestRetrieveAnchorFileFailsWithoutNormalFile) {
+    auto blob_file_1 = CreateFile("blob-file-1.blob", 1);
+    auto blob_file_2 = CreateFile("blob-file-2.blob", 2);
+    auto vector_store_file = CreateFile("vector-store.vector.parquet", 3);
+
+    ASSERT_NOK(
```
lszskye: ASSERT_NOK_WITH_MSG

lszskye (approved):
+1

src/paimon/core/operation/commit/conflict_detection_test.cpp
SteNicholas: You are right — `bucket = 0` cannot be persisted: `SchemaValidation` rejects `bucket < 1` for non-postpone tables, and Paimon Java has no `0 -> BUCKET_UNAWARE` mapping either (`AppendOnlyFileStore#bucketMode` only maps `-1`; the `BucketMode.UNAWARE_BUCKET = 0` constant is the bucket directory number an unaware table writes to, not an option value). Removed the `"0"` leg from this test. Dropping the `bucket == 0` branch from `ResolveBucketMode` itself is left to a follow-up to keep this change focused, since the branch predates this PR — removing it would also align the hypothetical `bucket = 0` with Java, which resolves it to `HASH_FIXED`.

src/paimon/core/operation/data_evolution_split_read.cpp
SteNicholas: Done — `DeletionVector::GetCardinality` now returns `Result<int64_t>`, and a failed `IsValid` propagates instead of asserting and returning a guess. The ripple stays small because the interface is internal: `BitmapDeletionVector` and the test fake return their infallible values unchanged, and the two production callers (`DeletionFileWriter::Write`, `DataSplitImpl::RawMergedRowCount`) propagate with `PAIMON_ASSIGN_OR_RAISE`.

test/inte/data_evolution_table_test.cpp
SteNicholas: Adjusted — dropped the redundant `.status()` in all 23 places this PR had added it, across `data_evolution_table_test.cpp` and `blob_table_inte_test.cpp` (`ASSERT_OK` goes through `GenericToStatus`, which has `Result<T>` overloads).

src/paimon/core/utils/data_evolution_utils_test.cpp
SteNicholas: Done — both `ASSERT_NOK`s here use `ASSERT_NOK_WITH_MSG` now. The pinned message is byte-identical to the `checkState` message in Java's `DataEvolutionUtils#retrieveAnchorFile`, so the assertion also guards the port staying aligned.

lxy-9602 (approved):
+1

JingsongLi (approved):
+1

PR #194, opened 2026-08-11
feat(core): read source-backed primary-key BTree indexes

### LINKED ISSUE #192: [Feature] Read source-backed primary-key BTree indexes
### Search before asking

### Motivation

Paimon release-2.0.0 (Java) introduced source-backed global scalar indexes for primary-key tables (`pk-btree` / `pk-bitmap`). When an index payload covers a data level, the Java batch planner evaluates indexed predicates against the payload and narrows candidate rows down to physical row ranges inside individual data files, instead of scanning the files in full.

paimon-cpp cannot consume these indexes today: a C++ engine scanning a Paimon 2.0 primary-key table falls back to full file scans even for highly selective predicates.

The existing building blocks are not sufficient on their own:

- The BTree global index payload reader/writer is already byte-compatible with Java, but nothing plans against it for primary-key tables.
- The Data Evolution global index read path addresses rows by table-wide row id, while primary-key source-backed indexes address rows by ordinal inside a validated group of source files, later localized to per-file physical positions. The two address spaces are not interchangeable.
- #179 already landed the metadata carrier (`GlobalIndexMeta._SOURCE_META`, commit-message v12), but nothing decodes or validates the carried source metadata.

### Solution

Implement the primary-key source-backed global scalar-index read path on top of the landed metadata carrier, following the Java release-2.0.0 wire and scan semantics:

1. Decode `PrimaryKeyIndexSourceMeta` v1 using the big-endian layout and Java modified UTF-8 file names, with defensive count validation and trailing-byte rejection.

2. Validate coverage independently for each bucket, indexed field, index type, and positive data level. A level is covered only when exactly one payload exists and its ordered source file names and row counts exactly match all active COMPACT files of that level, with matching field id, index type, payload row count, and row range. Missing, corrupt, duplicate, or mismatched payloads leave that level uncovered and it is scanned normally.

3. Plan ordinary primary-key batch scans against validated groups. Evaluate the indexed part of the predicate once per group, localize group ordinals to file-local physical row positions using source row-count prefixes, and keep deletion files aligned with their data files. An out-of-range group ordinal invalidates all covered files in that group. Invalid file-local positions or results requiring more than 4096 ranges fall back only the affected file.

4. Consume indexed splits in the raw read path by intersecting file-local row ranges with file-index selection and deletion vectors. The reader continues to apply the complete original predicate, so index results cannot change row visibility or query semantics.

5. Wire BTree payload evaluation in this issue. `pk-bitmap` definitions are resolved but conservatively use normal scans until a dedicated reader is available. Vector and full-text definitions are recognized during definition resolution but are not part of the scalar-index planner.

### Anything else?

Per community guidance, no separate PIP is needed because this ports an existing Java capability.

The original three stacked branches have been consolidated into one self-contained implementation in #194.

This issue is scoped to read-path acceleration for ordinary primary-key batch scans. The PR also includes `PkSortedIndexFile::Build`, which can construct one source-backed BTree payload from value-sorted input for tests and future tooling. Automatic payload construction and maintenance during compaction are not included. Streaming scans, read-optimized scans, Data Evolution indexes, and Bitmap / vector / full-text payload readers are also outside this issue's scope.

wangyong9999:
### Purpose

Linked issue: Closes #192

This PR adds C++ read-side support for source-backed primary-key BTree indexes produced by Apache Paimon Java 2.0, together with an internal payload builder used by tooling and tests.

The main changes are:

- Decode and serialize `PrimaryKeyIndexSourceMeta` v1 using its big-endian layout and length-prefixed file-name bytes, with validation for unsupported versions, invalid counts, truncation, and trailing bytes.
- Parse the `pk-btree`, `pk-bitmap`, `pk-vector`, and `pk-full-text` column definitions, including field-scoped JSON options and duplicate or cross-family assignment checks.
- Validate index payload coverage against the active ordered `COMPACT` files of each positive data level, including source names, order, row counts, field ID, index type, complete row range, and known zero delete-row counts.
- Decode BTree file metadata through checked `Result` APIs and reject missing, truncated, type-invalid, one-sided, empty non-null, or reversed boundaries before pruning.
- Normalize predicates in `GlobalIndexEvaluatorImpl` and evaluate indexed predicates conservatively: `AND` may use safely evaluable children, while `OR` requires every branch to be evaluable.
- Evaluate each validated BTree group once, localize group ordinals to file-local physical row ranges, and produce `IndexedSplit` instances.
- Route indexed primary-key splits through the physical-position read path, recheck legacy raw-read eligibility, intersect row ranges with file-index pruning, subtract deletion vectors, and retain the complete original predicate for residual filtering.
- Add `PkSortedIndexFile::Build` for writing one BTree payload from value-sorted source-group input.
- Add Java-generated Parquet and ORC fixtures covering historical snapshots, mixed indexed and unindexed files, multi-file source groups, deletion vectors, partitions, buckets, fallback branches, and disabled-index behavior.

The planner falls back conservatively when metadata or coverage is incomplete, BTree metadata is malformed, any source file has an unknown or nonzero delete-row count, index evaluation fails, positions are invalid, or a file would require more than 4096 ranges. Snapshot-loading and index-evaluation failures are logged before fallback. Indexed splits retain their deletion-file association, and externally supplied indexed splits are rechecked before raw dispatch, so using the index does not change row visibility or query results.

The optimization applies to snapshot-scoped, non-read-optimized primary-key batch scans outside the Data Evolution path. Source splits that are not eligible for file-local reads remain unchanged; generated `IndexedSplit` instances are routed explicitly through the physical-position reader.

Only the BTree payload reader is wired into scan planning. Bitmap, vector, and full-text definitions are recognized but fall back to ordinary scans. Automatic payload construction and maintenance during compaction remain outside this change.

### Tests

- Built `paimon-common-test` and `paimon-core-test` with GCC 8.3.0, C++17, Debug mode, ORC enabled, and `-Wall -Werror`.
- Ran the complete `paimon-common-test` suite: 1,441/1,441 passed.
- Ran the complete `paimon-core-test` suite: 1,728/1,728 passed.
- Ran `paimon-primary-key-sorted-index-inte-test`: 12/12 passed across Parquet and ORC.
- Ran the Apache Paimon Java 2.0 fixture generator test: 1/1 passed.
- Ran `ci/scripts/test_cmake_modules.sh`; all AArch64 `-march` and target-architecture checks passed.
- Ran `pre-commit run --all-files`.
- Ran `check-clang-tidy` on all changed C++ files.
- Ran Apache RAT over all 26 newly added non-fixture files: 26/26 Apache-licensed.
- Ran `git diff --check`.

The integration and regression tests cover snapshots 2/3/5, mixed indexed and unindexed files, multi-source payload localization, deletion vectors with residual predicates, legacy unknown/nonzero delete counts, malformed BTree metadata, safe `AND`/`OR` fallback, disabled indexes, partitions and buckets, fallback-branch routing, and unsupported scored splits.

### API and Format

- Adds public option constants for `pk-btree.index.columns`, `pk-bitmap.index.columns`, `pk-vector.index.columns`, and `pk-full-text.index.columns`.
- Keeps the existing `GlobalIndexer::CreateReader` API. BTree readers create their private four-thread executor only when multiple payloads can submit work; single-payload reads, including primary-key groups, stay inline.
- Introduces no new storage format or protocol. It consumes the existing `GlobalIndexMeta._SOURCE_META` carrier and source-metadata v1 layout used by Java 2.0.
- File-name encoding is compatible with Java `writeUTF` for ASCII and non-NUL BMP UTF-8 names. Complete modified UTF-8 support for NUL and supplementary code points is not included.

### Documentation

Adds `docs/source/user_guide/primary_key_global_index.rst` and links it from the user guide index. The page documents prerequisites, validation rules, fallback behavior, compatibility, and current limitations.

### Generative AI tooling

Generated-by: OpenAI Codex (GPT-5) and Claude Code (Fable 5)


wangyong9999:
cc @lxy-9602 @lucasfang could you please take a look when convenient? Thanks~

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -120,8 +123,9 @@ Result<std::shared_ptr<GlobalIndexReader>> BTreeGlobalIndexer::CreateReader(
         }
         read_buffer_size = static_cast<int32_t>(tmp_buffer_size);
     }
-    // TODO(lisizhuo.lsz): Allow users to specify an executor
-    std::shared_ptr<Executor> executor = CreateDefaultExecutor();
+    // Readers are created per payload group and may coexist for many buckets. Share the
+    // process-wide executor instead of creating a dedicated thread pool for every group.
+    std::shared_ptr<Executor> executor = GetGlobalDefaultExecutor();
     return std::make_shared<LazyFilteredBTreeReader>(read_buffer_size, files, key_type, file_reader,
```
lxy-9602: @lszskye Please evaluate whether using `GlobalDefaultExecutor` here could cause any issues. I’d lean toward passing the same executor to each reader instead.

Also, the executor thread count in Java is based on `GLOBAL_INDEX_THREAD_NUM`, rather than using the machine core count as `GetGlobalDefaultExecutor()` does.

A more complete approach would be to let `C++ GlobalIndexer:` accept an external executor and, like Java, have the scan layer create and pass it in based on `global-index.thread-num`, instead of having `BTreeGlobalIndexer` implicitly choose `GetGlobalDefaultExecutor()`.

src/paimon/common/utils/java_modified_utf8.h
```diff
@@ -0,0 +1,49 @@
...
+namespace paimon {
+/// Converts between standard UTF-8 and the "modified UTF-8" used by Java's
+/// `DataOutputStream#writeUTF` / `DataInputStream#readUTF`:
+/// - U+0000 is encoded as the two-byte sequence 0xC0 0x80 instead of a single zero byte;
+/// - supplementary code points (U+10000 and above) are encoded as a UTF-16 surrogate pair,
+///   each surrogate written as an independent three-byte sequence (CESU-8), instead of the
+///   four-byte standard UTF-8 form.
+class JavaModifiedUtf8 {
+ public:
+    JavaModifiedUtf8() = delete;
```
lxy-9602: This seems to be for binary compatibility of the source data file name in `PrimaryKeyIndexSourceMeta`, but in Java, under what scenarios would the file name contain Chinese characters or other special characters?

src/paimon/core/index/pk/primary_key_index_definition.h
```diff
@@ -0,0 +1,75 @@
...
+    enum class Family {
+        VECTOR,
+        BTREE,
+        BITMAP,
+        FULL_TEXT,
+    };
+
+    PrimaryKeyIndexDefinition(std::string column, int32_t field_id, std::string index_type,
+                              std::map<std::string, std::string> options, Family family)
+        : column_(std::move(column)),
```
lxy-9602, line 41: Could we move the `Family family` parameter before `options`?

src/paimon/common/utils/java_modified_utf8.h
lxy-9602: I think I have a rough understanding of the issue now. It’s not limited to this spot—`FileIndexFormat`, `DataSplit.bucketPath`, and `DeletionFile.path` all have similar problems with Chinese characters when converting `string` to UTF.

I’d suggest removing this part from the current PR for now, and then submitting a separate PR later to fully address all `writeUTF` / `readUTF` related issues.

src/paimon/core/index/pk/primary_key_index_definitions.cpp
```diff
@@ -0,0 +1,209 @@
...
+Status ValidateUniqueColumns(std::set<std::string>* indexed_columns,
+                             const std::vector<std::string>& columns) {
+    for (const std::string& column : columns) {
+        if (!indexed_columns->insert(column).second) {
+            return Status::Invalid(
+                fmt::format("Column '{}' can own at most one primary-key index.", column));
+        }
+    }
+    return Status::OK();
+}
```
lxy-9602, line 92: I’m a bit curious whether `ValidateNoDuplicates` and `ValidateUniqueColumns` could be refactored into a shared helper function, with different error reporting as needed. Also, could we move the output parameter to the end of the parameter list?

src/paimon/core/index/pk/primary_key_index_definitions.cpp
```diff
@@ -0,0 +1,209 @@
...
+    PAIMON_RETURN_NOT_OK(ValidateUniqueColumns(&indexed_columns, bitmap_columns));
+    PAIMON_RETURN_NOT_OK(ValidateUniqueColumns(&indexed_columns, full_text_columns));
+
+    std::vector<PrimaryKeyIndexDefinition> definitions;
+    for (const DataField& field : schema.Fields()) {
+        const std::string& column = field.Name();
+        if (ObjectUtils::Contains(btree_columns, column)) {
+            Result<std::map<std::string, std::string>> definition_options =
+                SortedIndexOptions(options, column, kBTreeOptionFamily, kBTreeAlgorithmPrefix);
+            PAIMON_RETURN_NOT_OK(definition_options.status());
```
lxy-9602: Please use `ASSERT_OK_AND_ASSIGN` instead of calling `PAIMON_RETURN_NOT_OK` first and then accessing `value()`. If there are similar cases elsewhere, could you fix them as well?

src/paimon/core/index/pk/primary_key_index_definitions_test.cpp
```diff
@@ -0,0 +1,214 @@
...
+                                         {"fields.price.pk-btree.index.options", R"({"":"v"})"}}));
+        ASSERT_NOK(PrimaryKeyIndexDefinitions::Create(*schema));
+    }
+}
+
+TEST(PrimaryKeyIndexDefinitionsTest, RejectsDuplicateColumnWithinFamily) {
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<TableSchema> schema,
+                         MakeSchema({{Options::PK_BTREE_INDEX_COLUMNS, "price,price"}}));
+    ASSERT_NOK(PrimaryKeyIndexDefinitions::Create(*schema));
+}
```
lxy-9602, line 183: Please make the error message explicit. I’d recommend using `ASSERT_NOK_WITH_MSG`.

src/paimon/core/index/pk/primary_key_index_source_meta.cpp
```diff
@@ -0,0 +1,198 @@
...
+    out->push_back(static_cast<char>(bits & 0xFF));
+}
+
+void AppendBigEndian64(int64_t value, std::string* out) {
+    auto bits = static_cast<uint64_t>(value);
+    for (int32_t shift = 56; shift >= 0; shift -= 8) {
+        out->push_back(static_cast<char>((bits >> shift) & 0xFF));
+    }
+}
+
```
lxy-9602: Please try to reuse `DataOutputStream`, `DataInputStream`, `MemorySegmentOutputStream` for endianness conversion and data input/output.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
wangyong9999: Updated. `GlobalIndexer::CreateReader` now accepts an external executor. `PrimaryKeyIndexBatchScan` creates one scan-scoped executor from `global-index.thread-num` and shares it across BTree readers; when unset, it preserves the existing C++ CPU-count default. The legacy `GlobalIndexScanImpl` path keeps its outer executor separate to avoid nested synchronous work on the same fixed-size pool. No executor is created when the plan has no valid index group.

src/paimon/core/index/pk/primary_key_index_definitions.cpp
wangyong9999: Updated. Both validation paths now reuse `AddUniqueColumns` with a caller-supplied duplicate-error callback, and the output set parameter is last.

src/paimon/common/utils/java_modified_utf8.h
wangyong9999: Configured `data-file.prefix` can introduce Unicode. However, this is a stream-wide issue, so the PR-local codec has been removed and source metadata now uses the existing stream primitives. A separate change can address `writeUTF` / `readUTF` consistently across `FileIndexFormat`, `DataSplit.bucketPath`, `DeletionFile.path`, and source metadata. The remaining supplementary-code-point limitation is documented.

src/paimon/core/index/pk/primary_key_index_definitions_test.cpp
wangyong9999: Updated. Both duplicate-column tests now use `ASSERT_NOK_WITH_MSG` and check the exact diagnostics.

src/paimon/core/index/pk/primary_key_index_definitions.cpp
wangyong9999: Updated. This is production code, so it now uses `PAIMON_ASSIGN_OR_RAISE` rather than the test-only `ASSERT_OK_AND_ASSIGN`. Similar patterns in the new tests were changed to `ASSERT_OK_AND_ASSIGN` or `EXPECT_OK_AND_ASSIGN`.

src/paimon/core/index/pk/primary_key_index_source_meta.cpp
wangyong9999: Updated. Deserialization now uses `ByteArrayInputStream` with `DataInputStream`, and serialization uses `MemorySegmentOutputStream` for pooled big-endian output. `DataOutputStream` cannot wrap `MemorySegmentOutputStream` without an adapter or extra copy. The string length and bytes remain explicit because `MemorySegmentOutputStream::WriteString` narrows through `int16_t`, while this wire field supports the full `uint16_t` range. Boundary and null-input tests were added.

src/paimon/core/index/pk/primary_key_index_definition.h
wangyong9999: Updated. `Family family` now precedes `options`; the member and accessor order and all call sites were changed consistently.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
```diff
@@ -0,0 +1,115 @@
...
+            "Sorted index build must produce exactly one payload file, but produced {}.",
+            io_metas.size()));
+    }
+    const GlobalIndexIOMeta& io_meta = io_metas[0];
+
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<Bytes> source_meta_bytes, source_meta.Serialize(pool));
+    std::optional<std::string> external_path;
+    if (is_external_path) {
+        external_path = io_meta.file_path;
+    }
```
lxy-9602, line 110: Please change this to something like:

```cpp
if (is_external_path) {
    PAIMON_ASSIGN_OR_RAISE(Path path, PathUtil::ToPath(io_meta.file_path));
    external_path = path.ToString();
}
```

Could we normalize the path here? Regular global index writing does this in `global_index_write_task.cpp`. In Java, `org.apache.paimon.fs.Path` parses and normalizes the URI during construction, so the final stored value is the normalized form from `Path.toString()`.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace paimon {
+Result<std::shared_ptr<IndexFileMeta>> PkSortedIndexFile::Build(
+    const DataField& field, const std::string& index_type,
+    const std::map<std::string, std::string>& options, int32_t data_level,
+    const std::vector<PrimaryKeyIndexSourceFile>& source_files,
+    const std::shared_ptr<arrow::Array>& sorted_values, std::vector<int64_t> sorted_ordinals,
+    const std::shared_ptr<GlobalIndexFileWriter>& file_writer, bool is_external_path,
+    const std::shared_ptr<MemoryPool>& pool) {
+    PAIMON_ASSIGN_OR_RAISE(PrimaryKeyIndexSourceMeta source_meta,
```
lxy-9602, line 48: The current implementation puts all planned sorted data for an entire index file into a single `sorted_values`, which could lead to excessive memory usage. While Java uses an external sort buffer here.

Given the scope of the current PR, I’d suggest adding a TODO to clearly document this as a known issue and plan to fix it in a follow-up PR.

src/paimon/core/operation/raw_file_split_read.h
```diff
@@ -65,16 +64,26 @@ class RawFileSplitRead : public AbstractSplitRead {
...
         const BinaryRow& partition, int32_t bucket,
         const std::vector<std::shared_ptr<DataFileMeta>>& files,
-        DeletionVector::Factory dv_factory);
+        const std::vector<std::optional<DeletionFile>>& deletion_files,
+        const std::optional<std::vector<Range>>& local_row_ranges);
+
+    Result<std::unique_ptr<BatchReader>> CreateReader(
+        const BinaryRow& partition, int32_t bucket,
+        const std::vector<std::shared_ptr<DataFileMeta>>& files, DeletionVector::Factory dv_factory,
+        const std::optional<std::vector<Range>>& local_row_ranges = std::nullopt);
```
lxy-9602: Please avoid using default arguments in production code.

src/paimon/core/operation/raw_file_split_read.cpp
```diff
@@ -64,6 +67,21 @@ RawFileSplitRead::RawFileSplitRead(const std::shared_ptr<FileStorePathFactory>&
...
+        PAIMON_RETURN_NOT_OK(indexed_split->Validate());
+        const std::shared_ptr<DataSplit>& inner_split = indexed_split->GetDataSplit();
+        auto inner_split_impl = std::dynamic_pointer_cast<DataSplitImpl>(inner_split);
+        if (!inner_split_impl) {
+            return Status::Invalid("cannot cast indexed inner split to data_split");
+        }
+        if (inner_split_impl->DataFiles().size() != 1) {
+            return Status::Invalid(
+                "indexed splits with file-local row ranges must contain exactly one file");
+        }
```
lxy-9602, line 110: If the indexed split contains scores, is read supported right now? I don’t see any handling logic for that. If it’s not supported, please fail fast and add a TODO to mark it clearly.

src/paimon/core/operation/raw_file_split_read.h
lxy-9602: Also, I’m not sure it’s necessary to add a new `CreateReader` function. Could we just extend the existing one with a `local_row_ranges` parameter instead?

src/paimon/core/table/source/primary_key_sorted_index_result.cpp
```diff
@@ -0,0 +1,146 @@
...
+Result<std::shared_ptr<DataSplitImpl>> ToSingleFileSplit(
+    const PrimaryKeySortedIndexScan::FilePlan& file) {
+    const std::shared_ptr<DataSplitImpl>& source = file.SourceSplit();
+    std::vector<std::shared_ptr<DataFileMeta>> data_files{file.DataFile()};
+    DataSplitImpl::Builder builder(source->Partition(), source->Bucket(), source->BucketPath(),
+                                   std::move(data_files));
+    builder.WithSnapshot(source->SnapshotId())
+        .WithTotalBuckets(source->TotalBuckets())
+        .IsStreaming(false)
+        .RawConvertible(source->RawConvertible());
```
lxy-9602: It seems in Java, `rawConvertible` is always `false` here. The current implementation would make the scan results inconsistent with Java.

src/paimon/core/table/source/primary_key_index_batch_scan.cpp
```diff
@@ -0,0 +1,298 @@
...
+    auto leaf_predicate = std::dynamic_pointer_cast<LeafPredicate>(predicate);
+    return leaf_predicate != nullptr &&
+           leaf_predicate->GetFunction().GetType() == Function::Type::IS_NOT_NULL;
+}
+
+/// Flattens nested same-function compounds and, inside an AND, removes `f IS NOT NULL`
+/// leaves made redundant by a null-rejecting sibling on the same field. Pruning must not
+/// consider `f IS NULL` as constraining: dropping IS NOT NULL from
+/// "f IS NULL AND f IS NOT NULL" would turn the empty result into the set of null rows.
+Result<std::shared_ptr<Predicate>> NormalizePredicate(const std::shared_ptr<Predicate>& predicate) {
```
lxy-9602: This class contains a lot of complex logic. Please add unit tests to cover various predicate scenarios.

lxy-9602: I’d suggest moving some of the complex func into `.h` as `static` methods so it can be tested at a finer granularity.

src/paimon/core/table/source/primary_key_index_batch_scan.cpp
```diff
@@ -0,0 +1,298 @@
...
+    }
+    if (normalized_children.size() == 1) {
+        return normalized_children[0];
+    }
+    if (is_and) {
+        return PredicateBuilder::And(normalized_children);
+    }
+    return PredicateBuilder::Or(normalized_children);
+}
+}  // namespace
```
lxy-9602: In Java, this `NormalizePredicate` logic seems to live in `GlobalIndexEvaluator`, which also applies to append table queries. I’d suggest following the same approach as Java.

src/paimon/core/table/source/primary_key_sorted_index_scan.h
```diff
@@ -0,0 +1,188 @@
...
+ public:
+    PrimaryKeySortedIndexScan() = delete;
+    ~PrimaryKeySortedIndexScan() = delete;
+
+    /// One active data file and its complete field-local payload groups.
+    class FilePlan {
+     public:
+        FilePlan(std::shared_ptr<DataSplitImpl> source_split, int32_t file_index,
+                 std::map<int32_t, std::shared_ptr<PkSortedIndexGroup>> groups)
+            : source_split_(std::move(source_split)),
```
lxy-9602, line 63: `PkSortedIndexGroup::Create` returns `optional<PkSortedIndexGroup>`, but the call sites seem to use `std::shared_ptr<PkSortedIndexGroup>`. Should we adjust the return type of `Create` for consistency?

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
```diff
@@ -0,0 +1,568 @@
...
+    GREATER_THAN,
+    GREATER_OR_EQUAL,
+    IN,
+    NOT_IN,
+    STARTS_WITH,
+    ENDS_WITH,
+    CONTAINS,
+    LIKE,
+};
+
```
lxy-9602: Could you help clarify the distinction between `QueryOperation` and `Function`? They seem somewhat overlapping to me, so I’m wondering why both are needed separately.

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
```diff
@@ -0,0 +1,568 @@
...
+    }
+
+    std::vector<PrimaryKeyIndexDefinition> scalar_definitions;
+    for (const PrimaryKeyIndexDefinition& definition : definitions) {
+        if (definition.GetFamily() == PrimaryKeyIndexDefinition::Family::BTREE ||
+            definition.GetFamily() == PrimaryKeyIndexDefinition::Family::BITMAP) {
+            scalar_definitions.push_back(definition);
+        }
+    }
+
```
lxy-9602, line 353: Could `ScalarDefinitions` be handling something similar here?

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
```diff
@@ -0,0 +1,548 @@
...
+TEST_F(PrimaryKeySortedIndexScanTest, BuildRejectsDuplicateOrdinals) {
+    std::vector<int64_t> ordinals;
+    ordinals.reserve(kTotalRows);
+    for (int64_t i = 0; i < kTotalRows; i++) {
+        ordinals.push_back(i);
+    }
+    ordinals[1] = 0;
+    ASSERT_NOK_WITH_MSG(BuildPayload(std::move(ordinals)).status(),
+                        "Row id 0 appears more than once");
+}
```
lxy-9602, line 379: `ASSERT_NOK_WITH_MSG(Func(), error_message);` is OK here.

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
```diff
@@ -0,0 +1,548 @@
...
+    std::vector<std::shared_ptr<DataFileMeta>> files = {
+        MakeDataFile("a.parquet", kFileARows, 5, FileSource::Compact())};
+    DataSplitImpl::Builder builder(BinaryRow::EmptyRow(), /*bucket=*/0, base_path_,
+                                   std::move(files));
+    builder.WithSnapshot(kSnapshotId + 1).IsStreaming(false).RawConvertible(true);
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<DataSplitImpl> split, builder.Build());
+    Result<PrimaryKeySortedIndexScan::Plan> plan = PrimaryKeySortedIndexScan::CreatePlan(
+        kSnapshotId, {split}, definitions_, MakeEntries(payload));
+    ASSERT_NOK(plan.status());
+}
```
lxy-9602, line 686: ASSERT_NOK(PrimaryKeySortedIndexScan::CreatePlan(
        kSnapshotId, {split}, definitions_, MakeEntries(payload)));

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
```diff
@@ -0,0 +1,548 @@
...
+
+}  // namespace paimon::test
```
lxy-9602, line 688: The newly added tests mainly cover internal logic such as `CreatePlan`, `Evaluate`, and `ToSplits`. Please also add E2E tests similar to `test/inte/global_index_test.cpp`, covering the complete path:

```text
Java writes the table and PK BTree index
    → C++ TableScan::CreatePlan
    → IndexedSplit / DataSplit
    → TableRead::CreateReader
    → validate the returned rows
```

These tests should not only invoke `PkSortedIndexFile::Build`, `PrimaryKeySortedIndexScan::Evaluate`, or mock readers directly. Otherwise, they do not cover Java/C++ format compatibility, manifest/source metadata parsing, index reader creation, split conversion, and actual data-file reading.

Please consider covering the following scenarios:

1. **Index written by Java and read by C++**

   Let Java write a primary-key table with a `pk-btree` index, then query it from C++ using a predicate on the indexed field.

   In addition to validating the returned rows, verify that the scan plan contains an `IndexedSplit` with the expected single data file and file-local row ranges.

2. **A mixture of indexed and non-indexed files**

   First compact the table to generate indexed files, then append new L0 files without another compaction. The query range should match rows from both:

   - a compacted file covered by a PK BTree index;
   - a newly appended L0 file without an index.

   Verify that the plan contains both `IndexedSplit` and regular `DataSplit` instances, and that the final read returns rows from both groups without false negatives.

3. **One index payload covering multiple source data files**

   Use Java to perform a compaction that produces multiple data files while creating only one PK BTree index group. The query result should span at least two source files.

   Verify that C++ correctly converts group-global row ordinals into file-local row ranges and creates the corresponding single-file `IndexedSplit` instances. Execute `TableRead` and verify that matching rows from all source files are returned.

4. **Deletion Vector together with a residual predicate**

   Let Java compact the table and build the index, then perform updates or deletes to generate a deletion vector. Use a predicate containing both an indexed and a non-indexed field, for example:

   ```text
   score = 10 AND tag = 'keep'
   ```

   Verify that the PK BTree row ranges, deletion vector, and complete residual predicate are all applied. Deleted rows and rows that fail the residual predicate must not be returned.

5. **Safe fallback for AND/OR predicates**

   Test both:

   ```text
   score = 10 AND tag = 'keep'
   score = 10 OR tag = 'keep'
   ```

   For the AND predicate, the `score` index may narrow the row ranges while `tag` remains a residual predicate.

   For the OR predicate, if one branch cannot be evaluated by the index, the scan must not prune data using only the indexed branch. It should safely fall back to a regular scan. Both cases should execute the complete scan-and-read path and validate the final rows.

6. **Global index disabled**

   Set the scan option:

   ```cpp
   {{Options::GLOBAL_INDEX_ENABLED, "false"}}
   ```

   or equivalently:

   ```text
   global-index.enabled=false
   ```

   Verify that no `IndexedSplit` is produced, the regular scan/read path is used, and the returned rows are identical to those returned with the index enabled.

7. **Historical snapshots and index versions**

   Read the following snapshots from the same Java-generated fixture:

   - snapshot 2: use the index generated by the first compaction;
   - snapshot 3: use the old index for the compacted file and fall back for the new L0 file;
   - snapshot 5: use the index rebuilt by the later compaction.

   Validate both the plan and the final rows for each snapshot. This should ensure that C++ only applies index payloads belonging to the requested snapshot and matching the exact source files, rather than incorrectly applying the snapshot 5 index to files from snapshot 2 or 3.

8. **Partitions and multiple buckets**

   Prepare a Java fixture with multiple partitions and multiple buckets, with PK BTree index groups in different buckets. The query should match rows across partitions and buckets.

   Verify that each `IndexedSplit` has the correct partition, bucket, source data file, and index-file path, and validate the final read result.

9. **Routing `IndexedSplit` through `FallbackTableRead`**

   Configure `scan.fallback-branch` and create a scenario where the main branch produces an `IndexedSplit`.

   Verify that:

   - an `IndexedSplit` from the main branch is sent to the main table reader;
   - a split from the fallback branch is sent to the fallback table reader;
   - an `IndexedSplit` is not rejected or routed to the wrong reader because of its split type;
   - the final read result is correct.

10. **Both Parquet and ORC**

    Please generate equivalent Java fixtures for Parquet and ORC and parameterize the tests with `TEST_P`, for example:

    ```cpp
    using ParamType = std::string;

    INSTANTIATE_TEST_SUITE_P(
        FileFormat,
        PrimaryKeySortedIndexE2ETest,
        ::testing::Values("parquet", "orc"));
    ```

    Every case should cover the actual `TableScan + TableRead` path rather than only validating the plan. Depending on the scenario, the tests should also verify the split type, source data file, file-local row ranges, and deletion file.

These tests could be added to a dedicated `primary_key_sorted_index_inte_test.cpp` or to the existing `global_index_test.cpp`. The Java-generated fixtures should be stored under `test/test_data/{parquet,orc}/...`, with a README documenting the schema, writes, compactions, and expected state of each snapshot.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
wangyong9999: Updated. External payload paths are now normalized through PathUtil::ToPath and Path::ToString, with regression coverage for duplicate separators.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
wangyong9999: Added an author-tagged TODO to replace the in-memory input with an external sort buffer and bounded writer batches.

src/paimon/core/operation/raw_file_split_read.h
wangyong9999: Updated. The existing overloads now take local_row_ranges explicitly; the forwarding overload and production default argument were removed, and callers pass std::nullopt explicitly.

src/paimon/core/operation/raw_file_split_read.cpp
wangyong9999: Updated. Primary-key reads now fail fast for scored IndexedSplits before routing or force-keep-delete fallback; a TODO and regression coverage were added.

src/paimon/core/table/source/primary_key_sorted_index_result.cpp
wangyong9999: Updated. Derived single-file splits now use rawConvertible(false), and IndexedSplits are routed explicitly to the physical-position reader while unindexed splits use the normal merge path.

src/paimon/core/table/source/primary_key_index_batch_scan.cpp
wangyong9999: Added focused GlobalIndexEvaluatorImpl tests for supported and unsupported leaves, compound-predicate flattening, and safe AND/OR semantics, including IS NULL and IS NOT NULL.

src/paimon/core/table/source/primary_key_index_batch_scan.cpp
wangyong9999: Updated. Predicate normalization now runs in GlobalIndexEvaluatorImpl, so append and primary-key index scans share the same behavior.

src/paimon/core/table/source/primary_key_sorted_index_scan.h
wangyong9999: Updated. Create now returns std::shared_ptr<PkSortedIndexGroup>, and bucket state and file plans reuse the same group instance.

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
wangyong9999: Agreed. QueryOperation was only a cache-key tag, so it has been removed and the cache now uses Function::Type directly.

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
wangyong9999: Updated. Scalar-family filtering is centralized in PrimaryKeyIndexDefinitions::ScalarDefinitions and reused by planning and evaluation.

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
wangyong9999: Updated. The Result is now passed directly to ASSERT_NOK_WITH_MSG.

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
wangyong9999: Updated. ASSERT_NOK now checks CreatePlan directly.

src/paimon/core/table/source/primary_key_sorted_index_scan_test.cpp
wangyong9999: Added Java-generated Parquet/ORC fixtures and parameterized TableScan-to-TableRead E2E coverage for indexed plus visible unindexed APPEND-source files, multi-file payloads and file-local ranges, deletion vectors with AND residual filtering, safe OR fallback, disabled indexes, snapshots 2/3/5, partitions and buckets, and fallback-branch routing. The fixture READMEs document the exact snapshot construction.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
wangyong9999: Follow-up correction: each primary-key index group currently contains exactly one payload, and `UnionGlobalIndexReader` evaluates a single reader inline, so the scan-scoped executor could never receive work. It has been removed, and the PK path now explicitly passes `nullptr`. The four-argument BTree reader retains its pre-PR private four-thread executor behavior for `GlobalIndexScanImpl`; `global-index.thread-num` is not applied to the single-payload PK path.

include/paimon/global_index/global_indexer.h
```diff
@@ -70,6 +72,27 @@ class PAIMON_EXPORT GlobalIndexer {
...
+    ///                       that the reader should evaluate sequentially.
+    /// @return A `Result` containing a shared pointer to the created `GlobalIndexReader`,
+    ///         or an error if the index cannot be loaded or is incompatible, etc.
+    virtual Result<std::shared_ptr<GlobalIndexReader>> CreateReader(
+        ::ArrowSchema* arrow_schema, const std::shared_ptr<GlobalIndexFileReader>& file_reader,
+        const std::vector<GlobalIndexIOMeta>& files, const std::shared_ptr<MemoryPool>& pool,
+        const std::shared_ptr<Executor>& executor) const {
+        static_cast<void>(executor);
+        return CreateReader(arrow_schema, file_reader, files, pool);
+    }
```
lxy-9602: Since pk is currently always passed as `nullptr`, is it really necessary to add this extra public interface with the executor?

wangyong9999: Good point. PK groups contain one payload, so the executor overload was unnecessary. I removed it and kept the existing four-argument API; BTree now creates its private executor only for multiple payloads, preserving legacy parallel evaluation while single-payload PK reads stay inline.

lxy-9602 (approved):
+1

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
```diff
@@ -0,0 +1,549 @@
...
+        const std::shared_ptr<IndexFileMeta>& payload = group.Payload();
+        const std::optional<GlobalIndexMeta>& payload_meta = payload->GetGlobalIndexMeta();
+        if (payload_meta == std::nullopt) {
+            // Group validation guarantees the metadata; degrade to a normal scan of the
+            // covered files if it is ever violated, like the Java reader factory.
+            return Status::Invalid(fmt::format(
+                "Sorted index payload {} has no global index metadata.", payload->FileName()));
+        }
+        std::vector<GlobalIndexIOMeta> io_metas;
+        io_metas.emplace_back(path_factory->ToPath(payload), payload->FileSize(),
```
SteNicholas, line 551: Reject missing or malformed BTree metadata. `_INDEX_META` is nullable, and the group validation currently accepts a null payload, but this value is passed directly to `BTreeFileMetaSelector`. Its eager deserialization dereferences null metadata and reads fields without checking the buffer length, so an incomplete persisted index can crash the process instead of triggering the intended per-file fallback. Validate the BTree metadata before creating the reader—ideally by making deserialization return a checked `Result`—and mark the index group unusable when validation fails.

src/paimon/core/table/source/key_value_table_read.cpp
```diff
@@ -74,7 +75,33 @@ void KeyValueTableRead::ForceKeepDelete(bool force_keep_delete) {
...
+            return Status::NotImplemented(
+                "Primary-key reads do not support scored indexed splits yet.");
+        }
+        // Primary-key indexed splits carry physical positions and are routed independently
+        // of the inner split's raw-convertible marker, matching Java's dedicated provider.
+        const std::shared_ptr<DataSplit>& inner_split = indexed_split->GetDataSplit();
+        if (!force_keep_delete_) {
+            for (const auto& read : split_reads_) {
+                if (dynamic_cast<RawFileSplitRead*>(read.get()) != nullptr) {
+                    return read->CreateReader(indexed_split);
```
SteNicholas: Preserve the legacy raw-read safety check. This direct dispatch bypasses `RawFileSplitRead::Match`, whose primary-key path deliberately rejects files with an unknown `delete_row_count` because legacy metadata cannot prove that they contain no DELETE rows. `MergeTreeSplitGenerator` still treats a missing count as raw-convertible, while the new index-source policy checks only the COMPACT source and level, so an indexed legacy file can reach this path and expose tombstones or obsolete rows instead of using merge semantics. Require a known zero delete count for every indexed file, or reuse the existing match guard and fall back to the inner merge read.

SteNicholas (commented):
@wangyong9999, I have found two correctness issues in the source-backed primary-key BTree index read path. PTAL.

src/paimon/core/table/source/primary_key_sorted_index_scan.cpp
wangyong9999: Fixed in 30ea746. BTree metadata is now checked before reader creation; malformed metadata returns an error and the PK scan falls back safely.

src/paimon/core/table/source/key_value_table_read.cpp
wangyong9999: Fixed in 30ea746. Indexed planning now requires a known zero delete count for every file in the source split, and indexed splits are rechecked before raw dispatch; otherwise reading falls back to merge semantics.

SteNicholas (approved):
LGTM.

PR #199, opened 2026-08-13
feat(realtime): support two-stage scan and read

lxy-9602:
### Purpose

Linked issue: #158 

This PR extends the real-time append read path introduced in #163 to support separate scan and
reader creation stages.

The main changes are:

- capture immutable memory views during scan planning and register them in `RealtimeContext`;
- add version, snapshot id, partition-bucket, committed offset, memory upper offset, and an opaque
  read-view ticket to `RealtimeSplit`;
- allow `ReadContext` to share the same `RealtimeContext` used by the writer and scanner;
- resolve the ticket when creating a reader and validate that its partition-bucket and offset range
  match the split;
- use `RealtimeReader` to keep the memory view alive while disk and memory data are being read;
- release abandoned or completed read views through a background release queue, avoiding large
  segment destruction on the query thread;
- preserve an already-created reader across commit, refresh, and memory-segment reclamation.

A real-time split ticket is currently single-use. Once reader creation consumes the ticket, retrying
the same reader task requires creating a new scan plan and `RealtimeSplit`. Tickets are also
process-local and must be resolved by the `RealtimeContext` that created them.

### Tests

- pinning, resolving, and releasing a read-view ticket;
- expiration and cleanup of abandoned read-view tickets;
- rejection when the same `RealtimeSplit` ticket is reused;
- reading from a reader created before commit and refresh after its memory segment is reclaimed;
- reading a pinned plan after writer close;
- updated Arrow memory-indexer reader ownership.

### API and Format
This PR changes public APIs:

- adds `ReadContextBuilder::WithRealtimeContext`;
- adds read-view ticket lifecycle APIs to `RealtimeContext`;
- adds `RealtimeReader`, which keeps a `MemReadView` alive for the reader lifetime;
- changes `MemIndexer::CreateQueryReaders` to return `RealtimeReader` instances.

There is no change to the Paimon table storage format, data-file format, snapshot format, or commit
protocol.

`RealtimeSplit` carries versioned real-time metadata, but cross-process split serialization is not
implemented in this PR. Its opaque ticket is currently resolved only inside the originating process.

### Generative AI tooling


include/paimon/defs.h
```diff
@@ -540,6 +540,10 @@ struct PAIMON_EXPORT Options {
     /// "scan.timestamp" can be used as an alternative string input for the same mode.
     static const char SCAN_TIMESTAMP_MILLIS[];
 
+    /// "realtime.read-view-ttl-millis" - Lifetime in milliseconds of a real-time memory view
+    /// pinned by scan planning before reader creation. Default value is 300000.
```
zjw1111: 300000 -> use a human-readable value? 300s or 5min

src/paimon/core/core_options.cpp
```diff
@@ -802,6 +803,12 @@ struct CoreOptions::Impl {
         std::string scan_timestamp_str;
         PAIMON_RETURN_NOT_OK(parser.Parse(Options::SCAN_TIMESTAMP, &scan_timestamp_str));
         PAIMON_RETURN_NOT_OK(parser.Parse(Options::SCAN_TIMESTAMP_MILLIS, &scan_timestamp_millis));
+        PAIMON_RETURN_NOT_OK(
+            parser.Parse(Options::REALTIME_READ_VIEW_TTL_MILLIS, &realtime_read_view_ttl_millis));
```
zjw1111: use `ParseTimeDuration`

zjw1111: and `read-view-ttl-millis` -> `read-view-ttl`

src/paimon/core/realtime/realtime_context.cpp
```diff
@@ -105,6 +127,79 @@ class RealtimeContext::Impl {
...
+            auto iter = pinned_read_views_.find(opaque_ticket);
+            if (iter == pinned_read_views_.end()) {
+                return Status::Invalid("real-time read-view ticket does not exist or has expired");
+            }
+            if (iter->second.expire_at <= std::chrono::steady_clock::now()) {
+                read_view_release_queue_.push_back(std::move(iter->second.view));
+                pinned_read_views_.erase(iter);
+                expired = true;
+            } else {
+                return iter->second.view;
```
zjw1111: Resolving the ticket does not claim or remove it atomically. Two threads calling `CreateReader` for the same `RealtimeSplit` can both return the view here before either reaches `ReleaseReadView`, so both can create readers despite the documented single-use contract. Please make ticket acquisition/consumption atomic (for example, with an available/claimed state that is rolled back when reader creation fails) and add concurrent coverage.

zjw1111 (commented):
Concurrent consumption of the same real-time split needs to be prevented.

src/paimon/core/realtime/realtime_context_test.cpp
```diff
@@ -113,6 +128,15 @@ std::unique_ptr<ArrowSchema> MakeWriteSchema() {
     return c_schema;
 }
 
+TEST(RealtimeReaderTest, TestRejectsIncompleteReader) {
```
zjw1111: the source is in `mem_indexer.h` but the test is in `realtime_context_test.cpp`

include/paimon/realtime/mem_indexer.h
```diff
@@ -74,6 +75,53 @@ class PAIMON_EXPORT MemReadView {
     virtual std::optional<Range> GetOffsetRange() const = 0;
 };
 
+/// Outermost query reader that pins a `MemReadView`.
+class PAIMON_EXPORT RealtimeReader final : public BatchReader {
```
zjw1111: Is it needed to put in include dir?

src/paimon/core/table/source/append_only_table_read.cpp
```diff
@@ -65,9 +67,27 @@ Result<std::unique_ptr<BatchReader>> AppendOnlyTableRead::CreateReader(
...
+    if (!realtime_context) {
+        return Status::Invalid("reading a real-time split requires a real-time context");
+    }
+    PAIMON_ASSIGN_OR_RAISE(RealtimePartitionBucketView memory,
+                           realtime_context->ResolveReadView(realtime_split->OpaqueTicket()));
     std::vector<std::unique_ptr<BatchReader>> readers;
     readers.reserve(realtime_split->DiskSplits().size() + 1);
+    const RealtimePartitionBucket expected_partition_bucket(realtime_split->Partition(),
+                                                            realtime_split->Bucket());
+    if (!(memory.partition_bucket == expected_partition_bucket)) {
```
zjw1111: add a `!=` operator for `RealtimePartitionBucket`

include/paimon/realtime/realtime_context.h
zjw1111: These 6 methods is not used by external user(include dir), maybe move to impl in src dir?

src/paimon/core/realtime/realtime_context.cpp
lxy-9602: Added comments.

src/paimon/core/realtime/realtime_context_impl.h
```diff
@@ -0,0 +1,107 @@
...
+    int64_t initial_offset;
+};
+
+struct RealtimePartitionBucketView {
+    RealtimePartitionBucket partition_bucket;
+    std::shared_ptr<MemIndexer> indexer;
+    std::shared_ptr<MemReadView> read_view;
+};
+
+class RealtimeContextImpl final : public RealtimeContext {
```
SteNicholas: This class is compiled under the repository-wide `-fvisibility=hidden`, but `realtime_context_test.cpp` and `realtime_write_inte_test.cpp` call its out-of-line methods while their executables link against `paimon_shared`. Those methods are therefore private/hidden in the shared library (for example `Cast`, `GetOrCreateMemIndexer`, and `ResolveReadView`), so the test link will fail with undefined symbols. Please either export the test-used class/symbols, link this implementation into those tests, or exercise it through public APIs.

src/paimon/core/table/source/append_only_table_read.cpp
```diff
@@ -80,19 +104,21 @@ Result<std::unique_ptr<BatchReader>> AppendOnlyTableRead::CreateReader(
...
             PAIMON_ASSIGN_OR_RAISE(memory_reader, PredicateBatchReader::Create(
                                                       std::move(memory_reader),
                                                       context_->GetPredicate(), GetMemoryPool()));
         }
-        readers.push_back(std::move(memory_reader));
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<RealtimeReader> realtime_reader,
+                               RealtimeReader::Create(memory.read_view, std::move(memory_reader)));
+        readers.push_back(std::move(realtime_reader));
     }
+    PAIMON_RETURN_NOT_OK(realtime_context_impl->ReleaseReadView(realtime_split->OpaqueTicket()));
```
SteNicholas: This consumes the ticket before the public `TableRead::CreateReader(vector<Split>)` call has succeeded as a whole. With two real-time splits, split 1 can release its ticket here and split 2 can then fail (for example, a plugin error or ticket expiry); the vector overload returns an error, but retrying that failed call now fails on split 1 because its ticket is gone. Please make the ticket claims/consumption rollbackable across the vector overload, or defer consumption until all per-split readers have been created, and add a two-split regression test.

lxy-9602: Good catch! Fixed.

src/paimon/core/realtime/realtime_context_impl.h
lxy-9602: Thanks for the concern. The current paimon-cpp test setup allows tests to access internal classes and non-public methods, which is already an established pattern in the repository. Therefore, these implementation details do not need to be exported as public APIs solely for testing.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

src/paimon/core/table/source/realtime_split.h
```diff
@@ -59,25 +74,27 @@ class RealtimeSplit : public Split {
...
 
  private:
+    int32_t version_;
+    std::optional<int64_t> snapshot_id_;
     std::map<std::string, std::string> partition_;
     int32_t bucket_;
     std::vector<std::shared_ptr<Split>> disk_splits_;
-    std::shared_ptr<MemIndexer> indexer_;
-    std::shared_ptr<MemReadView> read_view_;
     int64_t committed_offset_;
```
luoyuxia: Would it be more intuitive to represent committed progress as an exclusive end offset? The committed snapshot covers offsets < committed_end_offset, and the reader starts from committed_end_offset inclusively. This avoids the extra +1 conversion, represents an empty committed range naturally with 0, and makes adjacent ranges compose cleanly as half-open intervals.

Correspondingly, could memory_upper_offset be represented as an exclusive memory_end_offset? At the RealtimeSplit level, the in-memory range could then be expressed directly as [start_offset, end_offset): start_offset equals committed_end_offset, and end_offset equals memory_end_offset. The reader consumes exactly that range, which makes both boundaries explicit and avoids mixing inclusive committed_offset and memory_upper_offset semantics.

include/paimon/realtime/realtime_context.h
```diff
@@ -66,29 +65,12 @@ struct PAIMON_EXPORT RealtimePartitionBucket {
...
 /// Shared context that owns the `MemIndexer` instances used by a real-time writer.
 ///
-/// Applications share one context between `WriteContext` and `ScanContext`. The context uses
-/// either the default Arrow implementation or an application-provided factory and keeps each
-/// created indexer available across writes, prepare-commit operations, and process-local reads.
+/// Applications share one context between `WriteContext`, `ScanContext`, and `ReadContext`. The
+/// context uses either the default Arrow implementation or an application-provided factory and
+/// keeps each created indexer available across writes, prepare-commit operations, and
+/// process-local reads.
 class PAIMON_EXPORT RealtimeContext {
```
luoyuxia, line 76: The current RealtimeContext hierarchy seems unclear. RealtimeContext::Create() constructs a RealtimeContextImpl and upcasts it to the public base type, while  the write, scan, and read paths immediately downcast it through RealtimeContextImpl::Cast() before performing any operation.
As a result, the effective dependency is still RealtimeContextImpl; RealtimeContext only transports the concrete instance and does not define a behavioral contract. If only one implementation is intended, a concrete RealtimeContext with Pimpl would make the call hierarchy clearer. Otherwise, the operations required by the write, scan, and read paths should be expressed on the shared abstraction and invoked without downcasting.

The name RealtimeContext also seems misleading for its current role. Unlike WriteContext, ScanContext, or ReadContext, this object is not an operation-scoped collection of configuration and dependencies. It is a long-lived, stateful object shared by the writer, scanner, and reader, and it owns the mem-indexer  registry, committed progress, pinned read views, and their lifecycle. Would RealtimeStore better describe this ownership model and make the relationship with writer, scan, and read more explicit?

This structure also makes it difficult to support another realtime-store implementation. The current extension point is MemIndexerFactory, which can replace the per-partition-bucket memory indexer, but it cannot replace the writer, scan/planning, split, and reader behavior of the realtime store as a whole. A different  RealtimeContext implementation cannot be used through the existing APIs because all write, scan, and read paths eventually require RealtimeContextImpl::Cast(). Supporting another store would therefore require modifying these core call sites and adding implementation-specific branches.

Would it be clearer to make RealtimeStore the behavioral boundary and introduce a RealtimeStoreFactory that creates the complete store implementation? The  existing mem-indexer-based implementation could remain one RealtimeStore, while another store could provide its own writer, planner, split, and reader behavior  without changing the core dispatch path.

luoyuxia: I am not fully sure whether the following API shape covers all of your intended use cases, so please treat it as a design sketch for discussion.

The overall object model is inspired by Paimon’s Java CatalogFactory → Catalog → Table creation model. This is an object-creation and ownership hierarchy rather than inheritance between these three types: a factory creates a backend-scoped store, the store loads a table-specific handle, and the table handle creates or wraps the write, scan, and read operations.

Please note the follow api proposal is is roughly the API shape I have in mind. I have not examined every related code path in detail, so the concrete class names and method signatures may not be completely accurate. The main point is the overall creation and composition hierarchy:
```
RealtimeStoreFactory
      └── creates RealtimeStore
              └── loads RealtimeTable
                      ├── creates write operations
                      ├── creates scan operations
                      └── creates read operations
```
Concrete backends implement the corresponding factory, store, and table interfaces. The exact API may differ, but this is roughly the abstraction boundary and dependency direction I have in mind.

## Propose API
```
/// Store-level configuration shared by all tables opened from one realtime store.
struct PAIMON_EXPORT RealtimeStoreOptions {
      std::map<std::string, std::string> options;
      xxx
  };

/// Identifies and configures one table loaded from a RealtimeStore.
  struct PAIMON_EXPORT RealtimeTableDescriptor {
      Identifier identifier;
      std::string table_path;
      std::string branch;
      std::shared_ptr<Schema> schema;
      std::map<std::string, std::string> options;
  };

/// Factory for creating a backend-level RealtimeStore.
  ///
  /// Implementations may be registered through the existing Paimon Factory
  /// mechanism using a unique identifier.
  class PAIMON_EXPORT RealtimeStoreFactory : public Factory {
   public:
      ~RealtimeStoreFactory() override = default;

      /// Creates a RealtimeStore using this factory.
      virtual Result<std::shared_ptr<RealtimeStore>> Create(
          const RealtimeStoreOptions& options) const = 0;
  };

/// A long-lived realtime backend that may serve multiple tables.
  ///
  /// Implementations may own backend-wide resources. Table-specific state
  /// should be represented by RealtimeTable.
  class PAIMON_EXPORT RealtimeStore {
   public:
      virtual ~RealtimeStore() = default;

      /// Loads or opens a table from this realtime store.
      ///
      /// The returned table may retain shared backend resources owned by this
      /// store. Implementations may cache table state internally.
      virtual Result<std::shared_ptr<RealtimeTable>> LoadTable(
          const RealtimeTableDescriptor& table) = 0;
  };

/// A table-specific realtime handle.
  ///
  /// The same table handle may be shared by write, scan, and read operations.
  /// An in-process implementation may keep table-level indexers, committed
  /// progress, and query views here. A remote implementation may keep only
  /// table metadata and backend clients.
  class PAIMON_EXPORT RealtimeTable {
   public:
      virtual ~RealtimeTable() = default;

      /// Creates the write path for this realtime table.
      ///
      /// The returned object follows the existing FileStoreWrite contract,
      /// including prepare-commit and committed-snapshot notification.
      virtual Result<std::unique_ptr<FileStoreWrite>> CreateWrite(
          std::unique_ptr<WriteContext> context) = 0;

      /// Creates the scan/planning path for this realtime table.
      ///
      /// CreatePlan() may combine the committed disk snapshot with the
      /// realtime portion and return implementation-specific Split objects.
      virtual Result<std::unique_ptr<TableScan>> CreateScan(
          std::unique_ptr<ScanContext> context) = 0;

      /// Creates the read path for splits planned by CreateScan().
      ///
      /// The implementation is responsible for recognizing its own realtime
      /// splits and composing disk and realtime readers as required.
      virtual Result<std::unique_ptr<TableRead>> CreateRead(
          std::unique_ptr<ReadContext> context) = 0;
  };
```
## Example usage
```
RealtimeStoreOptions store_options;
store_options.options = options;

PAIMON_ASSIGN_OR_RAISE(
      std::shared_ptr<RealtimeStore> realtime_store,
      RealtimeStoreFactory::Create(realtime_store_type, store_options));

  RealtimeTableDescriptor table_descriptor{
      identifier,
      table_path,
      branch,
      schema,
      table_options};

  PAIMON_ASSIGN_OR_RAISE(
      std::shared_ptr<RealtimeTable> realtime_table,
      realtime_store->LoadTable(table_descriptor));

  Create a writer:

  WriteContextBuilder write_builder(table_path, commit_user);
  write_builder.SetOptions(write_options);

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<WriteContext> write_context,
      write_builder.Finish());

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<FileStoreWrite> writer,
      realtime_table->CreateWrite(std::move(write_context)));

  Create a two-stage scan and read:

  ScanContextBuilder scan_builder(table_path);
  scan_builder.SetOptions(scan_options);

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<ScanContext> scan_context,
      scan_builder.Finish());

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<TableScan> scan,
      realtime_table->CreateScan(std::move(scan_context)));

  PAIMON_ASSIGN_OR_RAISE(
      std::shared_ptr<Plan> plan,
      scan->CreatePlan());

  ReadContextBuilder read_builder(table_path);
  read_builder.SetOptions(read_options);

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<ReadContext> read_context,
      read_builder.Finish());

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<TableRead> table_read,
      realtime_table->CreateRead(std::move(read_context)));

  PAIMON_ASSIGN_OR_RAISE(
      std::unique_ptr<BatchReader> reader,
      table_read->CreateReader(plan->Splits()));

  Possible implementation hierarchy:

  RealtimeStoreFactory
  ├── MemoryRealtimeStoreFactory
  └── OtherRealtimeStoreFactory

  RealtimeStore
  ├── MemoryRealtimeStore
  │     └── LoadTable() → MemoryRealtimeTable
  └── OtherRealtimeStore
        └── LoadTable() → OtherRealtimeTable

  RealtimeTable
  ├── MemoryRealtimeTable
  │     ├── CreateWrite() → mem-indexer-backed FileStoreWrite
  │     ├── CreateScan()  → memory union TableScan
  │     └── CreateRead()  → memory union TableRead
  └── OtherRealtimeTable
        ├── CreateWrite() → backend-specific FileStoreWrite
        ├── CreateScan()  → backend-specific TableScan
        └── CreateRead()  → backend-specific TableRead
```

With this structure, the current mem-indexer implementation could become MemoryRealtimeStore and MemoryRealtimeTable. MemIndexerFactory would remain an implementation detail of that backend.

The write, scan, and read paths would invoke the RealtimeTable interface directly. They would no longer upcast a concrete implementation to RealtimeContext and  later recover it through RealtimeContextImpl::Cast(). This proposal assumes that every realtime-table implementation exposed through paimon-cpp can conform to the existing FileStoreWrite, TableScan, and TableRead contracts. If that assumption does not match the intended writer or commit lifecycle, the operation interfaces may need to be separated further.

luoyuxia (commented):
@lxy-9602 Thanks for the pr. The direction is sound. I left some comments from api design.

include/paimon/realtime/realtime_context.h
lxy-9602: Thank you very much for the review. At the moment, the entire paimon-cpp library is designed around operation-level APIs. In typical usage, callers create a `ReadContext` / `WriteContext` and then construct a reader / writer from that; there is currently no table-level abstraction or usage pattern. To keep this PR focused and clear, I would prefer to preserve the current design in this change and support realtime use cases via shared read/write context or store. A follow-up PR can introduce a table-level abstraction, for example together with catalog integration.

luoyuxia: That makes sense to me for the scope of this PR. Keeping the operation-level API is consistent with the current paimon-cpp design, and I do not think a table-level abstraction needs to be introduced in this change.

My main concern is future extensibility. I hope the API can eventually allow other realtime context/store implementations to integrate without requiring implementation-specific changes across the write, scan, and read paths. This does not need to be addressed in this PR, but it would be good to consider it in the follow-up API design.

src/paimon/core/table/source/realtime_split.h
lxy-9602: The reason I originally chose a closed interval on both ends was to reuse the existing Range class in Java/C++, but I agree that it introduces some additional complexity in both understanding and implementation. I’ll remove the realtime code’s dependency on Range and replace it with support based on a left-closed, right-open OffsetRange type.

include/paimon/realtime/realtime_context.h
lxy-9602: `RealtimeContext` is not actually intended to be customizable—at least, there is no such requirement at the moment. The customizable parts, namely in-memory storage and retrieval, are already encapsulated in `MemIndexer`. I’ll adjust the implementation to avoid the ambiguity in the current header file.

zjw1111: Thanks for the detailed analysis. You are right that `RealtimeContext` currently only transports the concrete `RealtimeContextImpl` and does not define a behavioral contract of its own.

That said, this is a general design question in our codebase rather than something introduced by this PR. `Split`, `Predicate` and several other public types follow the same shape: an abstract public base plus a downcast to the concrete implementation at the call sites. Reworking the hierarchy only for `RealtimeContext` here would leave the codebase inconsistent and would grow the scope of this PR considerably.

So we would prefer not to address it in this PR. We will discuss this hierarchy design as a whole and then apply a consistent change across all of these types in follow-up PRs. Would that work for you?

luoyuxia (approved):

src/paimon/core/realtime/realtime_context_impl.cpp
```diff
@@ -0,0 +1,355 @@
...
+            // offset. Continue after all retained rows instead of restarting across a seal.
+            if (memory_range->end > initial_offset) {
+                initial_offset = memory_range->end;
+            }
+        }
+        return RealtimeStoreState{iter->second, initial_offset};
+    }
+    Result<std::shared_ptr<RealtimeStore>> store_result =
+        factory_->Create(std::move(write_schema), options, memory_pool);
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<RealtimeStore> store, std::move(store_result));
```
zjw1111, line 123: direct `PAIMON_ASSIGN_OR_RAISE`?

SteNicholas (approved):
LGTM.

PR #203, opened 2026-08-14
fix(build): repair singleton double-checked locking race and harden aarch64 portability

u70b3:
### Purpose

Following up on the aarch64 port (#181), an ARM portability audit (7-category static sweep plus on-hardware validation on a 128-core Kunpeng-920, ARMv8.2) found **one real concurrency bug currently hidden by x86 TSO** and three latent UB/divergence risks:

1. **`Singleton<T>` double-checked locking is broken on weak memory models.** The instance pointer was published with a plain store guarded only by a compiler-only `MEMORY_BARRIER`, while the fast path used a plain non-atomic load. A reader can therefore observe a non-null pointer before construction is visible. The fix uses `std::atomic<T*>` with an acquire fast-path load and release publication, while preserving `LazyInstantiation::Create(T*&)`'s public signature.

2. **`IOHook::Impl::mode_` has a data race.** `Reset()` writes the plain enum while IO threads read it in `Try()`. The fix makes it `std::atomic<Mode>` and stores it before the existing sequentially consistent `pos_` and `io_count_` stores.

3. **`SerializationUtils::DeserializeBinaryRow` violates strict aliasing.** It read an arity from a byte-filled buffer through `reinterpret_cast<int32_t*>`. The fix uses `memcpy`, matching the serialize side without changing the wire format.

4. **Extreme cache/block sizes can trigger undefined `double -> int` conversions.** `CacheManager` and `SstFileWriter` could produce architecture-dependent results for pathological configurations. A common-layer `SaturatingDoubleToInteger<T>` helper now implements Java-style saturation and is used at both sites.

`FieldSumAgg` INT8 sum/negation was also audited and intentionally left unchanged: modulo-256 addition and negation produce the same stored bits regardless of plain-`char` signedness. Unlike the min/max comparisons fixed in #181, signedness cannot change these results.

Out of scope (documented follow-ups): tightening option validation for `btree-index.block-size` / `cache-page-size`, and the unchecked int64-to-int32 narrowing in `CoreOptions::GetCachePageSize`.

### Tests

The concurrency tests were verified red before the fix and TSan-clean after it on the Kunpeng-920:

| Test | Coverage |
|---|---|
| `SingletonTest.TestConcurrentFirstPublication` | Contended first publication of `Singleton<T>` on a translation-unit-local type, so no link order, `--gtest_shuffle`, or `--gtest_filter` can pre-publish the instance and silently degrade the gate into the fast path; each thread also verifies full construction visibility |
| `IOHookTest.TestConcurrentResetAndTry` | Concurrent `Reset()` / `Clear()` and `Try()` access; a shared start barrier releases all threads together and the reset loop runs until every worker finishes, so overlap is structural rather than timing-dependent |
| `SerializationUtilsTest.TestDeserializeBinaryRowFromStream` | Stream round-trip and big-endian arity prefix |
| `CacheManagerTest` | Saturation, normal split, and eviction behavior |
| `SaturatingCastTest` | int64/int32 normal, boundary, infinity, and NaN cases |

The `FactoryCreator` first-publication storm from the initial revision was removed because static `REGISTER_PAIMON_FACTORY` constructors initialize it before `main()`, so it cannot test first publication reliably. For the same reason the storm no longer borrows `Singleton<IOHook>`: `GetInstance()` is now defined in the header, so the test instantiates `Singleton<T>` for its own TU-local type and controls first publication by construction. `FactoryCreator` and `IOHook` keep extern-template declarations with explicit instantiations in `singleton.cpp`, because the `-Bsymbolic` plugin shared libraries must share exactly one definition (implicit per-library copies split the singleton and factory registrations are lost).

Validation on the latest head (Kunpeng-920, aarch64):

- `paimon-common-factories-test`: 9/9 passed (Debug), including 20× `--gtest_shuffle` runs and a `--gtest_repeat=5 --gtest_shuffle` run proving order independence
- `paimon-common-test`: 1436/1436 passed and `paimon-common-sst-file-format-test`: 32/32 passed (Debug)
- ThreadSanitizer: factories binary clean; the two race tests stress-run 20× under `--gtest_shuffle` with zero reports
- `pre-commit run --files <changed>`: all checks passed
- `git diff --check`: clean

### API and Format

No public API or storage-format change. `LazyInstantiation::Create(T*&)` remains unchanged, and the big-endian arity prefix is preserved and pinned by tests.

### Documentation

No documentation changes needed; code comments explain the ordering and saturation rationale.

### Generative AI tooling

Generated-by: Claude Code (claude-opus-4-8)

### DIFF
--- a/src/paimon/common/factories/io_hook.cpp
+++ b/src/paimon/common/factories/io_hook.cpp
@@ -32,7 +32,7 @@ class IOHook::Impl {
         if (io_count_.fetch_add(1) < pos_.load()) {
             return Status::OK();
         } else {
-            switch (mode_) {
+            switch (mode_.load(std::memory_order_relaxed)) {
                 case IOHook::Mode::SILENT:
                     return Status::OK();
                 case IOHook::Mode::RETURN_ERROR:
@@ -49,9 +49,11 @@ class IOHook::Impl {
     }
 
     inline void Reset(int64_t pos, IOHook::Mode mode) {
+        // Store mode_ first: the seq_cst stores below then publish it, so a Try()
+        // that observes the reset pos_ also observes the new mode_.
+        mode_.store(mode, std::memory_order_relaxed);
         pos_ = pos;
         io_count_ = 0;
-        mode_ = mode;
     }
 
     int64_t IOCount() const {
@@ -65,7 +67,7 @@ class IOHook::Impl {
  private:
     std::atomic<int64_t> io_count_ = {0};
     std::atomic<int64_t> pos_ = {-1};
-    IOHook::Mode mode_ = IOHook::Mode::SILENT;
+    std::atomic<IOHook::Mode> mode_{IOHook::Mode::SILENT};
 };
 
 IOHook::IOHook() : impl_(std::make_unique<IOHook::Impl>()) {}
--- a/src/paimon/common/factories/io_hook.cpp
+++ b/src/paimon/common/factories/io_hook.cpp
@@ -24,6 +24,7 @@
 #include <stdexcept>
 
 #include "fmt/format.h"
+#include "paimon/macros.h"
 #include "paimon/status.h"
 
 namespace paimon {
@@ -31,6 +32,42 @@ namespace paimon {
 class IOHook::Impl {
  public:
     Status Try(const std::string& path) {
+        // Fast path: the hook is disabled, which is always the case in production;
+        // writers (Reset()/Clear()) only exist in tests. This keeps Try() a single
+        // atomic load on the IO path instead of a shared_mutex acquisition per IO.
+        if (PAIMON_UNLIKELY(armed_.load(std::memory_order_acquire))) {
+            return TryArmed(path);
+        }
+        return Status::OK();
+    }
+
+    inline void Reset(int64_t pos, IOHook::Mode mode) {
+        std::unique_lock<std::shared_mutex> lock(mutex_);
+        mode_ = mode;
+        pos_ = pos;
+        io_count_ = 0;
+        // Arm only after the configuration is complete: TryArmed() reads mode_/pos_
+        // under mutex_, which synchronizes with this store, so an observed armed state
+        // always implies a complete configuration.
+        armed_.store(true, std::memory_order_release);
+    }
+
+    int64_t IOCount() const {
+        std::shared_lock<std::shared_mutex> lock(mutex_);
+        return io_count_.load();
+    }
+
+    void Clear() {
+        std::unique_lock<std::shared_mutex> lock(mutex_);
+        // Disarm first so IO threads stop taking the lock as soon as possible.
+        armed_.store(false, std::memory_order_release);
+        mode_ = IOHook::Mode::SILENT;
+        pos_ = -1;
+        io_count_ = 0;
+    }
+
+ private:
+    Status TryArmed(const std::string& path) {
         std::shared_lock<std::shared_mutex> lock(mutex_);
         if (io_count_.fetch_add(1) < pos_) {
             return Status::OK();
@@ -51,24 +88,8 @@ class IOHook::Impl {
         }
     }
 
-    inline void Reset(int64_t pos, IOHook::Mode mode) {
-        std::unique_lock<std::shared_mutex> lock(mutex_);
-        mode_ = mode;
-        pos_ = pos;
-        io_count_ = 0;
-    }
-
-    int64_t IOCount() const {
-        std::shared_lock<std::shared_mutex> lock(mutex_);
-        return io_count_.load();
-    }
-
-    void Clear() {
-        Reset(-1, IOHook::Mode::SILENT);
-    }
-
- private:
     mutable std::shared_mutex mutex_;
+    std::atomic<bool> armed_ = {false};
     std::atomic<int64_t> io_count_ = {0};
     int64_t pos_ = -1;
     IOHook::Mode mode_ = IOHook::Mode::SILENT;
--- a/src/paimon/common/factories/io_hook.cpp
+++ b/src/paimon/common/factories/io_hook.cpp
@@ -19,6 +19,8 @@
 #include "paimon/common/factories/io_hook.h"
 
 #include <atomic>
+#include <mutex>
+#include <shared_mutex>
 #include <stdexcept>
 
 #include "fmt/format.h"
@@ -29,18 +31,19 @@ namespace paimon {
 class IOHook::Impl {
  public:
     Status Try(const std::string& path) {
-        if (io_count_.fetch_add(1) < pos_.load()) {
+        std::shared_lock<std::shared_mutex> lock(mutex_);
+        if (io_count_.fetch_add(1) < pos_) {
             return Status::OK();
         } else {
-            switch (mode_.load(std::memory_order_relaxed)) {
+            switch (mode_) {
                 case IOHook::Mode::SILENT:
                     return Status::OK();
                 case IOHook::Mode::RETURN_ERROR:
                     return Status::IOError(fmt::format(
-                        "io hook triggered io error at position {}, path {}", pos_.load(), path));
+                        "io hook triggered io error at position {}, path {}", pos_, path));
                 case IOHook::Mode::THROW_EXCEPTION:
                     throw std::runtime_error(fmt::format(
-                        "io hook throw io exception at position {}, path {}", pos_.load(), path));
+                        "io hook throw io exception at position {}, path {}", pos_, path));
                     return Status::OK();
                 default:
                     return Status::OK();
@@ -49,14 +52,14 @@ class IOHook::Impl {
     }
 
     inline void Reset(int64_t pos, IOHook::Mode mode) {
-        // Store mode_ first: the seq_cst stores below then publish it, so a Try()
-        // that observes the reset pos_ also observes the new mode_.
-        mode_.store(mode, std::memory_order_relaxed);
+        std::unique_lock<std::shared_mutex> lock(mutex_);
+        mode_ = mode;
         pos_ = pos;
         io_count_ = 0;
     }
 
     int64_t IOCount() const {
+        std::shared_lock<std::shared_mutex> lock(mutex_);
         return io_count_.load();
     }
 
@@ -65,9 +68,10 @@ class IOHook::Impl {
     }
 
  private:
+    mutable std::shared_mutex mutex_;
     std::atomic<int64_t> io_count_ = {0};
-    std::atomic<int64_t> pos_ = {-1};
-    std::atomic<IOHook::Mode> mode_{IOHook::Mode::SILENT};
+    int64_t pos_ = -1;
+    IOHook::Mode mode_ = IOHook::Mode::SILENT;
 };
 
 IOHook::IOHook() : impl_(std::make_unique<IOHook::Impl>()) {}
--- a/src/paimon/common/factories/io_hook_test.cpp
+++ b/src/paimon/common/factories/io_hook_test.cpp
@@ -68,21 +68,27 @@ TEST(IOHookTest, TestThrowExceptionMode) {
     hook->Clear();
 }
 
-// Regression test for the data race on IOHook's mode: Reset()/Clear() run on one
-// thread while other threads call Try() concurrently. Under a ThreadSanitizer build
-// this deterministically reports the unsynchronized mode access; functionally it must
-// never crash and every Try() must return OK.
+// Regression test for torn IOHook configurations: Reset()/Clear() run on one thread
+// while other threads call Try() concurrently. A shared start barrier releases all
+// threads together, and the reset thread keeps hammering until every worker has
+// finished, so overlap is structural rather than timing-dependent. Under a
+// ThreadSanitizer build this deterministically reports any unsynchronized access;
+// functionally every Try() must return OK.
 TEST(IOHookTest, TestConcurrentResetAndTry) {
     auto hook = IOHook::GetInstance();
 
-    constexpr int32_t kResetIterations = 200000;
     constexpr int32_t kTryIterations = 50000;
     constexpr int32_t kNumWorkers = 4;
 
+    std::atomic<bool> start{false};
+    std::atomic<int32_t> workers_done{0};
     std::atomic<bool> observed_error{false};
 
-    std::thread reset_thread([hook]() {
-        for (int32_t i = 0; i < kResetIterations; i++) {
+    std::thread reset_thread([hook, &start, &workers_done]() {
+        while (!start.load(std::memory_order_acquire)) {
+            std::this_thread::yield();
+        }
+        while (workers_done.load(std::memory_order_relaxed) < kNumWorkers) {
             hook->Reset(INT64_MAX, IOHook::Mode::RETURN_ERROR);
             hook->Clear();
         }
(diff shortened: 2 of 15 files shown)


src/paimon/core/mergetree/compact/aggregate/field_sum_agg.cpp
```diff
@@ -32,12 +32,16 @@ Result<FieldSumAgg::FieldSumFunc> FieldSumAgg::CreateSumFunc(
...
-                               DataDefine::GetVariantValue<char>(input_field);
-                    return sum;
-                });
+            // The variant holds TINYINT as a plain char, whose signedness is fixed by the
+            // target ABI. The mod-256 wrap of the stored byte is identical either way, but add
+            // the signed values it stands for so the accumulator keeps Java's byte semantics if
+            // it is ever widened or compared.
+            return FieldSumFunc([](const VariantType& accumulator,
+                                   const VariantType& input_field) -> VariantType {
+                int8_t sum = static_cast<int8_t>(DataDefine::GetVariantValue<char>(accumulator) +
```
SteNicholas: The current `clang-debug-x86_64` job is red because `check-clang-tidy` reports `modernize-use-auto` on this declaration. Please use `const auto sum = static_cast<int8_t>(...)` (or otherwise satisfy the check) so the required Clang job passes: https://github.com/apache/paimon-cpp/actions/runs/31763982737/job/94658447782

src/paimon/core/mergetree/compact/aggregate/field_sum_agg.cpp
```diff
@@ -93,9 +97,10 @@ Result<FieldSumAgg::FieldNegFunc> FieldSumAgg::CreateNegFunc(
     arrow::Type::type type = field_type->id();
     switch (type) {
         case arrow::Type::type::INT8:
+            // As in CreateSumFunc, negate the signed value the plain char stands for.
             return FieldNegFunc([](const VariantType& input_field) -> VariantType {
-                char value = DataDefine::GetVariantValue<char>(input_field);
-                return static_cast<char>(-value);
+                auto value = static_cast<int8_t>(DataDefine::GetVariantValue<char>(input_field));
```
SteNicholas: This still relies on an implementation-defined C++17 narrowing conversion. On an unsigned-`char` ABI, bytes `0x80`–`0xff` arrive here as 128–255, which are outside `int8_t`; the sum path has the same issue when narrowing values up to 510. The new test also casts back to `int8_t`, so it repeats rather than detects this assumption. Please decode and wrap through `uint8_t` plus explicit in-range arithmetic (for example, map bits >= 128 to `int16_t(bits) - 256`) instead of an out-of-range signed cast.

u70b3:
Thanks for the lightning-fast review! I’m working through the new comments and the CI failure now, and iterating on the fixes and validation. I opened the PR early to get review started before running the local x86 suite; I’ll push the fixes and updated validation results shortly.

src/paimon/common/factories/io_hook.cpp
```diff
@@ -49,9 +49,11 @@ class IOHook::Impl {
     }
 
     inline void Reset(int64_t pos, IOHook::Mode mode) {
+        // Store mode_ first: the seq_cst stores below then publish it, so a Try()
+        // that observes the reset pos_ also observes the new mode_.
+        mode_.store(mode, std::memory_order_relaxed);
```
SteNicholas: The individual accesses are now data-race-free, but Reset() still does not publish a coherent configuration. For example, starting from (SILENT, -1, 0), Reset(INT64_MAX, RETURN_ERROR) can store the new mode, after which a concurrent Try() increments the old counter, reads the old -1 position, and returns IOError. Both stable configurations would return OK (the completed reset should not fire until INT64_MAX), so this is a torn state rather than either valid outcome. Since the new test explicitly supports concurrent Reset()/Try(), please publish an immutable combined state including a fresh counter, or synchronize the complete operations; a regression test can alternate Clear() with Reset(INT64_MAX, RETURN_ERROR) and assert that Try() never fails.

src/paimon/common/utils/saturating_cast.h
```diff
@@ -0,0 +1,52 @@
...
+
+/// Converts a double to a signed integral type with the saturation policy docs/code-style.md
+/// requires (the one Java applies and NumericPrimitiveCastExecutor::JavaFloatingToIntegerCast
+/// implements in core/): NaN converts to 0 and an out-of-range value saturates at the bounds
+/// of TargetType. A bare static_cast of an unrepresentable double is undefined behavior and
+/// diverges across architectures (x86 cvttsd2si yields the "integer indefinite" value, aarch64
+/// fcvtzs saturates), so doubles that are not provably in range must go through this helper.
+template <typename TargetType>
+inline TargetType SaturatingDoubleToInteger(double value) {
+    static_assert(std::is_integral_v<TargetType> && std::is_signed_v<TargetType>,
```
SteNicholas: This accepts every signed integral type, but it does not implement the Java/Paimon policy claimed above for narrow targets. For example, SaturatingDoubleToInteger<int8_t>(300.9) returns 127, while Java first converts to int32_t and then narrows the low bits, producing 44; a huge value similarly yields 127 here instead of -1. Please either implement the existing wide-then-narrow behavior for int8_t/int16_t, or restrict TargetType to the currently intended int32_t and int64_t instantiations and document that narrower contract.

src/paimon/common/factories/io_hook.cpp
u70b3: Good catch. I now synchronize the complete IOHook configuration with a shared mutex and updated the regression test to assert that concurrent Clear() / Reset(INT64_MAX, RETURN_ERROR) never returns an error.

src/paimon/common/utils/saturating_cast.h
u70b3: Agreed. I restricted the helper to the intended int32_t and int64_t targets and clarified that narrower Java conversions require a separate narrowing step.

u70b3:
ARM re-test on the latest head (cba5129) — all green ✅

**Environment**: Kunpeng-920 (aarch64, 128 cores), Ubuntu 22.04, GCC 11.4, Clang 14, CMake 3.22

**Debug (GCC 11.4)**
- `paimon-common-factories-test` (Singleton/IOHook): 9/9 passed
- `paimon-common-test` (SaturatingCast/SerializationUtils/CacheManager): 10/10 passed
- `paimon-common-sst-file-format-test` (SstFileWriter): 32/32 passed
- Full `unittest` suite (40 ctest entries): all passed

**ThreadSanitizer (Clang 14)**
- factories + common test binaries: passed, no TSan warnings
- Race tests stress-run 20× (`SingletonTest.*` + `IOHookTest.TestConcurrentResetAndTry`): 20/20 clean
- Same tests repeated 20× under GCC Debug: all passed

**Checks**: `pre-commit` (clang-format / cmake-format / codespell / C++ Lint) and `git diff --check` all clean.

The `shared_mutex`-based IOHook and the int32/int64-only `SaturatingDoubleToInteger` in cba5129 behave correctly on aarch64 — LGTM from the ARM side.

src/paimon/common/factories/singleton_test.cpp
```diff
@@ -0,0 +1,88 @@
...
+}  // namespace
+
+// Regression gate for the Singleton double-checked-locking publication race.
+// It only exercises the first construction if nothing has touched
+// Singleton<IOHook> before, so this must stay the first GetInstance() call in
+// this binary (singleton_test.cpp is the first source of common_factories_test
+// and this is its first test). A FactoryCreator storm cannot serve as the
+// gate: the REGISTER_PAIMON_FACTORY constructors in paimon_shared already
+// initialize Singleton<FactoryCreator> before main().
+TEST(SingletonTest, TestConcurrentIOHookGetInstance) {
```
SteNicholas: This regression gate depends on this test being the first code to instantiate `Singleton<IOHook>`, but CMake source order does not guarantee cross-translation-unit test registration/execution order across linkers, and `--gtest_shuffle` can reorder it. If another test calls `GetInstance()` first, this storm only exercises the already-published fast path and cannot detect the original publication race. Please make the first construction test-controlled, for example with a dedicated test singleton or an isolated test executable.

src/paimon/common/factories/io_hook_test.cpp
```diff
@@ -64,4 +68,49 @@ TEST(IOHookTest, TestThrowExceptionMode) {
...
+TEST(IOHookTest, TestConcurrentResetAndTry) {
+    auto hook = IOHook::GetInstance();
+
+    constexpr int32_t kResetIterations = 200000;
+    constexpr int32_t kTryIterations = 50000;
+    constexpr int32_t kNumWorkers = 4;
+
+    std::atomic<bool> observed_error{false};
+
+    std::thread reset_thread([hook]() {
```
SteNicholas: The reset thread starts before the worker threads are even created, so it can finish all iterations before any `Try()` runs. Because every iteration ends with `Clear()`, the test then passes in the final `SILENT` state without exercising concurrent `Reset()`/`Try()`. Please add a shared start barrier and keep the reset loop active until the workers signal completion so overlap is guaranteed.

SteNicholas (commented):
@u70b3, thanks for update. Two test-coverage issues can let the new concurrency regressions pass without exercising the races they are intended to guard.

src/paimon/common/factories/singleton_test.cpp
u70b3: Good catch — and agreed on the deeper hazard: a regression gate that can pass without exercising the race it guards is worse than no test at all. It does not just miss the bug, it silently certifies the fix (our "TSan-clean" claim rested on this storm), so a vacuous green actively masks the failure it was built to catch.

Fixed by making the first construction test-controlled instead of convention-controlled:

- `Singleton<T, InstPolicy>::GetInstance()` moved from singleton.cpp into the header, so a test can instantiate `Singleton<T>` for its own type — previously impossible, which is why the storm had to borrow `Singleton<IOHook>` and inherit the ordering fragility.
- The storm (renamed `TestConcurrentFirstPublication`) now runs on `FirstPublicationTarget`, a class local to singleton_test.cpp's anonymous namespace. Nothing else in the binary can name the type, so no link order, `--gtest_shuffle`, `--gtest_filter`, or `--gtest_repeat` can pre-publish it: the test is guaranteed to race the first publication by construction. Each thread also verifies full construction visibility (a 64-word payload written by the ctor), which is the essence of the original race.
- One subtlety worth recording: the first attempt dropped the explicit instantiations entirely, and local validation immediately caught it — the file-format plugins are separate `-Bsymbolic` shared libraries, so implicit per-library copies of the function-local static state split the singleton (registrations landed in a different instance than lookups, and the wider suites lost the 'orc' factory). `FactoryCreator` and `IOHook` therefore keep extern-template declarations plus explicit instantiations in singleton.cpp; only TU-local test types instantiate implicitly.

Validation on Kunpeng-920 (aarch64), latest head (1bc6650): factories suite 9/9 under 20× `--gtest_shuffle` plus a `--gtest_repeat=5 --gtest_shuffle` run; `paimon-common-test` 1436/1436 and `paimon-common-sst-file-format-test` 32/32 (these are the suites that caught the split-singleton issue); TSan clean over 20 shuffled stress runs of the race tests.

src/paimon/common/factories/io_hook_test.cpp
u70b3: Agreed — same class of hazard: with the reset thread free-running 200k fixed iterations before the workers were even spawned, a loaded CI runner or a TSan-slowed build could serialize the two sides completely, and since every iteration ends in `Clear()` the workers would then only ever see SILENT — a silent no-op masquerading as a concurrency gate. A test that can pass with zero overlap does not just miss the torn-state bug, it masks it behind false confidence.

Fixed structurally:

- All five threads now block on a shared start barrier and are released together.
- The reset loop has no fixed iteration count: it hammers `Reset(INT64_MAX, RETURN_ERROR)` / `Clear()` until every worker has completed all its `Try()` iterations (a `workers_done` counter), so overlap covers the workers' entire lifetime by construction.
- Side benefit: the fixed 200k-iteration cost is gone, so the test is faster under TSan while being strictly stronger (~120ms in Debug).

Validation on Kunpeng-920 (aarch64), latest head (1bc6650): 20× shuffled stress runs of this test plus the singleton storm under TSan — zero reports; Debug suite green under 20× `--gtest_shuffle` and `--gtest_repeat=5 --gtest_shuffle`.

src/paimon/common/factories/io_hook.cpp
```diff
@@ -29,18 +31,19 @@ namespace paimon {
 class IOHook::Impl {
  public:
     Status Try(const std::string& path) {
-        if (io_count_.fetch_add(1) < pos_.load()) {
+        std::shared_lock<std::shared_mutex> lock(mutex_);
```
lucasfang, line 71: The race fix itself is welcome — Reset()/Clear() and Try() were genuinely racy before. My concern is where the lock landed: Try() is invoked by CHECK_HOOK on every local-file IO (in local_file.cpp, hook_ is always initialized to IOHook::GetInstance(), so the guard never skips it). The old implementation was lock-free (fetch_add + an atomic pos_ load); the new one takes a std::shared_mutex shared lock on every IO.
Suggested fix: gate the lock behind an "armed" flag. The hook is never armed in production — the default state is exactly what Clear() sets up (pos_ = -1, SILENT), and writers only exist in tests. So we can keep the mutex for correctness on the armed path while restoring a lock-free fast path for the disabled state:

u70b3: Good point — done in 993b2d5 by gating the synchronized path behind an atomic `armed` flag:

- `Try()` now starts with a lock-free fast path: a single acquire load, returning OK immediately in the disabled state (the production default). The `shared_mutex` is only taken once armed.
- `Reset()` publishes the complete configuration under the mutex and then release-stores `armed`; armed readers acquire the mutex before touching `mode_`/`pos_`, so an observed armed state always implies a complete configuration.
- `Clear()` disarms first (still under the mutex) so IO threads drop off the lock as soon as possible.
- One semantic change worth noting: the disabled fast path no longer increments the counter, so `IOCount()` now counts only while armed. All `IOCount()` consumers are tests, and the header documents the contract.
- Coverage: new `IOHookTest.TestDisabledFastPath` pins the disabled behavior (Try always OK, no counting, re-arm/disarm restores it). The existing `TestConcurrentResetAndTry` hammers `Reset(INT64_MAX, RETURN_ERROR)`/`Clear()` continuously, so workers keep switching between the fast and armed paths under TSan.

Validation on Kunpeng-920 (aarch64), head 993b2d5: factories suite 10/10 (plus 20× `--gtest_shuffle` and a `--gtest_repeat=5 --gtest_shuffle` run); `paimon-common-test` 1450/1450; `paimon-common-sst-file-format-test` 32/32; TSan clean over 20 shuffled stress runs of the race tests; pre-commit and `git diff --check` clean.

src/paimon/core/mergetree/compact/aggregate/field_sum_agg.cpp
u70b3: We ended up reverting the INT8 change entirely rather than fixing the cast: modulo-256 addition and negation produce the same stored bits regardless of plain-`char` signedness, so the original `char` arithmetic is already ABI-stable here — unlike the min/max comparisons fixed in #181, where signedness does change results. The PR description documents this audit conclusion.

This also moots the clang-tidy `modernize-use-auto` failure from the other thread, since that declaration no longer exists. PTAL.

lucasfang:
+1

u70b3:
@SteNicholas @lucasfang Both review rounds are addressed in the pushed head (890fad4, rebased onto latest main). The `armed`-flag change also arms the hook in #209's `TestPreBufferWindowLimit`, which relied on the old count-while-disabled behavior.

Re-validated locally on aarch64 (gcc Debug): common-test 1459/1459, factories + sst + parquet suites all green, pre-commit clean. The earlier gcc-debug-aarch64 segfault did not reproduce and looks unrelated to this PR.

Could you approve the pending workflow runs?

lxy-9602:
Thank you very much for the PR and for the multiple rounds of updates. Once the current CI passes, we’ll merge it. Thanks again for your patience!

SteNicholas (approved):
LGTM.

PR #205, opened 2026-08-14
feat(fs): add OSS file system

mrdrivingduck:
### Purpose

This PR adds a native, read-only OSS filesystem backed by [Alibaba Cloud OSS SDK V2](https://github.com/aliyun/alibabacloud-oss-cpp-sdk-v2).

### Static linking and distribution

Some integration scenarios require Paimon-cpp to be linked statically into another library, such as a DuckDB extension. JindoSDK is delivered as a separate dynamic library, which makes this integration and distribution model less flexible.

### Shutdown reliability in PostgreSQL processes

While integrating PostgreSQL with DuckDB, we observed an intermittent shutdown hang along the PostgreSQL → DuckDB → duckdb-paimon → paimon-cpp path when JindoSDK was loaded through the Paimon filesystem. The captured shutdown stack was:

```text
PostgreSQL backend process

proc_exit()
  → exit()
    → C++ atexit/static destructor
      → bvar::PassiveStatus::~PassiveStatus()
        → bvar::detail::Sampler::destroy()
          → pthread_mutex_lock()   ← waits indefinitely
```

The process also had a `bvar_sampler` thread running. This points to a fork/exit lifecycle issue in JindoSDK's embedded bRPC/bvar dependencies. Since JindoSDK is closed source and brings a relatively large dependency set, it is difficult to investigate or address from Paimon-cpp.

### An open-source OSS filesystem

Given that OSS SDK V2 is open source, adding an OSS implementation to Paimon-cpp would provide an open and self-contained option for OSS access. The existing object-store abstraction introduced for S3 keeps this addition small and avoids duplicated filesystem logic.

## Benchmark

The benchmark directly reads a 133 MiB Parquet data file belonging to a Paimon table in a real OSS environment. Both filesystems were built in a clean Release build, and each workload was run six times. Values are medians.

| Workload | OSS Median | Jindo Median |
|---|---:|---:|
| Open(path) | 10.63 ms (-8.3%) | 11.59 ms |
| Open(known status) | 0.73 ms (-94.9%) | 14.25 ms |
| 1 MiB range read | 47.37 ms (-27.3%) | 65.13 ms |
| Single 8 MiB sequential read | 96.83 ms (-20.4%) | 121.70 ms |

`Open(known status)` reopens a file with previously obtained metadata. It avoids an extra metadata request through the status propagation introduced in [#189](https://github.com/apache/paimon-cpp/pull/189). The range and sequential workloads each issue a single read request.


mrdrivingduck:
I identified an OSS SDK v2 issue where a retried request can retain a stale `Authorization` header after its timestamp is regenerated, causing `SignatureDoesNotMatch`.

This PR temporarily applies a small build-time patch to overwrite the header on retry. The issue has been reported to the OSS SDK maintainers, and they will address it upstream soon.

src/paimon/fs/oss/oss_file_system_factory.h
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright 2026-present Alibaba Inc.
```
zjw1111: please use ASF license

CMakeLists.txt
```diff
@@ -85,6 +88,9 @@ endif()
 if(PAIMON_ENABLE_JINDO)
     add_definitions(-DPAIMON_ENABLE_JINDO)
 endif()
+if(PAIMON_ENABLE_OSS)
```
zjw1111, line 99: could you add oss and s3 options in the docs (`docs/source/building.rst` and `docs/source/build_system.rst`)

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1359,6 +1373,106 @@ macro(build_jindosdk_nextarch)
...
+                                      "perl -0pi -e 's/\"-Wall\"\\s+\"-Werror\"\\s+\"-pedantic\"\\s+\"-Wextra\"/\"-Wall\" \"-pedantic\" \"-Wextra\"/g' <SOURCE_DIR>/sdk/CMakeLists.txt"
+                        COMMAND bash -c
+                                "perl -0pi -e 's/#if LIBCURL_VERSION_NUM >= 0x074400/#if 0/g' <SOURCE_DIR>/sdk/src/transport/curl/CurlMultiTransport.cpp"
+                        COMMAND bash -c
+                                "perl -0pi -e 's/#if LIBCURL_VERSION_NUM >= 0x074200/#if 0/g' <SOURCE_DIR>/sdk/src/transport/curl/CurlMultiTransport.cpp"
+                        COMMAND bash -c
+                                "perl -0pi -e 's/request->headers\\.emplace\\(\"Authorization\", credentialHeader\\)\\x3b/request->headers.insert_or_assign(\"Authorization\", credentialHeader)\\x3b/g' <SOURCE_DIR>/sdk/src/signer/SignerV1.cpp"
+                        COMMAND bash -c
+                                "perl -0pi -e 's/request->headers\\.emplace\\(\"Authorization\", ss\\.str\\(\\)\\)\\x3b/request->headers.insert_or_assign(\"Authorization\", ss.str())\\x3b/g' <SOURCE_DIR>/sdk/src/signer/SignerV4.cpp"
+                                ${THIRDPARTY_LOG_OPTIONS}
```
zjw1111: could you make the patch as a diff file (like `cmake_modules/jieba.diff`), and add the diff file in `LICENSE` file like ↓

Image

mrdrivingduck: It will be removed once the fix https://github.com/aliyun/alibabacloud-oss-cpp-sdk-v2/commit/fe7d847b0df25f561912e0e24432ce13cede6b1a is merged. I plan not to involve the patch in this commit.

mrdrivingduck:
> ## Benchmark
> The benchmark directly reads a 133 MiB Parquet data file belonging to a Paimon table in a real OSS environment. Both filesystems were built in a clean Release build, and each workload was run six times. Values are medians.
>
> `Open(known status)` reopens a file with previously obtained metadata. It avoids an extra metadata request through the status propagation introduced in [#189](https://github.com/apache/paimon-cpp/pull/189). The range and sequential workloads each issue a single read request.

Oh, here is a Paimon semantic-level benchmark supplement to the pure filesystem benchmark above. FYI.

- `regular_wide_rg_64k`: regular append-only wide table with 64K-row Parquet row groups; used for first-batch, full, projected, and no-prebuffer scans.
- `regular_wide_rg_1k`: the same style of table with 1K-row groups; used for the selective predicate scan.
- `regular_wide_small_files`: regular table split into small data files.
- `regular_partitioned`: partitioned regular table; used for partition and prefetch scans.

| Scenario | OSS FS | Jindo FS | OSS/Jindo |
|---|---:|---:|---:|
| First batch | 121 ms | 114 ms | +6.1% |
| Full scan | 232.7 ms | 257.8 ms | -9.7% |
| Projected scan | 123.3 ms | 113.8 ms | +8.4% |
| Predicate scan | 25 ms | 24 ms | +4.2% |
| Small-files scan | 255.7 ms | 239.7 ms | +6.7% |
| Partition scan | 65.3 ms | 64.4 ms | +1.4% |
| Prefetch scan | 765 ms | 855 ms | -10.5% |
| Scan without Parquet prebuffer | 489 ms | 485 ms | +0.8% |

Overall, OSS FS and Jindo FS show broadly comparable performance under Paimon scan semantics. The observed differences are generally within a single-digit range and do not indicate a consistent performance disadvantage, but much more like noise.

src/paimon/fs/oss/oss_file_system_factory.cpp
```diff
@@ -0,0 +1,156 @@
...
+    std::string security_token = GetOption(options, bucket, "securityToken");
+    if (security_token.empty()) {
+        security_token = GetOption(options, bucket, "sessionToken");
+    }
+
+    oss2::ClientConfiguration config = oss2::ClientConfiguration::loadDefault();
+    if (!endpoint.empty()) {
+        config.endpoint = NormalizeEndpoint(endpoint);
+    }
+    config.region = region;
```
zjw1111: Thanks for wiring up the region handling here. I think there is a gap worth closing before this lands.

`InferRegion` only recognizes `oss-<region>[-internal].aliyuncs.com` and returns `""` for anything else (custom CNAME domains, `oss-accelerate.aliyuncs.com`, transfer-acceleration and dual-stack endpoints). Since the guard above only rejects the case where *both* endpoint and region are empty, configuring just `fs.oss.endpoint` with one of those forms passes validation and leaves `config.region` as an empty string.

The SDK defaults to v4 signing (`ClientImplBase::resolveSigner`: `if ("v1" == config.signatureVersion.value_or("v4"))`) and passes the region through unvalidated (`options_.region = config.region.value_or("")`), so `SignerV4` ends up building a credential scope of `<date>//oss/aliyun_v4_request`. Every request then fails server-side with `SignatureDoesNotMatch`, and because `Create()` succeeds the user only sees it at the first I/O rather than at configuration time.

Could you make the missing-region case explicit? Either returning `Status::Invalid` asking for `fs.oss.region` when inference fails and no explicit `signatureVersion` was given, or falling back to v1 signing (which needs no region, and is closer to what the Hadoop OSS connector does on the Java side) would work. It would also be a little safer to assign `config.region` only when it is non-empty, so the SDK's own `has_value()` checks stay meaningful.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1967,6 +2081,9 @@ if(PAIMON_ENABLE_JINDO)
     build_jindosdk_c()
     build_jindosdk_nextarch()
 endif()
+if(PAIMON_ENABLE_OSS)
+    resolve_dependency(OSS_SDK_V2)
```
zjw1111, line 2146: `resolve_dependency` takes its SYSTEM branch through `find_package(${DEPENDENCY_NAME}Alt REQUIRED MODULE)`, and all 16 existing dependencies routed through this macro ship a matching `cmake_modules/Find<Name>Alt.cmake`. This PR does not add `FindOSS_SDK_V2Alt.cmake`, and `OSS_SDK_V2_SOURCE` is not declared in `DefineOptions.cmake` either (every other dependency has one).

The default `AUTO` path uses a QUIET lookup and falls back to bundled, so CI does not hit this — but `-DPAIMON_DEPENDENCY_SOURCE=SYSTEM`, which is a documented option, fails at configure time with a "Could not find module FindOSS_SDK_V2Alt.cmake" message that gives no hint it is about OSS.

Would you prefer to add the Find module plus the `OSS_SDK_V2_SOURCE` option, or to declare the SDK bundled-only? For the latter, `paimon_set_dependency_source_default(OSS_SDK_V2 BUNDLED "...")` is already used for Arrow and ORC and would shield this from the global setting.

src/paimon/CMakeLists.txt
```diff
@@ -190,11 +190,13 @@ set(PAIMON_COMMON_SRCS
 # The shared HTTP client is used by both the object store file systems and the
 # rest catalog.
 set(PAIMON_CURL_LINK_LIBS)
-if(PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST)
+if(PAIMON_ENABLE_OSS
```
zjw1111: The matching test gate further down (`PAIMON_COMMON_HTTP_CLIENT_TEST_SRCS`, still `if(PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST)`) was not updated, and the comment right above it says "so its test follows the same gate". With `-DPAIMON_ENABLE_OSS=ON` and S3/REST off, `http_client.cpp` is compiled into `paimon_shared` while `http_client_test.cpp` is excluded.

That said, I wonder whether this hunk is needed at all. The OSS plugin uses the SDK's own curl transport, and neither `oss_file_system.cpp` / `oss_file_system_factory.cpp` nor `common/fs/object_store_file_system.cpp` includes `paimon/common/utils/http_client.h`. `src/paimon/fs/oss/CMakeLists.txt` already lists `CURL::libcurl` in its own `DEPENDENCIES`, and `alibabacloud_oss_v2::oss` links it via INTERFACE, so `PAIMON_CURL_LINK_LIBS` is not required here either.

Could you revert this hunk and keep only the `object_store_file_system.cpp` gate below? The top-level `CMakeLists.txt` change that makes CURL discoverable for the SDK build should of course stay. If you would rather keep it, then the test gate needs the same update.

src/paimon/fs/oss/oss_file_system.cpp
```diff
@@ -0,0 +1,188 @@
...
+                fmt::format("OSS GetObject read {} bytes for oss://{}/{}, expected {}", written,
+                            path.bucket, path.key, size));
+        }
+        return written;
+    }
+
+    void GetObjectRangeAsync(const ObjectStorePath& path, int64_t offset, int64_t size,
+                             char* buffer, std::function<void(Status)>&& callback) const override {
+        std::shared_ptr<const OssObjectStoreClient> self = shared_from_this();
+        GetGlobalDefaultExecutor()->Add([self = std::move(self), path, offset, size, buffer,
```
zjw1111: This reuses the process-wide singleton pool. `GetGlobalDefaultExecutor()` (`common/executor/executor.cpp:139`) is a function-local static sized to `hardware_concurrency()`, and it currently has no production callers at all — only `default_executor_test.cpp` — so this would be the first one.

The S3 client instead owns a dedicated `std::unique_ptr<Executor>` built by `CreateDefaultExecutor()` (`s3_file_system.cpp:904` and `:960`). Since `GetObjectRangeAsync` is the only implementation path behind `ObjectStoreFileSystem::ReadAsync`, sharing the global pool means that once callers also submit work to it and block on OSS reads inside those tasks, the fixed-size pool can starve itself. It also makes the thread count non-configurable per file system.

Could you follow the S3 pattern and hold a dedicated `Executor` in `OssObjectStoreClient`, optionally exposing the thread count through an `fs.oss.*` option?

src/paimon/fs/oss/oss_file_system_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+    std::map<std::string, std::string> options = {
+        {kOssAccessKeyIdOption, ""},
+        {kOssAccessKeySecretOption, ""},
+        {kOssEndpointOption, ""},
+        {"fs.oss.bucket.bucket.accessKeyId", "access-key"},
+        {"fs.oss.bucket.bucket.accessKeySecret", "secret-key"},
+        {"fs.oss.bucket.bucket.endpoint", "oss-cn-hangzhou.aliyuncs.com"},
+    };
+    ASSERT_TRUE(factory.Create("oss://bucket/key", options).ok());
+}
```
zjw1111, line 325: Thanks for adding the option-validation tests. The client layer is still uncovered though: `HeadObject`, `ListObjects`, `GetObjectRange` and `GetObjectRangeAsync` have no tests, and neither do `InferRegion` or `ParseTimeMillis`. For comparison, `s3_file_system_test.cpp` is around 500 lines and covers response parsing plus error paths through a `MockHttpClient`.

The equivalent seam exists for OSS: `alibabacloud/oss2/transport/HttpTransport.h` exposes a public abstract `HttpTransport` (with `NopHttpTransport` as a starting point) and `ClientConfiguration::httpTransport` accepts an injected instance, so this can be tested without any network access.

Would it be possible to add at least table-driven cases for `InferRegion` (standard / internal / accelerate / custom domain), a malformed-input case for `ParseTimeMillis`, and success plus error paths for `HeadObject`, `ListObjects` and `GetObjectRange`? The error mapping in `IsNotFoundError`, the range-header construction and the short-read check are hard to verify any other way.

src/paimon/fs/oss/oss_file_system_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "gtest/gtest.h"
+#include "paimon/fs/oss/oss_file_system_factory.h"
+
+namespace paimon::oss {
+
+TEST(OssFileSystemFactoryTest, TestOptionValidation) {
+    OssFileSystemFactory factory;
+    std::map<std::string, std::string> options;
+    ASSERT_FALSE(factory.Create("s3://bucket/key", options).ok());
```
zjw1111: Could you use `ASSERT_NOK` / `ASSERT_OK` from `paimon/testing/utils/testharness.h` instead of `ASSERT_FALSE(...ok())` / `ASSERT_TRUE(...ok())`? They accept `Result<T>` directly and print the actual status on failure, which makes diagnosing a broken test much easier. This applies to the other assertions in this file as well (lines 31, 37, 41, 44, 48 and 61); `s3_file_system_test.cpp` already uses these macros.

src/paimon/fs/oss/oss_file_system_factory.cpp
```diff
@@ -0,0 +1,156 @@
...
+Result<std::string> GetRequiredOption(const std::map<std::string, std::string>& options,
+                                      const std::string& bucket, const std::string& suffix) {
+    std::string value = GetOption(options, bucket, suffix);
+    if (value.empty()) {
+        return Status::Invalid(fmt::format("OSS option 'fs.oss.{}' must not be empty", suffix));
+    }
+    return value;
+}
+
+Result<bool> ParseBool(std::string value, const std::string& option) {
```
zjw1111: There is already a shared helper for this: `StringUtils::StringToValue<bool>` (`common/utils/string_utils.h`), and `OptionsUtils::GetValueFromMap<bool>` wraps it into the `Result<T>` form that the S3 factory uses (`s3_file_system.cpp:646`). Could you reuse one of those instead of a local `ParseBool`?

One thing to be aware of: the accepted spellings differ slightly. The shared helper takes `t/true/y/yes/1` and `f/false/n/no/0`, so `on` and `off` would no longer be accepted. Aligning with the shared helper still seems preferable to keeping a separate dialect just for OSS options.

src/paimon/fs/oss/oss_file_system.cpp
```diff
@@ -0,0 +1,188 @@
...
+        if (result.getContentLength() < 0) {
+            return Status::IOError("OSS HeadObject response is missing Content-Length");
+        }
+        return ObjectMetadata{path.key, result.getContentLength(),
+                              ParseTimeMillis(result.getLastModified())};
+    }
+
+    Result<ListObjectsResult> ListObjects(const ObjectStorePath& path,
+                                          const std::string& continuation_token,
+                                          int max_keys) const override {
```
zjw1111: `docs/code-style.md` asks for fixed-width integer types, and the base declaration is `int32_t max_keys` (`common/fs/object_store_file_system.h:59`); the S3 client uses `int32_t` as well. Could you change this to `int32_t`?

src/paimon/fs/oss/oss_file_system.cpp
```diff
@@ -0,0 +1,188 @@
...
+        PAIMON_RETURN_NOT_OK(ValidateBucket(path));
+        oss2::models::ListObjectsV2Request request;
+        request.setBucket(path.bucket).setPrefix(path.key).setDelimiter("/");
+        if (!continuation_token.empty()) {
+            request.setContinuationToken(continuation_token);
+        }
+        if (max_keys > 0) {
+            request.setMaxKeys(std::min(max_keys, 999));
```
zjw1111: Minor: the OSS `max-keys` limit is 1000, so 999 reads as an unexplained magic number. The base class also only ever calls this with 1 or 0 today (`object_store_file_system.cpp:367` and `:434`), so the clamp never actually kicks in. Could you use 1000 with a short comment, or drop the clamp?

src/paimon/fs/oss/oss_file_system.h
```diff
@@ -0,0 +1,45 @@
...
+
+#include "paimon/common/fs/object_store_file_system.h"
+
+namespace alibabacloud::oss2 {
+class OSSClient;
+}
+
+namespace paimon::oss {
+
+inline constexpr char kOssAccessKeyIdOption[] = "fs.oss.accessKeyId";
```
zjw1111, line 34: These constants and the keys actually being looked up are two separate sources of truth: the factory builds its keys from literal suffixes (`GetOption(options, bucket, "accessKeyId")`), so only `kOssUsePathStyleOption` is referenced from production code, and `kOssSecurityTokenOption` / `kOssSessionTokenOption` are not referenced anywhere. Renaming or adding an option means remembering to update both places.

Could you either have `GetOption` take the full constant and derive the bucket-scoped key from it, or drop the ones that are unused? Relatedly, the error message in `GetRequiredOption` ("OSS option 'fs.oss.{}' must not be empty") always reports the global key, even when the bucket-scoped `fs.oss.bucket.<bucket>.<suffix>` is the one that was being looked up.

src/paimon/fs/oss/oss_file_system.cpp
mrdrivingduck: Yeah, I have witnessed this issue during my local testing. Will use a dedicated pool for OSS.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -1359,6 +1373,106 @@ macro(build_jindosdk_nextarch)
...
+    set(OSS_SDK_V2_INCLUDE_DIR "${OSS_SDK_V2_PREFIX}/include")
+    set(OSS_SDK_V2_INSTALL_LIBDIR "${CMAKE_INSTALL_LIBDIR}")
+    set(OSS_SDK_V2_LIB_DIR "${OSS_SDK_V2_PREFIX}/${OSS_SDK_V2_INSTALL_LIBDIR}")
+    set(OSS_SDK_V2_STATIC_LIB
+        "${OSS_SDK_V2_LIB_DIR}/${CMAKE_STATIC_LIBRARY_PREFIX}alibabacloud-oss-cpp-sdk-v2${CMAKE_STATIC_LIBRARY_SUFFIX}"
+    )
+
+    set(OSS_SDK_V2_CMAKE_CXX_FLAGS "${EP_CXX_FLAGS} -Wno-error")
+    set(OSS_SDK_V2_CMAKE_C_FLAGS "${EP_C_FLAGS} -Wno-error")
+    string(REPLACE "-Werror" "" OSS_SDK_V2_CMAKE_CXX_FLAGS ${OSS_SDK_V2_CMAKE_CXX_FLAGS})
```
zjw1111: maybe can direct use `EP_CXX_FLAGS`/`EP_C_FLAGS` 
Image

src/paimon/fs/oss/oss_file_system_factory.h
mrdrivingduck: Done.

CMakeLists.txt
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_factory.cpp
mrdrivingduck: Done.

cmake_modules/ThirdpartyToolchain.cmake
mrdrivingduck: Done.

src/paimon/CMakeLists.txt
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_test.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_test.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_factory.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.cpp
mrdrivingduck: Done.

cmake_modules/ThirdpartyToolchain.cmake
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.h
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.cpp
```diff
@@ -0,0 +1,196 @@
...
+        return Status::NotExist(message);
+    }
+    if (error.getCode() == "RequestCanceled") {
+        return Status::Cancelled(message);
+    }
+    return Status::IOError(message);
+}
+
+int64_t ParseTimeMillis(const std::string& value) {
+    time_t seconds = curl_getdate(value.c_str(), nullptr);
```
zjw1111: `curl_getdate` cannot parse the timestamp format that `ListObjectsV2` returns, so this helper returns `0` for every listed object.

`curl_getdate` only accepts formats carrying an English month name (RFC 822/1123, RFC 850, `asctime` and their loose variants), plus one compact `YYYYMMDD` form. In curl's `lib/parsedate.c`, `w->mon` is only ever set by `checkmonth()` on an alphabetic token or by the `num_digits == 8` branch in `datenum()`, and `datecheck()` fails immediately when `mon == -1`.

`ListObjectsV2` returns ISO 8601 in the XML body and the SDK stores it verbatim in `ObjectSummary::lastModified`. For `2024-01-01T00:00:00.000Z` the parse actually fails even earlier: the tokens are `2024` (taken as year), `01` (taken as mday), and the third `01` has no field left to fill, so `datenum()` returns `PARSEDATE_FAIL`.

`HeadObject` is unaffected because `HeadObjectResult::getLastModified()` reads the HTTP `Last-Modified` header, which is an HTTP-date and does contain a month name. That asymmetry makes this easy to miss.

Downstream, every `FileStatus` produced by `ListDir` / `ListStatus` gets `modification_time == 0`, and `OrphanFilesCleanerImpl` then rejects it (`GetModificationTime() <= MIN_VALID_FILE_MODIFICATION_MS` returns `Status::Invalid("... is not in millisecond")`), so orphan file cleanup fails on OSS. Returning `0` also collides with `FileStatus::kUnknownModificationTime`, which is `-1`, so callers cannot distinguish "unknown" from the epoch.

`s3_file_system.cpp` has the same defect in `ParseModificationTime` (S3's `ListObjectsV2` `LastModified` is ISO 8601 as well), and the two also differ in the failure check: S3 uses `seconds == static_cast<time_t>(-1)` while this uses `seconds < 0`, which is always false where `time_t` is unsigned.

Would it be possible to extract one shared helper used by both file systems — handling ISO 8601 explicitly (`strptime("%Y-%m-%dT%H:%M:%S")` + `timegm`, plus the optional `.SSS` and the `Z` / `+hh:mm` suffix), falling back to `curl_getdate` for HTTP-date, and returning `kUnknownModificationTime` on failure?

It would also help to assert a parsed value in the tests rather than only the failure path. `TestListObjects` already feeds `<LastModified>2024-01-01T00:00:00.000Z</LastModified>` but only checks `statuses.size()`, and `TestHeadObjectParsesMetadata` uses `not-a-timestamp` and asserts `0`, so this slipped through.

src/paimon/fs/oss/oss_file_system_factory.cpp
```diff
@@ -0,0 +1,201 @@
...
+    if (suffix == std::string::npos || suffix + kSuffix.size() != endpoint.size()) {
+        return "";
+    }
+    std::string region = endpoint.substr(kPrefix.size(), suffix - kPrefix.size());
+    constexpr std::string_view kInternal = "-internal";
+    if (region.size() > kInternal.size() &&
+        region.compare(region.size() - kInternal.size(), kInternal.size(), kInternal) == 0) {
+        region.erase(region.size() - kInternal.size());
+    }
+    if (region == "accelerate" ||
```
zjw1111: Thanks for tightening this after the previous round — failing fast instead of silently guessing is the right direction. I think the deny-list still lets one real endpoint through.

`oss-accelerate-overseas.aliyuncs.com` is a real endpoint (the SDK produces exactly this string for `EndpointType::Overseas` in `regionToEndpoint()`). It is neither equal to `"accelerate"` nor suffixed with `-dualstack`, so `InferRegion` returns `"accelerate-overseas"`, which is not a region. That value is assigned to `config.region`, and since the SDK defaults `signatureVersion` to `v4`, the region feeds the V4 signing scope and the derived signing key — so every request fails with `SignatureDoesNotMatch`, with an error mentioning nothing about endpoint or region configuration. Returning `""` here would instead surface the clear "OSS region must be configured ..." error at creation time.

The `-dualstack` branch also looks unreachable: real dual-stack endpoints are `<region>.oss.aliyuncs.com` (again per `regionToEndpoint()`), which does not start with `oss-` and is already rejected by the prefix check above. `TestEndpointRegionValidation` uses `oss-cn-hangzhou-dualstack.aliyuncs.com`, a form OSS does not serve, so that case passes for the wrong reason.

Would it be possible to invert this into an allow-list instead — accepting the inferred value only when it matches a region shape such as `^[a-z]{2,3}-[a-z]+(-[0-9]+)?$`? Then `accelerate`, `accelerate-overseas` and any future non-region endpoint all fall into the explicit-error path automatically, and the `-dualstack` branch could be dropped. It would be good to also change that test case to a real `cn-hangzhou.oss.aliyuncs.com` and add one for `oss-accelerate-overseas.aliyuncs.com`.

One more small gap in the same function: an endpoint with an explicit port such as `oss-cn-hangzhou.aliyuncs.com:443` fails the `.aliyuncs.com`-at-the-end check and lands in the same "region must be configured" error, even though the endpoint itself is standard. Stripping a trailing `:<port>` after the path is removed would cover that.

src/paimon/fs/oss/oss_file_system_factory.cpp
```diff
@@ -0,0 +1,201 @@
...
+    constexpr std::string_view kSuffix = ".aliyuncs.com";
+    size_t scheme = endpoint.find("://");
+    if (scheme != std::string::npos) {
+        endpoint.erase(0, scheme + 3);
+    }
+    size_t slash = endpoint.find('/');
+    if (slash != std::string::npos) {
+        endpoint.erase(slash);
+    }
+    if (endpoint.rfind(kPrefix, 0) != 0) {
```
zjw1111: `StringUtils::StartsWith` / `EndsWith` already exist in `common/utils/string_utils.h`, and this file already includes it and uses `StringUtils::StringToValue` elsewhere. The four hand-written prefix/suffix checks in this function — this one, the `.aliyuncs.com` check below, the `-internal` strip, and the `-dualstack` check — would read better through them.

One caveat if you do this: the helpers take `const std::string&`, while `kPrefix` / `kSuffix` / `kInternal` are `constexpr std::string_view`, which does not convert implicitly — so those constants would need to become `constexpr const char*` (or `StringUtils` would need `string_view` overloads). Also note the `.aliyuncs.com` check currently uses `find()` (first occurrence) plus an end-position check, so switching to `EndsWith` slightly changes behaviour for inputs like `oss-x.aliyuncs.com.aliyuncs.com`; that difference disappears if the allow-list suggestion above is adopted.

src/paimon/fs/oss/oss_file_system.cpp
```diff
@@ -0,0 +1,196 @@
...
+                                          int32_t max_keys) const override {
+        PAIMON_RETURN_NOT_OK(ValidateBucket(path));
+        oss2::models::ListObjectsV2Request request;
+        request.setBucket(path.bucket).setPrefix(path.key).setDelimiter("/");
+        if (!continuation_token.empty()) {
+            request.setContinuationToken(continuation_token);
+        }
+        if (max_keys > 0) {
+            // OSS limits ListObjectsV2 requests to 1000 keys.
+            request.setMaxKeys(std::min(max_keys, 1000));
```
zjw1111: Minor: could `1000` become a named constant, e.g. `constexpr int32_t kMaxKeysPerRequest = 1000;`? The comment above explains where the limit comes from, but a named constant keeps it greppable if it is needed elsewhere.

The same nit applies in `oss_file_system_factory.cpp`, where `"-dualstack"` is spelled inline three times within one expression while the neighbouring `-internal` is already a named `constexpr std::string_view kInternal`. That one goes away if the `InferRegion` allow-list suggestion is taken.

src/paimon/fs/oss/oss_file_system_test.cpp
```diff
@@ -0,0 +1,320 @@
...
+    void AddResponse(int status_code, oss2::HeaderCollection headers, std::string body = "") {
+        std::shared_ptr<std::iostream> response_body;
+        if (!body.empty()) {
+            response_body = std::make_shared<std::stringstream>(std::move(body));
+        }
+        responses.emplace_back(std::make_unique<oss2::ResponseMessage>(oss2::ResponseMessage{
+            status_code, "", std::move(headers), std::move(response_body), nullptr}));
+    }
+
+    std::vector<std::unique_ptr<oss2::ResponseMessage>> responses;
```
zjw1111: Small naming nit: `MockHttpTransport` is a class, so Google style wants a trailing underscore on these two data members (`responses_` / `requests_`). `MockHttpClient::status_code_` in `s3_file_system_test.cpp` follows that already. `send` / `getName` above are SDK overrides, so keeping those names is of course correct.

CMakeLists.txt
```diff
@@ -75,8 +76,10 @@ endif()
 if(PAIMON_ENABLE_REST)
     add_definitions(-DPAIMON_ENABLE_REST)
 endif()
-# libcurl backs the HTTP client shared by the S3 file system and the rest catalog.
-if(PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST)
+# libcurl backs the HTTP client shared by the object store file systems and the rest catalog.
```
zjw1111: Now that `PAIMON_ENABLE_OSS` is part of this condition, the comment is no longer accurate. The shared HTTP client (`common/utils/http_client.cpp`) is still gated on `PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST` in `src/paimon/CMakeLists.txt`, so the OSS file system does not use it. OSS needs libcurl for two unrelated reasons: the OSS SDK's own curl transport, and the direct `curl_getdate` call in `oss_file_system.cpp`.

Something like "libcurl is required by the REST catalog and the S3 HTTP client, and by the OSS SDK's curl transport" would describe the dependency more accurately. (If the shared timestamp helper suggested on `oss_file_system.cpp` lands in common code, the direct dependency from the OSS sources goes away as well.)

zjw1111 (commented):
Thanks for the update — the previous round of comments is addressed well: the ASF headers, the `int32_t` / `1000` fixes, the dedicated `Executor` instead of the global singleton, the option-constant unification, the `PAIMON_COMMON_HTTP_CLIENT_TEST_SRCS` gate, and the injected `HttpTransport` tests are all in good shape.

Two functional issues came up on a closer read, plus a few small nits. Details inline:

- `ParseTimeMillis` cannot parse the timestamp format `ListObjectsV2` returns, so every listed entry ends up with modification time `0` (this breaks orphan file cleanup on OSS). Same root cause exists in the S3 implementation, so a shared helper would be nice.
- `InferRegion` still infers a non-existent region for `oss-accelerate-overseas.aliyuncs.com`.

src/paimon/common/fs/object_store_file_system.h
```diff
@@ -31,6 +36,68 @@
 
 namespace paimon {
 
+inline int64_t ParseObjectStoreModificationTime(const std::string& value) {
+    std::tm parsed_time{};
```
lxy-9602: Thank you very much for the PR and for your patience with the revisions. I have one small comment: would it be possible to move `ParseObjectStoreModificationTime` into a class as a static utility function? We generally try to avoid global functions in production code.

mrdrivingduck: OK, will do.

mrdrivingduck:
While testing with OSS file system, I found an issue of executor pool. Extract it for seperated PR and I think it should be addressed first: #233.

lxy-9602 (dismissed):
+1

cmake_modules/ThirdpartyToolchain.cmake
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_factory.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_factory.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system.cpp
mrdrivingduck: Done.

src/paimon/fs/oss/oss_file_system_test.cpp
mrdrivingduck: Done.

CMakeLists.txt
mrdrivingduck: Done.

src/paimon/common/fs/object_store_file_system.h
mrdrivingduck: Done.

mrdrivingduck:
> I identified an OSS SDK v2 issue where a retried request can retain a stale `Authorization` header after its timestamp is regenerated, causing `SignatureDoesNotMatch`.
> 
> This PR temporarily applies a small build-time patch to overwrite the header on retry. The issue has been reported to the OSS SDK maintainers, and they will address it upstream soon.

Fixed in https://github.com/aliyun/alibabacloud-oss-cpp-sdk-v2/releases/tag/0.1.2, so the initial patch has been removed from this commit.

mrdrivingduck:
rebased with #233

lxy-9602 (approved):
+1

PR #207, opened 2026-08-14
perf(parquet): reuse leaf column index set across fields in page-filtered reads

zhf999:
### Purpose

`PageFilteredRowGroupReader::ReadFilteredRowGroup()` reads one top-level field at a time and, for
every field, calls the patched Arrow API `parquet::arrow::FileReader::GetColumn(i, column_indices,
iterator_factory, out)` to build a leaf-filtered reader tree. That API used to take a
`const std::vector<int>&` and internally convert it with `VectorToSharedSet(column_indices)`, so a
fresh `std::unordered_set<int>` was allocated and populated on **every** call, i.e. once per
projected top-level field per row group. For wide projections and many filtered row groups this is a
repeated allocation plus hashing of the same, unchanging leaf index set on the hot read path. The time
complexity is O(n^2) where n is the number of columns.

In our experiments on a table with 1000 columns, `VectorToSharedSet` **may take fixed 3-4s time in total, which
take a large ratio of time in short queries**.

No behavioral change is intended: the set contents are identical to what `VectorToSharedSet()`
produced from the same `column_indices` vector, and the public `ReadFilteredRowGroup()` signature
(which still takes `const std::vector<int32_t>&`) is unchanged. Only the internal helper and the
vendored-Arrow patch signature change.

image

Note: because `cmake_modules/arrow.diff` is part of the change, the Arrow external project must be
re-patched and rebuilt for the new signature to take effect; a stale Arrow install directory will
fail to compile the caller.

### Tests

No new test is added.

### API and Format

No change to public API.

### Documentation

No new feature.

### Generative AI tooling

No.


src/paimon/format/parquet/page_filtered_row_group_reader.cpp
```diff
@@ -397,12 +397,14 @@ Result<std::unique_ptr<arrow::RecordBatchReader>> PageFilteredRowGroupReader::Re
     std::vector<std::shared_ptr<arrow::ChunkedArray>> result_arrays;
     result_arrays.reserve(field_indices.size());
 
+    std::shared_ptr<std::unordered_set<int>> col_indices_set(new std::unordered_set<int>());
+    col_indices_set->insert(column_indices.begin(), column_indices.end());
     // TODO(zhouhongfeng.zhf): This loop could be parallelized.
```
lxy-9602, line 402: use `std::make_shared<std::unordered_set<int>>`

lxy-9602: Could we pass `const std::unordered_set<int>&` to `GetReader()` instead? The set is only queried while synchronously building the reader tree, so `ReaderContext` does not need to own it. This also avoids `shared_ptr` allocation/ref-counting and expresses read-only access. If minimizing changes, consider at least using `std::shared_ptr<const std::unordered_set<int>>`.

zhf999: `included_leaves` is a `std::shared_ptr<std::unordered_set<int>>`, so we must use a `shared_ptr`.

The way we constuct `shared_ptr` has been changed to `make_shared`, and we make the function parameter `const shared_ptr<std::unordered_set<int>>&`, this also aligns with the signature of Arrow function `GetFieldReader`.

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #208, opened 2026-08-17
refactor(fs): make FileStatus and BasicFileStatus concrete value types

lucasfang:
refactor(fs): make FileStatus and BasicFileStatus concrete value types

### Purpose

Linked issue: close #xxx

`FileStatus` and `BasicFileStatus` were pure-virtual interfaces with one
subclass per file system implementation (Local/Jindo/ObjectStore/Mock),
forcing a heap allocation per entry on every `GetFileStatus` /
`ListFileStatus` / `ListDir` call. None of the subclasses added any
behavior or extra state beyond `path` / `length` / `is_dir` /
`modification_time`, so the polymorphism was not pulling its weight.

This change turns both classes into concrete value types (similar to
`arrow::fs::FileInfo`):

- `FileStatus` holds path, length, is_dir and modification_time; all
  virtual methods (including the destructor) are removed.
- `BasicFileStatus` holds path and is_dir; all virtual methods are removed.
- `FileSystem::GetFileStatus` now returns `Result<FileStatus>`;
  `ListFileStatus` fills `std::vector<FileStatus>*`; `ListDir` fills
  `std::vector<BasicFileStatus>*`.
- The four subclasses (`LocalFileStatus`, `JindoFileStatus`,
  `ObjectStoreFileStatus`, `MockFileStatus`) and three basic subclasses
  (`LocalBasicFileStatus`, `JindoBasicFileStatus`,
  `ObjectStoreBasicFileStatus`) are deleted; `local_file_status.h` and
  `jindo_file_status.h` are removed.
- Jindo converts `JdoFileInfo` to paimon types eagerly instead of lazily
  wrapping it.

No behavioral semantics change: unknown modification time still reports
`kUnknownModificationTime`, directory semantics are unchanged, and
`ListDir` still returns basic information only.

### Tests

No new cases added; existing coverage exercises the changed paths:

- `file_system_test.cpp` (ListDir / GetFileStatus / ListFileStatus for
  local and object store), `object_store_file_system_test.cpp`,
  `resolving_file_system_test.cpp`, `jindo_file_system_test.cpp`
- Core suites: `orphan_files_cleaner`, `file_store_commit_impl` (gmock
  FileSystem), `snapshot_manager`, `manifest_file`, `merge_tree_writer`,
  `lookup_levels`, `append_only_writer`, `commit_message`, etc.
- Integration suites: read/write/scan/clean/pk-compaction inte tests

### API and Format

Yes, this is a breaking C++ API change in `include/paimon/fs/file_system.h`:

- `FileStatus` / `BasicFileStatus` are no longer polymorphic and cannot
  be subclassed; they are now passed by value.
- `FileSystem::GetFileStatus` signature changes from
  `Result<std::unique_ptr<FileStatus>>` to `Result<FileStatus>`;
  `ListFileStatus` and `ListDir` out-parameters change from
  `std::vector<std::unique_ptr<...>>*` to `std::vector<...>*`.
- Third-party `FileSystem` implementations must update their overrides.

No effect on storage format or wire protocol.

### Documentation

No new feature; only class doc comments were updated to reflect value
semantics.

### Generative AI tooling

Generated-by: Qoder


src/paimon/testing/mock/mock_file_system.h
```diff
@@ -121,15 +102,15 @@ class MockFileSystem : public FileSystem {
     Status Delete(const std::string& path, bool recursive = true) const override {
         return Status::OK();
     }
-    Result<std::unique_ptr<FileStatus>> GetFileStatus(const std::string& path) const override {
-        return std::make_unique<MockFileStatus>();
+    Result<FileStatus> GetFileStatus(const std::string& path) const override {
+        return FileStatus(/*path=*/"", /*length=*/0, /*is_dir=*/false, /*modification_time=*/0);
```
lxy-9602, line 106: Which value do we plan to use here — `0`, `kUnknownModificationTime` and `kNoSize`?

lucasfang: 0 will be fine for this mock class

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #209, opened 2026-08-17
feat(prefetch): support read-ahead cache for Parquet reads

### LINKED ISSUE #218: [Feature] Support read-ahead cache for Parquet reads
### Search before asking

### Motivation

# [Feature] Support read-ahead cache for Parquet reads

Labels: enhancement

### Search before asking

### Motivation

The framework-level prefetch (`PrefetchFileBatchReader` + shared read-ahead cache) currently only has a working `PreBufferRange()` implementation for ORC, so Parquet reads cannot feed the shared cache and get no latency hiding from it. Several concrete problems follow from that:

1. **Parquet reads cannot use the shared prefetch cache**: `ParquetFileBatchReader::PreBufferRange()` is unimplemented, so on high-latency storage (OSS/S3/HDFS) every column-chunk read is a synchronous round trip and sequential scans pay the full IO latency.
2. **Double prefetch / double memory**: enabling arrow's internal `pre_buffer` (`parquet.read.enable-pre-buffer`) as a workaround makes arrow's `ReadRangeCache` hold a second in-memory copy of the same column chunks, doubling memory for prefetched data.
3. **SeekToRow double-read amplification**: `FileReaderWrapper::SeekToRow()` issued before the first `Next()` rebuilds a batch reader that `PrepareForReading` immediately discards, and since the patched arrow `GetRecordBatchReader` eagerly reads every column chunk, building the reader twice doubles the requested bytes.
4. **No observability / configurability**: there are no hit/miss or prefetch IO counters for the cache, and users cannot switch cache prefetch off for predicate/bitmap-index queries where prefetching is wasteful.

### Solution

1. **Parquet pre-buffer ranges**: implement `ParquetFileBatchReader::PreBufferRange()` via `FileReaderWrapper::GetPreBufferRanges()`, computing dictionary-page-aware column-chunk ranges for fully-matched row groups and page-level byte ranges for page-filtered row groups. Unlike the arrow-internal PreBuffer path, this also covers row groups excluded by read-range dispatch, because the shared prefetch cache must serve data consumed by all sub-readers.
2. **Read-ahead cache integration**: `PrefetchFileBatchReaderImpl` creates a shared `ReadAheadCache` per file reader, registers ranges via `Init()` and dispatches prefetch via `Warmup()` ahead of the readers; positional reads issued through `CacheInputStream` are served from the cache via a copy-out `Read()` API, with in-flight fetches awaited instead of re-fetched.
3. **Public config-only API**: keep the cache implementation internal (`src/paimon/common/utils/read_ahead_cache.h`) and expose only `PrefetchCacheMode` (`ALWAYS` / `EXCLUDE_PREDICATE` / `EXCLUDE_BITMAP` / `EXCLUDE_BITMAP_OR_PREDICATE` / `NEVER`) and `CacheConfig` tuning knobs (buffer/range/hole/pre-buffer limits) in a new public header `include/paimon/utils/prefetch_cache_config.h`, consumed by `ReadContextBuilder::SetPrefetchCacheMode()` / `WithCacheConfig()`.
4. **Single prefetch layer**: default `parquet.read.enable-pre-buffer` to false so arrow's `ReadRangeCache` does not hold a second in-memory copy of column chunks already fetched by the shared cache; the option stays available to restore the arrow-internal prebuffer when the read-ahead cache is disabled.
5. **SeekToRow fix**: record a pre-initialization seek target in `FileReaderWrapper` (`pending_start_idx_`) and consume it in the deferred `PrepareForReading`, so no batch reader is built twice and no extra IO is issued.
6. **Metrics**: expose read-ahead cache counters (`read-ahead-cache.read.count/bytes`, `.read.hits/hit-bytes/misses/miss-bytes`, `.io.count/bytes`) merged into `BatchReader::GetReaderMetrics()`, with cache buffers released on reader `Close()` (`ReleaseBuffers()`) while keeping counters readable for post-scan metric aggregation.

### Anything else?

The cache design is adapted from the Apache ORC C++ reader cache (https://github.com/apache/orc/blob/main/c%2B%2B/src/io/Cache.hh). A bundled secondary fix aligns `ResolveBucketMode` with Java Paimon: `POSTPONE_MODE` is only resolved for primary-key tables.

lucasfang:
feat(prefetch): support read-ahead cache for Parquet reads

### Purpose

Linked issue: close #218 

The framework-level prefetch previously only had working `PreBufferRange()` support for ORC; Parquet reads could not feed the shared read-ahead cache. This change completes the Parquet side and makes the cache observable:

1. **Parquet pre-buffer ranges**: `ParquetFileBatchReader::PreBufferRange()` is implemented via `FileReaderWrapper::GetPreBufferRanges()`, which computes column-chunk ranges (dictionary-page aware) for fully-matched row groups and page-level byte ranges for page-filtered row groups.

2. **Read-ahead cache integration**: `PrefetchFileBatchReaderImpl` creates a shared `ReadAheadCache` per file reader when the read context enables it (`ReadContextBuilder::SetReadAheadCacheEnabled(bool)`, default true). The workloop registers the ranges (`Init`) and dispatches prefetch (`Warmup`) ahead of the readers; positional reads issued through `CacheInputStream` are served from the cache, with in-flight fetches awaited instead of re-fetched. The promise-backed entries are published under the write lock before their fetch IO is dispatched, so a reader racing the prefetch always finds the covering entries and waits on their futures instead of issuing a duplicate underlying read. The cache never evicts: every published range stays cached until `ReleaseBuffers()` / `Reset()`, which keeps every in-flight fetch's entry and future reachable while `ReleaseBuffers()` and the destructor drain the pending IOs; the memory footprint is bounded by one data file's prefetched ranges. `CacheConfig` defaults are aligned with the reader request granularity (32 MiB range limit / 8 KiB hole limit / 256 MiB pre-buffer window).

3. **Copy-out read API**: `ReadAheadCache::Read()` copies the cached bytes directly into the caller's buffer and returns a hit/miss boolean. Multi-segment hits (a request spanning several adjacent coalesced entries) are copied segment by segment into the destination without an intermediate assembled buffer; future waits and the data copy both happen outside the cache lock. Misses leave the destination untouched so `CacheInputStream` falls back to the underlying stream.

4. **Prefetch layering via `ReadHints`**: the framework passes its runtime read state down to the format layer through a new `ReadHints` struct (`prefetch_enabled` / `read_ahead_cache_enabled`) injected via `ReaderBuilder::WithReadHints(const std::optional<ReadHints>&)`, instead of mutating format options. When the hints are present and both framework prefetch and the shared read-ahead cache are active, parquet disables its own pre-buffer (`parquet.read.enable-pre-buffer`) so the column chunks are not buffered twice by arrow's `ReadRangeCache` and the shared cache — both derive their ranges from the same column-chunk metadata. When hints are absent or the shared cache is not active, the option keeps its default of true so those scans still get arrow's `ReadRangeCache` prefetching.

5. **Cache metrics surfaced to users**: the read-request counters (`read-ahead-cache.read.count` / `.bytes`), hit/miss counters (`read-ahead-cache.read.hits` / `.hit-bytes` / `.misses` / `.miss-bytes`) and the prefetch IO counters (`read-ahead-cache.io.count` / `.bytes`, counting the requests actually issued to the underlying stream) are merged into `BatchReader::GetReaderMetrics()`. This includes two fixes:
   - `ConcatBatchReader` closes file readers at EOF, and `Close()` used to reset the cache counters, so metrics read after the scan were always zero. The full `Reset()` now only happens on the reader-reuse entry points (`SetReadSchema` / `RefreshReadRanges`).
   - `ReadAheadCache::ReleaseBuffers()` is added so that `Close()` still frees the prefetched buffers of a finished file immediately (avoiding per-file buffer accumulation across a multi-file scan) while keeping the counters readable.

6. **ResolveBucketMode consistency** (secondary fix bundled in): `POSTPONE_MODE` is now only resolved for primary-key tables; the dead `UNAWARE_BUCKET` branch is removed and the constants are documented.

### Tests

- UT `read_ahead_cache_test.cpp` (15 cases): basics, repeated-read reuse, `TestNoEvictionKeepsAllRanges` (prefetched ranges stay cached no matter how much data accumulates), Init hole coalescing, double-init rejection, re-init after `Reset()`, metrics counters, null-metrics no-op, `TestReleaseBuffersKeepsMetrics` (buffers are dropped but counters survive `ReleaseBuffers()`, and `Reset()` zeroes them), `TestMultiSegmentContiguousHit` (a read spanning several adjacent entries is served segment by segment and counted as one hit), `TestPrefetchIOErrorPropagation` (a failed prefetch surfaces as an error `Status` from `Read()` for both single- and multi-segment reads, via `IOHook`), `TestWarmupPrefetchesBeforeFirstRead` / `TestWarmupWithEmptyRanges`, `TestPreBufferWindowLimit` (the `pre_buffer_limit` window truncates batched prefetch IOs), and `TestInFlightEntryServesRacingReader` (a gated async stream holds the prefetch IO; a concurrent reader is served from the published in-flight entry without triggering a second fetch).
- UT `parquet_file_batch_reader_test.cpp`: end-to-end pre-buffer range coverage incl. page-filtered row groups and cache-hit reads, plus `TestPreBufferReadHints` (pre-buffer is disabled only when the hints report both prefetch and the shared cache active — even if the option explicitly enables it; with partial hints, absent hints, or no override the option/default decides).
- UT `bucket_mode_test.cpp`: updated for the postpone/PK rule.
- IT `read_inte_test.cpp::TestReadAheadCacheMetrics`: parameterized over parquet/orc and read-ahead cache enabled/disabled; asserts the counters are present and `hits > 0` exactly when a cache is created, absent otherwise.

### API and Format

New public API surface; no storage format or protocol changes:

- `include/paimon/read_context.h`: `ReadContextBuilder::SetReadAheadCacheEnabled(bool)` (default true) and `ReadContext::ReadAheadCacheEnabled()`.
- `include/paimon/format/read_hints.h` (new): the `ReadHints` struct carrying the framework runtime read state (`prefetch_enabled` / `read_ahead_cache_enabled`).
- `include/paimon/format/reader_builder.h`: `WithReadHints(const std::optional<ReadHints>&)` (default no-op), letting each format adapt its internal behavior to the runtime state.

The cache itself (`ReadAheadCache`, `ReadAheadCacheMetrics` counter names, `CacheConfig` tuning knobs, `Warmup()`, `ReleaseBuffers()`, and the copy-out `Result<bool> Read(const ByteRange&, char*)`) lives in the internal header `src/paimon/common/utils/read_ahead_cache.h`.

### Documentation

No new configuration options; the read-ahead cache is enabled by default and can be toggled per read via the `ReadContextBuilder` API. Prefetch behavior is covered by the existing `docs/source/user_guide/prefetch.rst`. No documentation changes required.

### Generative AI tooling

Generated-by: Qoder


include/paimon/utils/read_ahead_cache.h
```diff
@@ -162,11 +173,28 @@ class PAIMON_EXPORT ReadAheadCache {
...
+    /// Init() only registers the ranges; without Warmup() the first fetch starts
+    /// when the first Read() arrives, racing the caller's own miss fetch.
+    void Warmup();
+
+    /// Collect hit/miss counters of Read() calls and the prefetch IO
+    /// counters into the given metrics as counters named after
+    /// `ReadAheadCacheMetrics`. Only reads issued through Read() are counted
+    /// as hits/misses; prefetch fetches dispatched by the cache itself are
+    /// counted in the fetch counters instead.
+    void CollectMetrics(const std::shared_ptr<Metrics>& metrics) const;
```
lxy-9602: Use `Metrics*` or `return std::shared_ptr<Metrics>`

src/paimon/core/table/bucket_mode_test.cpp
```diff
@@ -50,13 +50,23 @@ TEST(BucketModeTest, TestResolveBucketMode) {
...
+
+    // UNAWARE_BUCKET is a bucket id, not a bucket number, so it does not mean unaware mode here.
+    EXPECT_EQ(BucketMode::HASH_FIXED,
               ResolveBucketMode(BucketModeDefine::UNAWARE_BUCKET, pk_schema));
+    EXPECT_EQ(BucketMode::HASH_FIXED,
+              ResolveBucketMode(BucketModeDefine::UNAWARE_BUCKET, append_schema));
+
     EXPECT_EQ(BucketMode::HASH_FIXED, ResolveBucketMode(4, append_schema));
+    EXPECT_EQ(BucketMode::HASH_FIXED, ResolveBucketMode(4, pk_schema));
 }
```
lxy-9602, line 64: Do we really need a dedicated test for this error case? It feels quite ambiguous.

include/paimon/utils/read_ahead_cache.h
lucasfang: done

src/paimon/common/utils/read_ahead_cache_test.cpp
```diff
@@ -61,6 +64,25 @@ TestCacheEnv CreateTestFileAndCache(const std::string& filename, const std::stri
     return {path, cache, pool};
 }
 
+// Assert that reading the range is a cache hit filling the destination with
+// the expected content.
+void AssertReadEquals(ReadAheadCache& cache, const ByteRange& range, const std::string& expected) {
+    std::string dest(std::max<size_t>(range.length, 1), 'X');
```
lxy-9602, line 73: Could we move cache to the end and store it as a pointer?

src/paimon/common/utils/read_ahead_cache_test.cpp
```diff
@@ -139,21 +165,284 @@ TEST(TestReadAheadCache, TestCacheEviction) {
...
+    // Single entry: the prefetch is the first IO after the hook is armed.
+    {
+        auto env = CreateTestFileAndCache("data_file", content, config, {{0, 10}});
+        paimon::ScopeGuard guard([&io_hook]() { io_hook->Clear(); });
+        io_hook->Reset(0, paimon::IOHook::Mode::RETURN_ERROR);
+        std::string dest(5, 'X');
+        Result<bool> result = env.cache->Read({0, 5}, dest.data());
+        ASSERT_FALSE(result.ok());
+        EXPECT_NE(std::string::npos,
+                  result.status().ToString().find("io hook triggered io error at position"));
```
lxy-9602: ASSERT_NOK_WITH_MSG

src/paimon/common/utils/read_ahead_cache_test.cpp
```diff
@@ -139,21 +165,284 @@ TEST(TestReadAheadCache, TestCacheEviction) {
...
+    paimon::ScopeGuard guard([&io_hook]() { io_hook->Clear(); });
+    // Any new IO fails: the warmed-up reads must be served without fetching.
+    io_hook->Reset(0, paimon::IOHook::Mode::RETURN_ERROR);
+
+    AssertReadEquals(*env1.cache, {0, 5}, "abcde");
+    AssertReadEquals(*env1.cache, {8, 5}, "ijklm");
+
+    // Without Warmup() the first Read() starts the prefetch and sees the error.
+    std::string dest(5, 'X');
+    ASSERT_FALSE(env2.cache->Read({0, 5}, dest.data()).ok());
```
lxy-9602: ASSERT_NOK

src/paimon/format/parquet/file_reader_wrapper.cpp
```diff
@@ -376,16 +393,23 @@ Status FileReaderWrapper::PrepareForReadingLazy(
...
+std::vector<::arrow::io::ReadRange> FileReaderWrapper::DoCollectPreBufferRanges(
+    const std::vector<int32_t>& column_indices, bool skip_read_range_excluded, uint64_t start_idx) {
     std::vector<::arrow::io::ReadRange> ranges;
     auto file_metadata = file_reader_->parquet_reader()->metadata();
 
-    for (const auto& trg : target_row_groups_) {
-        if (trg.IsExcludedByReadRange()) continue;
+    for (uint64_t idx = start_idx; idx < target_row_groups_.size(); idx++) {
+        const auto& trg = target_row_groups_[idx];
+        if (skip_read_range_excluded && trg.IsExcludedByReadRange()) continue;
```
lxy-9602: {continue;}

src/paimon/format/parquet/file_reader_wrapper.cpp
```diff
@@ -411,6 +435,22 @@ std::vector<::arrow::io::ReadRange> FileReaderWrapper::CollectPreBufferRanges(
...
+            DoCollectPreBufferRanges(target_column_indices_,
+                                     /*skip_read_range_excluded=*/false, /*start_idx=*/0);
+        std::vector<std::pair<uint64_t, uint64_t>> pre_buffer_ranges;
+        pre_buffer_ranges.reserve(ranges.size());
+        for (const auto& range : ranges) {
+            pre_buffer_ranges.emplace_back(static_cast<uint64_t>(range.offset),
+                                           static_cast<uint64_t>(range.length));
+        }
+        return pre_buffer_ranges;
+    }
```
lxy-9602, line 467: Where could an exception be thrown here? `DoCollectPreBufferRanges`? If so, shouldn’t this function return a bad status instead?

src/paimon/format/parquet/parquet_file_batch_reader.cpp
```diff
@@ -635,14 +635,18 @@ Result<std::vector<std::pair<uint64_t, uint64_t>>> ParquetFileBatchReader::GenRe
...
+
 Result<::parquet::ReaderProperties> ParquetFileBatchReader::CreateReaderProperties(
     const std::shared_ptr<arrow::MemoryPool>& pool,
     const std::map<std::string, std::string>& options) {
     ::parquet::ReaderProperties reader_properties;
     // TODO(jinli.zjw): set more ReaderProperties (compare with java)
     PAIMON_ASSIGN_OR_RAISE(
         bool enable_pre_buffer,
-        OptionsUtils::GetValueFromMap<bool>(options, PARQUET_READ_ENABLE_PRE_BUFFER, true));
+        OptionsUtils::GetValueFromMap<bool>(options, PARQUET_READ_ENABLE_PRE_BUFFER, false));
```
zjw1111: The PR description says the format-level prebuffer is disabled through `ReaderBuilder::WithPreBufferEnabled(false)`, scoped to the case where the read-ahead cache is active. `WithPreBufferEnabled` does not exist anywhere in the codebase, and `include/paimon/format/reader_builder.h` is not part of this PR. What this change actually does is flip the default of `parquet.read.enable-pre-buffer` from `true` to `false` for every parquet read, both here and in `CreateArrowReaderProperties` (line 670).

Paths where neither prefetch layer is active after this change:

- `context_->EnablePrefetch() == false`: `abstract_split_read.cpp:148` never constructs `PrefetchFileBatchReaderImpl`, so there is no read-ahead cache at all.
- `PrefetchCacheMode::NEVER`: `prefetch_file_batch_reader_impl.cpp:85` does not create the cache.
- `EXCLUDE_PREDICATE` / `EXCLUDE_BITMAP` / `EXCLUDE_BITMAP_OR_PREDICATE` with a predicate or selection bitmap present: `NeedInitCache()` (`prefetch_file_batch_reader_impl.cpp:307-323`) returns false, so the cache is constructed but `Init()` is never called. Every `CacheInputStream::Read` still routes through `cache_->Read`, misses, and falls back to the underlying stream — a 100% miss rate, plus an extra `InputStream` opened at `prefetch_file_batch_reader_impl.cpp:84-88` and the miss counters folded into the metrics reported by `GetReaderMetrics()`.

In all three cases parquet reads degrade to per-chunk synchronous IO with no prefetch of any kind.

Related: the `enable_buffered_stream()` -> `disable_buffered_stream()` switch on lines 651-653 changes per-read IO granularity and memory behavior for all parquet reads, not just the prebuffer path. `parquet.read.enable-pre-buffer` is a user-visible option and `docs/` does not mention it, so this default change is undocumented.

zjw1111 (commented):
One blocking issue on the prebuffer default, details inline.

lxy-9602 (approved):
+1

SteNicholas (commented):
@lucasfang, I left some comments for this pull request. PTAL.

src/paimon/format/parquet/parquet_file_batch_reader.cpp
SteNicholas: Preserve prebuffering when the shared cache is inactive. This globally changes Parquet's prebuffer default to `false`, but `ReadContext` defaults `EnablePrefetch` to `false`, so ordinary scans never construct the shared cache and now have no prefetch layer. The same regression occurs with `PrefetchCacheMode::NEVER` and excluded predicate/bitmap modes. Please keep Arrow prebuffering enabled unless the shared cache will actually be initialized.

src/paimon/common/utils/read_ahead_cache.cpp
```diff
@@ -193,31 +266,96 @@ void ReadAheadCache::Impl::Reset() {
...
+}
+
+Result<bool> ReadAheadCache::Impl::Read(const ByteRange& range, char* dest) {
     if (range.length == 0) {
-        return ByteSlice{std::make_shared<Bytes>(0, memory_pool_.get()), 0, 0};
+        return true;
     }
+    read_count_.fetch_add(1, std::memory_order_relaxed);
+    read_bytes_.fetch_add(range.length, std::memory_order_relaxed);
     PreBuffer(range.offset);
```
SteNicholas, line 350: Publish in-flight entries before readers can miss.`PreBuffer` marks each range cached before `Cache` is called, but `Cache` does not publish its entries until `MakeCacheEntries` has dispatched or completed every `ReadAsync`. A concurrent reader can therefore see `is_cached=true` while `FindCoveringEntries` still finds nothing, then issue a duplicate underlying read. Please publish the promise-backed entries under the lock before starting I/O, or otherwise expose the pending state to readers.

include/paimon/fs/file_system.h
```diff
@@ -155,7 +155,8 @@ class PAIMON_EXPORT BasicFileStatus {
     }
 
     /// Get the path of this file or directory.
-    std::string GetPath() const {
+    /// @note The returned reference is valid as long as this object is alive.
+    const std::string& GetPath() const {
```
SteNicholas: Avoid changing the public `GetPath` return contract. Changing this public getter from `std::string` to `const std::string&` is unrelated to the cache work and changes source/ABI behavior. It can also turn previously safe code such as `const std::string& p = MakeStatus().GetPath()` into a dangling reference because the owning status temporary is destroyed. Please retain the value-returning API; the same applies to `FileStatus::GetPath` below.

src/paimon/format/parquet/file_reader_wrapper.cpp
```diff
@@ -376,39 +393,68 @@ Status FileReaderWrapper::PrepareForReadingLazy(
...
+Result<std::vector<std::pair<uint64_t, uint64_t>>> FileReaderWrapper::GetPreBufferRanges() {
+    PAIMON_ASSIGN_OR_RAISE(std::vector<::arrow::io::ReadRange> ranges,
+                           DoCollectPreBufferRanges(target_column_indices_,
+                                                    /*skip_read_range_excluded=*/false,
+                                                    /*start_idx=*/0));
+    std::vector<std::pair<uint64_t, uint64_t>> pre_buffer_ranges;
+    pre_buffer_ranges.reserve(ranges.size());
+    for (const auto& range : ranges) {
+        pre_buffer_ranges.emplace_back(static_cast<uint64_t>(range.offset),
+                                       static_cast<uint64_t>(range.length));
```
SteNicholas, line 466: Validate signed Parquet ranges before conversion. These Arrow ranges contain signed metadata values, but negative offsets or `total_compressed_size` values are cast directly to very large `uint64_t`s. `ReadAheadCache::Init()` then calls `CoalesceByteRanges()` before its bounds checks; the combiner performs unchecked `offset + length` arithmetic and can enter a pathological splitting loop, exhausting memory on a corrupt footer. Please validate that offset and length are non-negative and that their sum does not overflow before converting them.

src/paimon/common/utils/read_ahead_cache.cpp
```diff
@@ -89,38 +152,66 @@ class ReadAheadCache::Impl {
...
+
+            size_t total_size = 0;
+            for (const auto& e : entries_) {
+                total_size += e.range.length;
+            }
+            size_t limit = config_.GetBufferSizeLimit();
+            while (!entries_.empty() && total_size + new_entries_size > limit) {
+                auto iter = entries_.begin();
+                total_size -= entries_.front().range.length;
+                entries_.erase(iter);
```
SteNicholas: Keep evicted async reads tracked until completion. This can erase an entry whose asynchronous read is still in flight. The callback keeps the buffer alive, but `ReleaseBuffers()` and the destructor only wait for futures that remain in `entries_`; a later close/destruction can therefore release the stream or memory pool while an evicted request is still writing, and `Bytes` only retains a raw `MemoryPool*`. Please track every dispatched request independently until completion and wait for all of them during release, or avoid evicting in-flight entries.

SteNicholas (commented):
@lucasfang, I have found two additional lifecycle and input-validation issues that should be addressed.

src/paimon/common/utils/read_ahead_cache.cpp
lucasfang: Good point. Rather than tracking evicted requests separately, I've removed the eviction mechanism entirely: the cache now retains every published entry until ReleaseBuffers()/Reset(), since it only holds the prefetched ranges of a single data file and the memory footprint is acceptable. With no eviction, an in-flight fetch can never lose its entry, so the existing waits on entries_ in ReleaseBuffers() and the destructor already cover every dispatched request — no async read can outlive the stream or memory pool. buffer_size_limit is dropped from CacheConfig accordingly.

SteNicholas (commented):
LGTM.

SteNicholas (approved):
LGTM.

PR #210, opened 2026-08-17
feat(file-index): support writing file indexes

zjw1111:
### Purpose

Linked issue: #173

Add the File Index write path and initially support Bitmap and Range Bitmap indexes.

- Add the `FileIndexFormat::Writer` API and Java-compatible V1 index serialization.
- Build per-column indexes while writing append-only, key-value, and shredding data files.
- Embed small indexes in `DataFileMeta` and publish larger indexes as extra `.index` files.
- Share data-writer completion, metadata finalization, writer-level index publication, and abort bookkeeping through `DataFileWriterBase`.
- Add a pool-backed `ByteArrayOutputStream` on top of `MemorySegmentOutputStream`.
- Parse File Index options from core options and expose the manifest threshold through `CoreOptions`.
- Add an end-to-end append-table test which writes external Bitmap and Range Bitmap indexes and consumes them during predicate reads.

Bloom Filter and BSI writers are intentionally deferred to a follow-up change.

This PR covers writer-level and rolling-writer abort cleanup. Cleanup of `DataFileMeta.extra_files` from snapshot expiration, commit abort, and append/merge-tree intermediate compaction paths is tracked as follow-up work.

### Tests

- `cmake --build build --target unittest -j "$(nproc)"` built all unit and integration test targets. The initial run passed 26 of 27 test executables and exposed a null-buffer validation issue in the new `ByteArrayOutputStream` test; the issue was fixed.
- `./build/debug/paimon-common-test` passed all 1439 tests after the fix.
- `./build/debug/paimon-core-test --gtest_filter='DataFileIndexWriterTest.*:CoreOptionsTest.*:AppendOnlyWriterTest.*'` passed all 56 selected tests after relinking the final library.
- The latest review refinements rebuilt `paimon-core-test` and passed 10 focused tests covering append, key-value, `DataFileIndexWriter`, and `FileIndexOptions`.
- Added `WriteAndReadInteTest.TestAppendWithExternalBitmapAndRangeBitmapIndexes` to the parameterized end-to-end suite.
- `pre-commit run --from-ref upstream/main --to-ref HEAD` passed during implementation; the latest changed production files also passed `clang-format --dry-run --Werror` and `git diff --check`.

### API and Format

Yes. This adds the public `FileIndexFormat::Writer` API and writes the Java-compatible V1 File Index container format. Existing read formats and existing public APIs are not changed incompatibly.

### Documentation

No user-facing documentation changes in this PR.

### Generative AI tooling

### DIFF
--- a/include/paimon/file_index/file_index_format.h
+++ b/include/paimon/file_index/file_index_format.h
@@ -92,10 +92,14 @@ class PAIMON_EXPORT FileIndexFormat {
  public:
     class Reader;
     class Writer;
+
+    /// Serialized file indexes grouped as column name -> index type -> index bytes. A null bytes
+    /// pointer represents an empty index for that column and index type.
+    /// For example, indexes["col1"]["bsi"] = <bytes>;
     using ColumnIndexes = std::map<std::string, std::map<std::string, std::shared_ptr<Bytes>>>;
+
     /// Creates a `Reader` to parse a index file (may contain multiple indexes) from the given input
     /// stream.
-    ///
     /// @param input_stream Input stream containing serialized index data.
     /// @param pool Memory pool for temporary allocations during reading.
     /// @return A unique pointer to a `Reader` on success, or an error if the stream is invalid
@@ -130,7 +134,6 @@ class FileIndexFormat::Reader {
  public:
     virtual ~Reader() = default;
     /// Reads index data for a specific column from the index file.
-    ///
     /// @param column_name Name of the column to retrieve index data for.
     /// @param arrow_schema Arrow schema that must contain a field corresponding to `column_name`.
     /// @return A vector of shared pointers to FileIndexReader objects, each corresponding to a
--- a/include/paimon/file_index/file_index_format.h
+++ b/include/paimon/file_index/file_index_format.h
@@ -108,6 +108,10 @@ class PAIMON_EXPORT FileIndexFormat {
         const std::shared_ptr<InputStream>& input_stream, const std::shared_ptr<MemoryPool>& pool);
 
     /// Creates a `Writer` which serializes a complete V1 file index container.
+    ///
+    /// @param output_stream Destination stream for serialized index data.
+    /// @param pool Memory pool for writer-side allocations.
+    /// @return A unique pointer to a `Writer` on success.
     static Result<std::unique_ptr<Writer>> CreateWriter(
         const std::shared_ptr<OutputStream>& output_stream,
         const std::shared_ptr<MemoryPool>& pool);
@@ -126,6 +130,7 @@ class FileIndexFormat::Writer {
     /// Writes all column indexes. This is a terminal, one-shot operation.
     virtual Status WriteColumnIndexes(const FileIndexFormat::ColumnIndexes& indexes) = 0;
 
+    /// Flushes and closes the output stream supplied to `CreateWriter()`.
     virtual Status Close() = 0;
 };
 
--- a/include/paimon/file_index/file_index_format.h
+++ b/include/paimon/file_index/file_index_format.h
@@ -19,6 +19,7 @@
 #pragma once
 
 #include <cstdint>
+#include <map>
 #include <memory>
 #include <string>
 #include <vector>
@@ -32,6 +33,8 @@ struct ArrowSchema;
 namespace paimon {
 class InputStream;
 class MemoryPool;
+class Bytes;
+class OutputStream;
 
 /// Defines the on-disk format and versioning for Paimon file-level indexes.
 /// File index file format. Put all column and offset in the header.
@@ -88,6 +91,8 @@ class MemoryPool;
 class PAIMON_EXPORT FileIndexFormat {
  public:
     class Reader;
+    class Writer;
+    using ColumnIndexes = std::map<std::string, std::map<std::string, std::shared_ptr<Bytes>>>;
     /// Creates a `Reader` to parse a index file (may contain multiple indexes) from the given input
     /// stream.
     ///
@@ -98,10 +103,26 @@ class PAIMON_EXPORT FileIndexFormat {
     static Result<std::unique_ptr<Reader>> CreateReader(
         const std::shared_ptr<InputStream>& input_stream, const std::shared_ptr<MemoryPool>& pool);
 
+    /// Creates a `Writer` which serializes a complete V1 file index container.
+    static Result<std::unique_ptr<Writer>> CreateWriter(
+        const std::shared_ptr<OutputStream>& output_stream,
+        const std::shared_ptr<MemoryPool>& pool);
+
  public:
     static const int64_t MAGIC;
     static const int32_t EMPTY_INDEX_FLAG;
     static const int32_t V_1;
+};
+
+/// Writer for file index file.
+class FileIndexFormat::Writer {
+ public:
+    virtual ~Writer() = default;
+
+    /// Writes all column indexes. This is a terminal, one-shot operation.
+    virtual Status WriteColumnIndexes(const FileIndexFormat::ColumnIndexes& indexes) = 0;
+
+    virtual Status Close() = 0;
 };
 
 /// Reader for file index file.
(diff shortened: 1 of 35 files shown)


include/paimon/file_index/file_index_format.h
```diff
@@ -88,6 +91,8 @@ class MemoryPool;
 class PAIMON_EXPORT FileIndexFormat {
  public:
     class Reader;
+    class Writer;
+    using ColumnIndexes = std::map<std::string, std::map<std::string, std::shared_ptr<Bytes>>>;
     /// Creates a `Reader` to parse a index file (may contain multiple indexes) from the given input
```
lxy-9602, line 101: Please add a comment explaining what `ColumnIndexes` represents.

src/paimon/common/file_index/file_index_format.cpp
```diff
@@ -39,6 +41,102 @@ namespace paimon {
...
+    explicit FileIndexFormatWriterImpl(const std::shared_ptr<OutputStream>& output_stream)
+        : output_stream_(output_stream) {
+        assert(output_stream_);
+    }
+
+    Status WriteColumnIndexes(const FileIndexFormat::ColumnIndexes& indexes) override {
+        if (written_) {
+            return Status::Invalid("File index column indexes have already been written");
+        }
+        int64_t header_length = sizeof(int64_t) + sizeof(int32_t) * 3 + sizeof(int32_t);
```
lxy-9602: Please add comments for header_length, why not `sizeof(int32_t) * 4`

src/paimon/common/file_index/file_index_format.cpp
```diff
@@ -39,6 +41,102 @@ namespace paimon {
...
+        }
+        closed_ = true;
+        PAIMON_RETURN_NOT_OK(output_stream_->Flush());
+        return output_stream_->Close();
+    }
+
+ private:
+    template <typename T>
+    static Status AddChecked(T value, int64_t* total, const char* name) {
+        PAIMON_RETURN_NOT_OK(ValidateValueInRange<int32_t>(value, name));
```
lxy-9602, line 156: If `total` is an output param, please move to the last.

src/paimon/common/file_index/file_index_format_test.cpp
lxy-9602: The bytes in this test look very similar to those in `TestWriteEmptyIndexGoldenBytes`; both are for the empty case. Can we reuse them instead of duplicating them?

src/paimon/common/io/byte_array_output_stream.cpp
```diff
@@ -0,0 +1,79 @@
...
+
+Result<std::shared_ptr<Bytes>> ByteArrayOutputStream::Finish() {
+    PAIMON_RETURN_NOT_OK(Close());
+    if (result_) {
+        return result_;
+    }
+    // TODO(jinli.zjw): Support int64_t lengths in MemorySegmentUtils::CopyToBytes and remove this
+    // limit.
+    if (position_ > std::numeric_limits<int32_t>::max()) {
+        return Status::Invalid("Byte array output stream size exceeds INT32_MAX");
```
lxy-9602: Why isn’t `ValidateValueInRange` used here?

src/paimon/common/io/byte_array_output_stream.cpp
```diff
@@ -0,0 +1,79 @@
...
+    }
+    const std::vector<MemorySegment>& segments = output_.Segments();
+    result_ = std::shared_ptr<Bytes>(new Bytes(static_cast<size_t>(position_), pool_.get()),
+                                     [pool = pool_](Bytes* bytes) { delete bytes; });
+    MemorySegmentUtils::CopyToBytes(segments, /*offset=*/0, result_.get(),
```
lxy-9602, line 75: This implementation feels a bit unusual. Would it make sense to add a helper on `Bytes` instead, for example:

```cpp
static std::shared_ptr<Bytes> AllocateShared(
    size_t size, std::shared_ptr<MemoryPool> pool);
```

src/paimon/common/io/memory_segment_output_stream.cpp
```diff
@@ -54,12 +54,20 @@ void MemorySegmentOutputStream::WriteString(const std::string& str) {
 }
 
 void MemorySegmentOutputStream::Write(const char* data, uint32_t size) {
-    auto bytes = std::make_shared<Bytes>(size, pool_.get());
-    if (size != 0) {
-        memcpy(bytes->data(), data, size);
```
lxy-9602, line 59: Why doesn’t `Write(const char* data, uint32_t size)` delegate to `Write(const MemorySegment& segment, int32_t offset, int32_t len)`? `MemorySegment` can be constructed via `WrapView`.

src/paimon/core/io/data_file_index_writer.h
```diff
@@ -0,0 +1,93 @@
...
+                        const std::shared_ptr<MemoryPool>& pool);
+
+    Result<std::shared_ptr<Bytes>> SerializeContainer();
+    Status WriteExternal(const std::string& path, const std::shared_ptr<Bytes>& bytes);
+
+    std::vector<IndexWriterEntry> writers_;
+    int64_t in_manifest_threshold_;
+    std::shared_ptr<FileSystem> file_system_;
+    std::shared_ptr<DataFilePathFactory> path_factory_;
+    std::shared_ptr<MemoryPool> pool_;
```
lxy-9602, line 95: Is `pool_` placed last because none of the other member variables in this class use it?

src/paimon/core/io/data_file_writer.cpp
```diff
@@ -45,28 +46,13 @@ DataFileWriter::DataFileWriter(
...
 
-void DataFileWriter::SetMetadataFinalizer(MetadataFinalizer finalizer) {
-    metadata_finalizer_ = std::move(finalizer);
-}
-
 Status DataFileWriter::Write(ArrowArray* batch) {
     int64_t record_count = batch->length;
-    PAIMON_RETURN_NOT_OK(SingleFileWriter::Write(batch));
+    PAIMON_RETURN_NOT_OK(WriteRecord(batch, batch));
     seq_num_counter_->Add(record_count);
```
lxy-9602, line 52: The two parameters of `WriteRecord` are hard to understand in the current form.

src/paimon/core/io/file_index_options.cpp
```diff
@@ -0,0 +1,102 @@
...
+        }
+        const size_t index_type_length =
+            key.size() - std::string(kFileIndexPrefix).size() - std::string(kColumnsSuffix).size();
+        const std::string index_type =
+            key.substr(std::string(kFileIndexPrefix).size(), index_type_length);
+        if (index_type.empty()) {
+            return Status::Invalid(fmt::format("Invalid file index option {}", key));
+        }
+        for (std::string column_name : StringUtils::Split(value, ",", /*ignore_empty=*/false)) {
+            StringUtils::Trim(&column_name);
```
lxy-9602, line 64: ignore_empty=true or false?

zjw1111: We intentionally keep ignore_empty=false so leading or middle empty column entries, such as f1,,f2, are rejected instead of silently ignored. Java String.split drops trailing empty tokens, so malformed values such as f1,f2,, currently behave differently. I reverted the ad-hoc special case and added a TODO to align this together with ConfigParser::ParseList, so list option parsing can be changed consistently.

src/paimon/common/io/byte_array_output_stream.cpp
zjw1111: Thanks. I removed the custom deleter instead of adding AllocateShared. ByteArrayOutputStream::Finish now receives a MemoryPool* and constructs the result with std::make_shared<Bytes>. The API documents that the caller must keep the pool alive until the returned Bytes is destroyed, and the upper layer guarantees that lifetime. With this ownership contract, a helper that captures shared_ptr<MemoryPool> is no longer needed.

src/paimon/core/io/data_file_index_writer.h
zjw1111: Yes, this is about destruction order. Members are destroyed in reverse declaration order. The earlier members do not rely on DataFileIndexWriter::pool_ during destruction: each current FileIndexWriter keeps its own shared_ptr<MemoryPool>. pool_ here is used directly only while serializing the container, so declaring it after writers_ is safe. If a future member borrows this pool instead of owning it, pool_ should be declared before that member.

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #211, opened 2026-08-17
feat(metrics): expose scan and prefetch metrics

gripleaf:
### Purpose

  Linked issue: N/A

  Expose point-in-time metrics snapshots for scan planning, prefetch readers, and prefetch I/O.

  This change:

  - Exposes scan planning metrics through `TableScan::GetMetrics()`.
  - Adds scan duration, manifest processing, snapshot cache, and lazy decode metrics.
  - Adds prefetch queue, batch, latency, and error metrics.
  - Adds synchronous and asynchronous prefetch I/O metrics.
  - Ensures returned metrics are snapshots that cannot mutate internal metric state.
  - Preserves cumulative prefetch metrics across reader schema changes.

  ### Tests

  - Scan metric snapshots and scan duration histograms.
  - Snapshot manifest cache hits, misses, and lifecycle metrics.
  - Lazy decode scanned and materialized row metrics.
  - Prefetch batch, queue, latency, error, and I/O metrics.
  - Metrics retained across `SetReadSchema()`.
  - Adaptive prefetch disabling.
  - Default empty metrics for table scans without specialized metrics.

  No new integration tests were added.

  ### API and Format

  Yes, this change affects the public API under `include/paimon/`:

  - Adds `TableScan::GetMetrics()`.
  - Adds the public `ScanMetrics` metric names.
  - Adds the public `PrefetchMetrics` and `PrefetchIoMetrics` metric names.

  This change does not affect the storage format or protocol.

  ### Documentation

  Yes.

  A metrics reference has been added to the user guide. It documents metric names, types, units, meanings, snapshot behavior, and lifecycle semantics for scan, prefetch, and prefetch I/O metrics.

  ### Generative AI tooling


include/paimon/table/source/scan_metrics.h
```diff
@@ -30,6 +32,21 @@ class ScanMetrics {
...
+    static constexpr char SNAPSHOT_CACHE_HITS[] = "snapshotCacheHits";
+    static constexpr char SNAPSHOT_CACHE_MISSES[] = "snapshotCacheMisses";
+    static constexpr char LAST_SNAPSHOT_CACHE_LOAD_DURATION[] = "lastSnapshotCacheLoadDuration";
+    static constexpr char SNAPSHOT_CACHE_LOAD_DURATION[] = "snapshotCacheLoadDuration";
+    static constexpr char LAST_SNAPSHOT_CACHE_STORE_DURATION[] = "lastSnapshotCacheStoreDuration";
+    static constexpr char SNAPSHOT_CACHE_STORE_DURATION[] = "snapshotCacheStoreDuration";
+    // Candidate manifest-entry rows inspected by lazy scan filtering.
+    static constexpr char LAST_LAZY_DECODE_SCANNED_ROWS[] = "lastLazyDecodeScannedRows";
+    // Full manifest entries retained after lazy scan filtering.
+    static constexpr char LAST_LAZY_DECODE_MATERIALIZED_ROWS[] = "lastLazyDecodeMaterializedRows";
```
lxy-9602, line 51: Could you please add comments to clearly distinguish the metrics that exist only in C++ but not in Java?

lxy-9602:
Also, it looks like PR #209  already added metrics for the read-ahead cache. Could you please merge that in and clean things up a bit?

gripleaf:
> Also, it looks like PR #209 already added metrics for the read-ahead cache. Could you please merge that in and clean things up a bit?

I had rebase the main branch and fix the conflict.

lxy-9602:
Thank you for the PR. For the prefetch-related part, please wait for @lucasfang  review. Thanks again for your patience!

src/paimon/core/table/system/system_table_test.cpp
```diff
@@ -199,4 +201,17 @@ TEST(SystemTableTest, TestGlobalSystemTableWithoutCatalogReturnsNotImplemented)
...
+    SystemTableScan scan("/tmp/table");
+    std::shared_ptr<Metrics> metrics = scan.GetMetrics();
+    ASSERT_TRUE(metrics);
+    metrics->SetCounter("external", 1);
+
+    std::shared_ptr<Metrics> second_metrics = scan.GetMetrics();
+    ASSERT_TRUE(second_metrics);
+    Result<uint64_t> external_counter = second_metrics->GetCounter("external");
+    ASSERT_FALSE(external_counter.ok());
+    ASSERT_EQ(external_counter.status().code(), StatusCode::KeyError);
```
lxy-9602: Consider use `ASSERT_NOK_WITH_MSG`.

gripleaf:
Hi @lucasfang , just following up on this PR. The prefetch-related changes are still waiting for your review as suggested by @lxy-9602 . Whenever you have some time, could you please take a look? Thanks a lot!

lucasfang:
Thank you for your contribution. I reviewed the prefetch-related part at 33f83bb. Overall the design is sound — routing `DelegatingPrefetchReader::GetReaderMetrics()` through `prefetch_reader_` is a real fix (previously, when the adaptive strategy disabled prefetch, `GetReader()` returned `readers_[0]`, snapshot semantics via `MetricsImpl::Overwrite` are correct, `MetricsInputStream` covers every `InputStream`/`Stream` virtual, and the async callback captures the metrics state by `shared_ptr` so a late completion cannot outlive it. A few things I'd like to see addressed before merge.

lucasfang:
The PR description no longer matches the diff.** It still claims "Adds the public `ReadAheadCacheMetrics` metric names", "Adds `ReadAheadCache::GetMetrics()`", and read-ahead cache "hit, miss, wait, eviction, pending request, and latency metrics", but this diff touches no `read_ahead_cache.*` file — those belong to #209, as your own comment in `GetReaderMetrics()` states. Please also drop the corresponding bullets from the Tests and "API and Format" sections.

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -42,8 +42,135 @@ class Schema;
...
+    }
+
+    Result<int64_t> Length() const override {
+        return stream_->Length();
+    }
+
+ private:
+    template <typename ReadFunction>
+    Result<int64_t> RecordRead(ReadFunction&& read, int64_t size) {
+        metrics_->read_requests.fetch_add(1);
```
lucasfang: Unconditional instrumentation on the hot read path.** `MetricsInputStream::RecordRead` and the `ReadAsync` completion path each take two `steady_clock::now()` samples plus one `ObserveHistogram`, and `ObserveHistogram` acquires three mutexes in sequence (`MetricsImpl::histogram_lock_` → `HistogramWindowingImpl::mu_` → `HistogramImpl::mu_`) on a histogram shared by all prefetch reader threads. For page-level reads that is per-I/O lock traffic on the critical path with no way to disable it. There is precedent for gating this in the repo: ORC reader metrics are opt-in via `orc.read.enable-metrics`. Two options, either is fine by me:
- only wrap the stream in `MetricsInputStream` when I/O metrics are enabled by an option (zero cost when off), or
- keep the counters (they are plain relaxed atomics) but replace the per-read latency histogram with lock-free count/sum/min/max, and reserve the windowed histogram for coarser events (per batch, per async completion).

gripleaf: Thanks for highlighting the potential overhead on the I/O hot path. I’ve refined the implementation in two aspects:

  1. The latency histograms have been replaced with lightweight count and sum-us counters. Their atomic updates use std::memory_order_relaxed, minimizing the runtime cost of metrics collection.
  2. A new prefetch.io-metrics.enabled option has been introduced and is disabled by default. When it is disabled, MetricsInputStream is not instantiated at all, ensuring that the underlying I/O path remains free from instrumentation overhead.

lucasfang:
+1

gripleaf:
Hi @lucasfang. I’ve fixed the CI failure. The issue was that a newly added test file from the latest main still used the old Create() signature, which wasn’t covered by the earlier focused build. I’ve updated those call sites and verified the full test suite locally. Could you please rerun the CI?

lxy-9602 (approved):
+1

PR #212, opened 2026-08-18
perf(scan): lazily decode manifest bucket entries

gripleaf:
### Purpose

  Optimize snapshot live manifest cache rebuilding for bucket-filtered scans.

  Previously, rebuilding a bucket-scoped cache entry deserialized every manifest entry before filtering by bucket. This change reads the serialization version and bucket first, and performs full `ManifestEntry` deserialization only for entries belonging to the target bucket.

  The optimization is enabled by default and can be disabled with
  `scan.manifest-entry.lazy-decode.enabled`.

  This change also:

  - Preserves serialization-version validation before accessing version-dependent fields.
  - Reuses cached raw manifest bytes when reading bucket entries.
  - Ensures manifest readers are closed when batch processing exits early.
  - Retains the original full-deserialization path as a configurable fallback.

  ### Tests

  - Default and explicitly configured values of `scan.manifest-entry.lazy-decode.enabled`.
  - Manifest serialization-version validation.
  - Reading entries for individual buckets.
  - Reading a nonexistent bucket and returning an empty result.
  - Verifying that entries from non-target buckets are not fully deserialized.
  - Raw manifest byte-cache reuse across bucket reads.
  - Snapshot live manifest cache hits and cache-miss rebuilding.
  - Result equivalence when lazy decoding is disabled.
  - Compatibility with Java Paimon 0.9 and 1.1 Avro manifests.

  All 8 selected unit tests passed.

  ### API and Format

  This change adds the following public configuration API:

  - Options::SCAN_MANIFEST_ENTRY_LAZY_DECODE_ENABLED
  - CoreOptions::ScanManifestEntryLazyDecodeEnabled()

  The corresponding option key is
  scan.manifest-entry.lazy-decode.enabled, and its default value is true.

  This change does not modify the manifest storage format, serialization version, or external protocol.

  ### Documentation

  The new configuration option and affected internal APIs include English inline documentation.

  No standalone user documentation is added because this option controls an internal scan optimization and does not change scan results.

  ### Generative AI tooling

### DIFF
--- a/src/paimon/core/manifest/manifest_file.cpp
+++ b/src/paimon/core/manifest/manifest_file.cpp
@@ -24,6 +24,7 @@
 
 #include "arrow/c/abi.h"
 #include "arrow/c/bridge.h"
+#include "paimon/common/data/columnar/columnar_row.h"
 #include "paimon/common/utils/arrow/status_utils.h"
 #include "paimon/core/io/rolling_file_writer.h"
 #include "paimon/core/manifest/manifest_entry.h"
@@ -84,6 +85,24 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
     return std::unique_ptr<ManifestFile>(
         new ManifestFile(file_system, reader_builder, writer_builder, compression,
                          manifest_file_factory, target_file_size, pool, options, partition_type));
+}
+
+Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t bucket,
+                                       std::vector<ManifestEntry>* entries) const {
+    return ReadArrowBatches(
+        file_name,
+        [this, bucket, entries](const std::shared_ptr<arrow::StructArray>& batch) -> Status {
+            for (int64_t i = 0; i < batch->length(); i++) {
+                ColumnarRow row(batch->fields(), pool_, i);
+                PAIMON_RETURN_NOT_OK(ManifestEntrySerializer::ValidateVersion(row.GetInt(0)));
+                if (row.GetInt(3) != bucket) {
+                    continue;
+                }
+                PAIMON_ASSIGN_OR_RAISE(ManifestEntry entry, serializer_->FromRow(row));
+                entries->push_back(std::move(entry));
+            }
+            return Status::OK();
+        });
 }
 
 Result<std::vector<ManifestFileMeta>> ManifestFile::Write(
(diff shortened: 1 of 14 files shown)


lxy-9602 (commented):
Thank you for the contribution! I reviewed the code, and the overall design looks fine to me. While I’d like to better understand the practical benefit of this PR in production. For example, in a scenario where we scan 1k manifests and query a specific bucket, roughly how much scan latency reduction can we expect?

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -86,6 +87,24 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
 
+Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t bucket,
+                                       std::vector<ManifestEntry>* entries) const {
+    return ReadArrowBatches(
+        file_name,
+        [this, bucket, entries](const std::shared_ptr<arrow::StructArray>& batch) -> Status {
+            for (int64_t i = 0; i < batch->length(); i++) {
+                ColumnarRow row(batch->fields(), pool_, i);
+                PAIMON_RETURN_NOT_OK(ManifestEntrySerializer::ValidateVersion(row.GetInt(0)));
+                if (row.GetInt(3) != bucket) {
```
lxy-9602: Please add a `bucketGetter` in `ManifestEntrySerializer`, similar to the Java implementation, to avoid using the raw index 3 directly.

src/paimon/core/utils/objects_file.h
```diff
@@ -159,6 +186,7 @@ Status ObjectsFile<T>::Read(const std::string& file_name,
                            reader_builder_->Build(file_input_stream));
     auto reader = std::make_unique<ManifestMetaReader>(std::move(batch_reader),
                                                        serializer_->GetDataType(), pool_);
+    ScopeGuard close_reader([&reader]() { reader->Close(); });
```
lxy-9602: We probably don’t need `ScopeGuard` here, since cleanup is already handled by `ManifestMetaReader`’s destructor via `DoClose`.

gripleaf:
> Thank you for the contribution! I reviewed the code, and the overall design looks fine to me. While I’d like to better understand the practical benefit of this PR in production. For example, in a scenario where we scan 1k manifests and query a specific bucket, roughly how much scan latency reduction can we expect?

image
This optimization mainly improves cold-start performance. In our case, we have around 16k buckets, but only need the manifest for a single bucket. The latency improvement is roughly comparable to the difference between P99 Phase 1 and Current Phase 2 in the figure, which represents about a 60% reduction in latency.

lxy-9602:
Got it! It does look like manifest deserialization takes a significant amount of time. Please address the review comments so the PR can move forward and be merged.

lxy-9602 (approved):
+1

zjw1111:
Please link issue id and fix the wrong format of PR desc (lines after `Validation commands:`)

gripleaf:
> Please link issue id and fix the wrong format of PR desc (lines after `Validation commands:`)

done

SteNicholas (approved):
LGTM.

PR #213, opened 2026-08-18
feat(realtime): improve append table lifecycle and query support

lxy-9602:
### Purpose

Linked issue: #158

This PR completes several follow-up capabilities for append-table real-time reads and writes:

- support batch-level statistics and predicate pushdown in the default Arrow `RealtimeStore`;
- support count readers and independently schedulable disk splits;
- make `CommitWithProgress` retries idempotent;
- preserve real-time progress across compaction and require context recreation after rollback;
- clean up expired and orphan real-time offset files;
- extend integration coverage for commit retry, compaction, rollback, schema rename, predicate pushdown, offset cleanup, and memory reclamation.

### API and Format

- Adds the `realtime.store.stats-mode` option with `none` and `full` modes.
- Adds `StatisticsMode` to `RealtimeStoreFactory::Create`.
- Clarifies retry and rollback behavior in the public APIs.

### Generative AI tooling

### DIFF
--- a/include/paimon/realtime/realtime_store.h
+++ b/include/paimon/realtime/realtime_store.h
@@ -31,6 +31,7 @@
 #include "paimon/realtime/offset_range.h"
 #include "paimon/record_batch.h"
 #include "paimon/result.h"
+#include "paimon/statistics_mode.h"
 #include "paimon/visibility.h"
 
 struct ArrowSchema;
@@ -39,14 +40,6 @@ namespace paimon {
 
 class MemoryPool;
 class Predicate;
-
-/// Statistics collected by a `RealtimeStore` for query predicate pushdown.
-enum class StatisticsMode {
-    /// Do not collect statistics.
-    NONE,
-    /// Collect statistics for all supported fields.
-    FULL,
-};
 
 /// A table record batch and its framework-assigned contiguous offset range.
 ///
--- a/include/paimon/realtime/realtime_store.h
+++ b/include/paimon/realtime/realtime_store.h
@@ -40,6 +40,14 @@ namespace paimon {
 class MemoryPool;
 class Predicate;
 
+/// Statistics collected by a `RealtimeStore` for query predicate pushdown.
+enum class StatisticsMode {
+    /// Do not collect statistics.
+    NONE,
+    /// Collect statistics for all supported fields.
+    FULL,
+};
+
 /// A table record batch and its framework-assigned contiguous offset range.
 ///
 /// The batch contains only table write fields. Row `i` is associated with
@@ -156,13 +164,14 @@ class PAIMON_EXPORT RealtimeStoreFactory {
  public:
     virtual ~RealtimeStoreFactory() = default;
 
-    /// Creates a store configured with the supplied schema, options, and memory pool.
+    /// Creates a store configured with the supplied schema, statistics, options, and memory pool.
     /// @param write_schema Complete table write schema whose ownership is transferred to the
     /// factory. The factory may consume it or retain it in the created store.
+    /// @param statistics_mode Framework-parsed statistics collection mode.
     /// @param options Effective table options available to the store.
     /// @param memory_pool Memory pool provided by the write context.
     virtual Result<std::shared_ptr<RealtimeStore>> Create(
-        std::unique_ptr<::ArrowSchema> write_schema,
+        std::unique_ptr<::ArrowSchema> write_schema, StatisticsMode statistics_mode,
         const std::map<std::string, std::string>& options,
         const std::shared_ptr<MemoryPool>& memory_pool) = 0;
 };
--- a/src/paimon/core/core_options.h
+++ b/src/paimon/core/core_options.h
@@ -40,6 +40,7 @@
 #include "paimon/format/file_format.h"
 #include "paimon/fs/file_system.h"
 #include "paimon/result.h"
+#include "paimon/statistics_mode.h"
 #include "paimon/table/source/startup_mode.h"
 #include "paimon/type_fwd.h"
 #include "paimon/visibility.h"
@@ -48,7 +49,6 @@ namespace paimon {
 
 class ExpireConfig;
 class Cache;
-enum class StatisticsMode;
 
 class PAIMON_EXPORT CoreOptions {
  public:
@@ -107,8 +107,9 @@ class PAIMON_EXPORT CoreOptions {
     int64_t GetSourceSplitOpenFileCost() const;
     std::optional<int64_t> GetScanSnapshotId() const;
     std::optional<int64_t> GetScanTimestampMillis() const;
+    bool RealtimeEnabled() const;
     int64_t GetRealtimeReadViewTtlMillis() const;
-    /// Returns the statistics mode used by real-time memory indexers.
+    /// Returns the statistics mode used by the real-time store.
     StatisticsMode GetRealtimeStoreStatisticsMode() const;
     int32_t GetScanManifestEntryCacheMaxSnapshots() const;
 
--- a/src/paimon/core/core_options.h
+++ b/src/paimon/core/core_options.h
@@ -48,6 +48,7 @@ namespace paimon {
 
 class ExpireConfig;
 class Cache;
+enum class StatisticsMode;
 
 class PAIMON_EXPORT CoreOptions {
  public:
@@ -107,6 +108,8 @@ class PAIMON_EXPORT CoreOptions {
     std::optional<int64_t> GetScanSnapshotId() const;
     std::optional<int64_t> GetScanTimestampMillis() const;
     int64_t GetRealtimeReadViewTtlMillis() const;
+    /// Returns the statistics mode used by real-time memory indexers.
+    StatisticsMode GetRealtimeStoreStatisticsMode() const;
     int32_t GetScanManifestEntryCacheMaxSnapshots() const;
 
     int64_t GetManifestTargetFileSize() const;
--- a/src/paimon/core/operation/expire_snapshots.cpp
+++ b/src/paimon/core/operation/expire_snapshots.cpp
@@ -51,13 +51,14 @@ ExpireSnapshots::ExpireSnapshots(const std::shared_ptr<SnapshotManager>& snapsho
                                  const std::shared_ptr<ManifestList>& manifest_list,
                                  const std::shared_ptr<ManifestFile>& manifest_file,
                                  const std::shared_ptr<FileSystem>& fs, const ExpireConfig& config,
-                                 const std::shared_ptr<Executor>& executor)
+                                 bool realtime_enabled, const std::shared_ptr<Executor>& executor)
     : snapshot_manager_(snapshot_manager),
       path_factory_(path_factory),
       manifest_list_(manifest_list),
       manifest_file_(manifest_file),
       fs_(fs),
       config_(config),
+      realtime_enabled_(realtime_enabled),
       executor_(executor),
       logger_(Logger::GetLogger("ExpireSnapshots")) {}
 
@@ -159,16 +160,18 @@ Result<int32_t> ExpireSnapshots::ExpireUntil(int64_t earliest_snapshot_id,
     PAIMON_ASSIGN_OR_RAISE(Snapshot snapshot, snapshot_manager_->LoadSnapshot(end_exclusive_id));
     retained_snapshots.push_back(snapshot);
     std::set<std::string> retained_offset_files;
-    PAIMON_ASSIGN_OR_RAISE(std::vector<Snapshot> all_snapshots,
-                           snapshot_manager_->GetAllSnapshots());
-    for (const Snapshot& retained_snapshot : all_snapshots) {
-        if (retained_snapshot.Id() < end_exclusive_id) {
-            continue;
-        }
-        std::optional<std::string> offsets_path =
-            RealtimeCommitProperties::GetOffsetsPath(retained_snapshot);
-        if (offsets_path) {
-            retained_offset_files.insert(offsets_path.value());
+    if (realtime_enabled_) {
+        PAIMON_ASSIGN_OR_RAISE(std::vector<Snapshot> all_snapshots,
+                               snapshot_manager_->GetAllSnapshots());
+        for (const Snapshot& retained_snapshot : all_snapshots) {
+            if (retained_snapshot.Id() < end_exclusive_id) {
+                continue;
+            }
+            std::optional<std::string> offsets_path =
+                RealtimeCommitProperties::GetOffsetsPath(retained_snapshot);
+            if (offsets_path) {
+                retained_offset_files.insert(offsets_path.value());
+            }
         }
     }
     std::set<std::string> skipping_sets;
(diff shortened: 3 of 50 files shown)


src/paimon/core/table/source/append_count_reader.cpp
```diff
@@ -35,6 +36,23 @@ Result<int64_t> AppendCountReader::CountRows() {
...
 Result<int64_t> AppendCountReader::CountSingleSplit(const std::shared_ptr<Split>& split) const {
+    std::shared_ptr<RealtimeSplit> realtime_split = std::dynamic_pointer_cast<RealtimeSplit>(split);
+    if (realtime_split) {
+        if (realtime_split->Version() != RealtimeSplit::kCurrentVersion) {
+            return Status::Invalid("unsupported real-time split version");
+        }
+        if (realtime_split->MemoryEndOffset() < realtime_split->CommittedEndOffset()) {
+            return Status::Invalid("real-time memory upper offset is behind committed offset");
+        }
+
```
lxy-9602: No need to check again.

src/paimon/core/core_options.h
```diff
@@ -107,6 +108,8 @@ class PAIMON_EXPORT CoreOptions {
     std::optional<int64_t> GetScanSnapshotId() const;
     std::optional<int64_t> GetScanTimestampMillis() const;
     int64_t GetRealtimeReadViewTtlMillis() const;
+    /// Returns the statistics mode used by real-time memory indexers.
+    StatisticsMode GetRealtimeStoreStatisticsMode() const;
```
lxy-9602, line 113: real-time store.

zjw1111 (commented):
Reviewed the realtime lifecycle / query changes. A few questions on the commit idempotency model and on `StatisticsMode` semantics, two perf points worth confirming, plus one test-coverage gap and three style nits. Inline.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -929,6 +929,19 @@ Result<int64_t> FileStoreCommitImpl::CommitWithProgress(
         commit_messages.push_back(realtime_commit.commit_message);
     }
 
+    PAIMON_ASSIGN_OR_RAISE(std::optional<Snapshot> latest_snapshot,
+                           snapshot_manager_->LatestSnapshot());
+    PAIMON_ASSIGN_OR_RAISE(RealtimeOffsetMap committed_offsets,
+                           RealtimeCommitProperties::ReadOffsets(latest_snapshot, fs_));
+    PAIMON_ASSIGN_OR_RAISE(bool ranges_committed, RealtimeCommitProperties::AreRangesCommitted(
```
zjw1111, line 947: Question on the idempotency model here, not a change request.

This short-circuit decides "already committed" purely from realtime offset-range coverage. It does not look at `(commitUser, commitIdentifier)`, and it does not run `CheckFilesExistence`. Both differ from the non-realtime path in this same file: `FilterAndCommit` calls `CheckFilesExistence` and `FilterCommitted`, and `FilterCommitted` resolves the committed identifier via `LatestSnapshotOfUser(commit_user_)` — matching Java's `FileStoreCommitImpl.filterCommitted`. `CommitWithProgress` -> `Commit` -> `TryCommit` bypasses both.

Two consequences I want to check are intentional:

1. Two writers sharing the same offset ranges are indistinguishable here. If a zombie writer replays ranges that the current writer already committed, this returns `latest_snapshot->Id()` and silently drops `commit_messages`, whereas before this patch `RealtimeCommitProperties::Build` would have rejected the commit as non-contiguous. Coverage of the offset range is treated as proof that *this* writer's files are already in the snapshot, which only holds if offsets are globally owned by one writer.
2. Without `CheckFilesExistence`, a retry after partial cleanup can be accepted on offset coverage alone even if the referenced data files are gone.

Is offset coverage intended to be the sole idempotency key for realtime commits — i.e. offset ownership is guaranteed single-writer by contract above this layer — or is `(commitUser, commitIdentifier)` + `CheckFilesExistence` still expected to be layered in later? The surviving `TODO(xinyu.lxy): Support failure recovery and idempotent retry for real-time commits.` in `TryCommit` suggests the latter, but this block reads like the former.

src/paimon/core/realtime/arrow_realtime_store.cpp
```diff
@@ -264,16 +389,23 @@ Status ArrowRealtimeStore::Write(RealtimeWriteBatch&& write_batch) {
     }
     std::shared_ptr<arrow::StructArray> struct_array =
         checked_pointer_cast<arrow::StructArray>(data);
+    PAIMON_ASSIGN_OR_RAISE(std::optional<BatchStatistics> statistics,
```
zjw1111, line 389: Question on the intended scope of the "optional pruning aid" contract, not a change request.

`MayMatch` states it explicitly:

```cpp
// Statistics are only an optional pruning aid. An unsupported predicate or incomplete
// statistic must retain the batch to avoid false negatives.
return !result.ok() || result.value();
```

and the read path is tested for it (`TestMissingStatisticsRetainsNonMatchingBatch` tolerates `stored.min_values == nullptr`).

The write path here does the opposite: `CollectStatistics` failure propagates out of `Write` and fails the write. `CollectStatistics` depends on `arrow::compute::MinMax` and `arrow::StructArray::Make`, so any type/kernel/allocation failure there turns into a write failure. Net effect: `realtime.store.stats-mode=full` introduces a write failure mode that `none` does not have, for data that is otherwise perfectly writable — statistics are optional for pruning but mandatory for ingestion.

Is the "optional" contract meant to apply only to reads, with statistics collection being a hard requirement on write? Or should a failure here degrade to "no statistics for this batch" the way a missing statistic already degrades on the read side?

include/paimon/realtime/realtime_store.h
```diff
@@ -40,6 +40,14 @@ namespace paimon {
 class MemoryPool;
 class Predicate;
 
+/// Statistics collected by a `RealtimeStore` for query predicate pushdown.
+enum class StatisticsMode {
```
zjw1111: Question on the intended semantics of this enum name.

`StatisticsMode` is declared in top-level `namespace paimon` (and forward-declared in `core_options.h`), with a fully generic name. But everything about it is realtime-specific: it lives in `realtime_store.h`, the doc comment says "collected by a `RealtimeStore`", the option is `realtime.store.stats-mode`, and the accessor is `GetRealtimeStoreStatisticsMode()`.

The values are also a strict subset of Java's `metadata.stats-mode`, which is `none | counts | truncate(n) | full` with default `truncate(16)` (`CoreOptions.java`); Rust `core_options.rs` mirrors that set. Here it is `NONE | FULL` only, and `FULL` does not truncate STRING/BINARY min/max.

So: is `StatisticsMode` meant to eventually carry the same semantics as Java's `metadata.stats-mode` (in which case the missing `counts`/`truncate` values and the realtime-only header placement are staging steps), or is it deliberately a realtime-private concept that just happens to share the name? Judging by the name alone a reader will assume the former; judging by the implementation it is the latter. Worth pinning down now, since the name is already exposed at `paimon` scope and hard to walk back later.

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -157,8 +158,22 @@ Result<int32_t> ExpireSnapshots::ExpireUntil(int64_t earliest_snapshot_id,
     std::vector<Snapshot> retained_snapshots;
     PAIMON_ASSIGN_OR_RAISE(Snapshot snapshot, snapshot_manager_->LoadSnapshot(end_exclusive_id));
     retained_snapshots.push_back(snapshot);
+    std::set<std::string> retained_offset_files;
+    PAIMON_ASSIGN_OR_RAISE(std::vector<Snapshot> all_snapshots,
```
zjw1111: Perf point worth confirming — flagging only, no change requested.

`GetAllSnapshots()` is now called unconditionally on every expire round, purely to build `retained_offset_files`. Before this patch the retained side of `ExpireUntil` only did a point `LoadSnapshot(end_exclusive_id)`, so this adds a full snapshot-directory listing plus one snapshot read per retained snapshot, i.e. O(snapshot count) extra IO per expire round. On object storage that is a listing round-trip plus N GETs, and it happens even for tables that never produced a `.offsets` file at all — the common non-realtime case pays the full cost to build an empty set.

Please confirm this is acceptable for tables with a large retained snapshot count, or that you have measured it and it is noise relative to the manifest work already done in this loop.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -929,6 +929,19 @@ Result<int64_t> FileStoreCommitImpl::CommitWithProgress(
         commit_messages.push_back(realtime_commit.commit_message);
     }
 
+    PAIMON_ASSIGN_OR_RAISE(std::optional<Snapshot> latest_snapshot,
+                           snapshot_manager_->LatestSnapshot());
```
zjw1111, line 944: Perf point worth confirming — flagging only, no change requested.

`LatestSnapshot()` here (line 933) and `ReadOffsets()` (line 935) are both repeated further down the same commit path: `TryCommitOnce` does `snapshot_manager_->LatestSnapshot()` again at line 978, and `RealtimeCommitProperties::Build` re-reads the offsets file. So a single realtime commit now does two latest-snapshot resolutions and two `.offsets` reads, where the offsets read is a metadata-file GET on object storage.

That is on the hot path for every realtime commit, and since `Commit` is invoked with `/*retry_on_conflict=*/false` (line 949) there is no retry loop to amortize it against.

Please confirm the duplication is intentional (e.g. you want the coverage check against a snapshot read strictly before the commit attempt, independent of what `TryCommit` observes) rather than an accidental double read.

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -169,10 +184,22 @@ Result<int32_t> ExpireSnapshots::ExpireUntil(int64_t earliest_snapshot_id,
...
+            RealtimeCommitProperties::GetOffsetsPath(snapshot);
+        if (offsets_path) {
+            expired_offset_files.insert(offsets_path.value());
+        }
         auto status = fs_->Delete(snapshot_manager_->SnapshotPath(id));
         // delete quietly will ignore any status error
         (void)status;
     }
+    for (const std::string& offsets_path : expired_offset_files) {
+        if (retained_offset_files.count(offsets_path) == 0) {
```
zjw1111, line 202: This `retained_offset_files` guard is the one thing preventing a live `.offsets` file from being deleted, and nothing in the test suite exercises the case it exists for.

Sharing happens because `RealtimeCommitProperties::Build` forward-copies `kOffsetsKey` when `realtime_ranges.empty()`, so a non-realtime snapshot (compaction being the obvious one) inherits the `.offsets` path of the realtime snapshot before it. Expiring the realtime snapshot then puts that path into `expired_offset_files` while a retained snapshot still references it — exactly what the `count(...) == 0` check must catch.

`TestRealtimeOffsetFileLifecycle` does not reach that state: it uses two distinct `.offsets` paths and never commits a compaction. Across `test/inte/realtime_write_inte_test.cpp`, `ExpireSnapshots()` only appears in that one test and `CompactAndCommit` only in a different one, so no test ever has an expiring and a retained snapshot pointing at the same `.offsets` file. If the guard regresses, the failure is silent data loss — the offsets file is gone and the realtime data referenced by the retained snapshot becomes unreadable — and CI stays green.

Please add a test: realtime commit -> compaction commit (inherits the same `.offsets` via the empty-ranges forward-copy) -> expire the realtime snapshot -> assert the `.offsets` file still exists and `ReadOffsets` on the retained snapshot still returns the committed ranges.

src/paimon/core/realtime/arrow_realtime_store_test.cpp
```diff
@@ -86,6 +94,21 @@ class ArrowRealtimeStoreTest : public testing::Test {
         return c_schema;
     }
 
+    std::vector<int64_t> ReadIds(const BatchReader::ReadBatchWithBitmap& batch) const {
+        std::shared_ptr<arrow::Array> array =
+            arrow::ImportArray(batch.first.first.get(), batch.first.second.get()).ValueOrDie();
+        std::shared_ptr<arrow::StructArray> struct_array =
+            std::static_pointer_cast<arrow::StructArray>(array);
```
zjw1111: `docs/code-style.md` requires `checked_pointer_cast<T>` instead of `std::static_pointer_cast<T>`, and the example given there is literally this cast:

```cpp
// Good
auto struct_array = checked_pointer_cast<arrow::StructArray>(array);
```

Applies to line 103 as well. Please switch both to `checked_pointer_cast`.

src/paimon/core/core_options.h
```diff
@@ -107,6 +108,8 @@ class PAIMON_EXPORT CoreOptions {
     std::optional<int64_t> GetScanSnapshotId() const;
     std::optional<int64_t> GetScanTimestampMillis() const;
     int64_t GetRealtimeReadViewTtlMillis() const;
+    /// Returns the statistics mode used by real-time memory indexers.
```
zjw1111: "memory indexers" is stale terminology. The concept is `RealtimeStore` now — the option is `realtime.store.stats-mode`, the accessor is `GetRealtimeStoreStatisticsMode`, the enum doc in `realtime_store.h` says "collected by a `RealtimeStore`", and this PR itself renames indexer -> store elsewhere (e.g. `TestReusesIndexer...` -> `TestReusesStore...` in `realtime_context_test.cpp`).

Suggest: `/// Returns the statistics mode used by real-time stores.`

src/paimon/core/realtime/arrow_realtime_store_test.cpp
```diff
@@ -86,6 +94,21 @@ class ArrowRealtimeStoreTest : public testing::Test {
...
+        std::shared_ptr<arrow::Int64Array> ids =
+            std::static_pointer_cast<arrow::Int64Array>(struct_array->field(/*id=*/1));
```
zjw1111: Nit: the parameter of `arrow::StructArray::field` is named `pos`, not `id`. `.clang-tidy` enables `bugprone-argument-comment`, so this mismatched comment is a lint violation. Please use `/*pos=*/1`.

src/paimon/core/operation/file_store_commit_impl.cpp
lxy-9602: Opening a small file should not be very expensive. We can address any performance issues together when we do performance testing.

zjw1111 (approved):
+1

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -931,6 +936,30 @@ Result<int64_t> FileStoreCommitImpl::CommitWithProgress(
 
     std::shared_ptr<ManifestCommittable> committable =
         CreateManifestCommittable(identifier, commit_messages, watermark, /*properties=*/{});
+    PAIMON_ASSIGN_OR_RAISE(std::vector<std::shared_ptr<ManifestCommittable>> pending_committables,
+                           FilterCommitted({committable}));
+    const bool identifier_committed = pending_committables.empty();
```
SteNicholas, line 941: `FilterCommitted` only compares this identifier with the latest identifier for the commit user; it does not establish that this exact `(commit_user, commit_identifier)` produced a snapshot. If commit A succeeds but its response is lost, and a later commit B from the same user succeeds before A is retried, this branch can classify A as committed merely because B has a higher identifier. When A’s ranges are covered it then returns the latest snapshot (B) at line 959, despite the public API promising the snapshot produced by this commit. Please look up the snapshot for the exact identity and return its ID; if the ranges were committed by another identity, reject the retry.

lxy-9602: Thanks for pointing this out. `CommitWithProgress` intentionally follows the same identifier high-watermark semantics as `FilterAndCommit`; its return value is the latest snapshot covering the requested offset progress, so it can be passed to `RefreshCommittedSnapshot`. It is not intended to identify the exact snapshot produced by the retried commit.

I’ve clarified this contract in the API documentation and added a test covering A → B → retry A, where the retry returns B’s latest snapshot ID.

SteNicholas (approved):
LGTM.

PR #215, opened 2026-08-18
feat(read): support file index and predicate pushdown for data evolution

### LINKED ISSUE #214: [Feature] Support file index and predicate pushdown in DataEvolutionSplitRead
### Search before asking

### Motivation

Java Paimon supports file-index evaluation and predicate pushdown for data-evolution reads since [apache/paimon#8839](https://github.com/apache/paimon/pull/8839).

paimon-cpp currently does not support this optimization:

- `DataEvolutionSplitRead` passes a null predicate when creating underlying format readers.
- `DataEvolutionSplitRead::ApplyIndexAndDvReaderIfNeeded` rejects a non-null predicate.
- File indexes therefore cannot be used to skip files or select matching row positions.
- Parquet/ORC readers cannot use predicates to skip row groups or pages.

As a result, queries with selective predicates may scan all data belonging to a data-evolution split, even when file-index metadata or format statistics could eliminate most of it.

Predicate pushdown needs special handling for data evolution. Files in the same field group are merged by physical row position. Applying row-level predicates independently to these readers could filter different positions and break column alignment. In addition, when multiple files contain the same field, the newest field value wins, so an index built from an overwritten copy must not incorrectly eliminate the group.

### Solution

Follow [apache/paimon#8839](https://github.com/apache/paimon/pull/8839) to support predicate pushdown in `DataEvolutionSplitRead`:

- Push applicable predicates to file indexes and format readers for single-file groups.
- For multi-file groups, only skip the whole group when safe; do not filter child readers independently.
- Map predicates by field ID to support schema evolution.
- Keep residual filtering to guarantee correctness.

### Anything else?

_No response_

lxy-9602:
### Purpose

Linked issue: close #214 

Support predicate pushdown in `DataEvolutionSplitRead`, aligned with the Java implementation.

- Push predicates to file indexes and format readers for single-file row-id groups.
- Use file indexes only to skip whole merged groups, avoiding row misalignment between files.
- Exclude system fields and predicates outside the read schema from pushdown.
- Ignore stale indexes for fields overwritten by newer files.
- Compose bitmap index selections with row ranges and deletion vectors.
- Skip blob and vector-store files when evaluating merged-group indexes.

### Tests

Added and updated tests covering:

- Predicate selection by read schema and system-field exclusion.
- Single-file bitmap and format-level predicate pushdown.
- Merged-group pruning and column alignment.
- Stale indexes on overwritten fields.
- Missing physical predicate columns.
- Bitmap selections intersecting row ranges with empty, partial, and non-empty results.
- Physical row-id preservation after bitmap selection.
- Parquet page-level and ORC row-group predicate pushdown without file indexes.

### Documentation
Generated-by: Codex (GPT-5)

### DIFF
--- a/src/paimon/core/operation/data_evolution_split_read.cpp
+++ b/src/paimon/core/operation/data_evolution_split_read.cpp
@@ -483,8 +483,14 @@ Result<bool> DataEvolutionSplitRead::SkipByFileIndex(
         }
         std::vector<DataField> written_fields;
         if (file->write_cols) {
-            PAIMON_ASSIGN_OR_RAISE(written_fields,
-                                   data_schema->GetFields(file->write_cols.value()));
+            std::vector<std::string> data_write_cols;
+            data_write_cols.reserve(file->write_cols->size());
+            for (const auto& write_col : file->write_cols.value()) {
+                if (!SpecialFields::IsSystemField(write_col)) {
+                    data_write_cols.push_back(write_col);
+                }
+            }
+            PAIMON_ASSIGN_OR_RAISE(written_fields, data_schema->GetFields(data_write_cols));
         } else {
             written_fields = data_schema->Fields();
         }
--- a/src/paimon/core/operation/data_evolution_split_read.cpp
+++ b/src/paimon/core/operation/data_evolution_split_read.cpp
@@ -22,6 +22,7 @@
 #include <cassert>
 #include <limits>
 #include <map>
+#include <set>
 #include <string>
 #include <string_view>
 #include <thread>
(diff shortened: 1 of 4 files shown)


lszskye (approved):
+1

src/paimon/core/operation/data_evolution_split_read.cpp
```diff
@@ -424,17 +445,104 @@ Result<std::unique_ptr<BatchReader>> DataEvolutionSplitRead::InnerCreateReader(
...
+            continue;
+        }
+
+        std::shared_ptr<TableSchema> data_schema = context_->GetTableSchema();
+        if (file->schema_id != data_schema->Id()) {
+            PAIMON_ASSIGN_OR_RAISE(data_schema, schema_manager_->ReadSchema(file->schema_id));
+        }
+        std::vector<DataField> written_fields;
+        if (file->write_cols) {
+            PAIMON_ASSIGN_OR_RAISE(written_fields,
```
SteNicholas: `write_cols` can contain `_ROW_ID` or `_SEQUENCE_NUMBER` (the existing row-tracking projection helper explicitly skips them), but these system fields are not part of `TableSchema::Fields()`. Calling `GetFields` directly therefore returns `Invalid` for a filtered merged-group read whenever file-index reads are enabled, even if the file has no index. Please reuse `ProjectFieldsForRowTrackingAndDataEvolution` (removing system fields before index evaluation if needed), or filter system names here. A merged-group regression test with explicit row-tracking columns would cover this.

SteNicholas (approved):
LGTM.

PR #216, opened 2026-08-18
fix: optimize shared-shredding read & fix ORC read-size estimation for nested columns

lszskye:
### Purpose
Linked issue: #220 

This change:
(1) Improves shared-shredding MAP reading performance
(2) Fix incorrect ORC read-size estimation that prevents file prefetching when prefetch is enabled
(3) Fix shared-shredding MAP values with nested dictionary-encoded types.

### Tests
- Added selected-key STRUCT projection tests that verify primitive value-buffer sharing.
- Added ORC coverage for MAP values with `list<dictionary-encoded string>`.
- Added `OrcFileBatchReaderTest` cases for:
  - primitive LIST
  - deeply nested LIST
  - primitive MAP
  - deeply nested MAP
  - STRUCT projection containing LIST and MAP fields
  - unsupported partial projection inside LIST and MAP

### API and Format
No public API changes.

### DIFF
--- a/src/paimon/format/orc/orc_file_batch_reader.cpp
+++ b/src/paimon/format/orc/orc_file_batch_reader.cpp
@@ -238,8 +238,7 @@ Status OrcFileBatchReader::CollectTargetColumnIds(const ::orc::Type* src_type,
             }
             break;
         }
-        case ::orc::TypeKind::LIST:
-        case ::orc::TypeKind::MAP: {
+        default: {
             // Partial field recall inside list/map types is unsupported, so the target must match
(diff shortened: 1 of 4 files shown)


src/paimon/format/orc/orc_file_batch_reader.cpp
lxy-9602: Why not just change `target_column_ids->push_back(src_type->getColumnId());` in default case to `CollectAllColumnIds(src_type, target_column_ids);`?

lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #217, opened 2026-08-19
chore(build): reformat arrow.diff patch sections

### LINKED ISSUE #221: [Feature] Normalize cmake_modules/arrow.diff generation via git diff
### Search before asking

### Motivation

[Feature] Normalize cmake_modules/arrow.diff generation via git diff

Labels: enhancement

### Search before asking

### Motivation

`cmake_modules/arrow.diff` is the patch applied to the bundled Arrow ExternalProject via `PATCH_COMMAND` in `cmake_modules/ThirdpartyToolchain.cmake`. Historically its per-file patch sections were appended ad hoc as changes accumulated, so section ordering was arbitrary and hunks were sometimes grouped under the wrong file headers. This makes the patch hard to review, and every future edit risks producing large noisy diffs just from reordering. Commit `e776426` ("chore: format arrow diff") cleaned up the current file (sections sorted by file path, hunks split under their proper headers, content verified byte-for-byte identical), but there is no documented convention to keep it that way, so the file can drift back into an unstable state.

### Solution

Standardize on generating `arrow.diff` with `git diff` so the section ordering is always stable and reproducible:

1. Check out the pinned Arrow version (matching `PAIMON_ARROW_BUILD_SHA256_CHECKSUM` / the version in `third_party/versions.txt`) in a clean local repository.
2. Apply the intended modifications directly on that checkout.
3. Regenerate the patch with `git diff > cmake_modules/arrow.diff`. Git emits per-file sections in stable alphabetical path order, which guarantees a deterministic layout regardless of the order in which the changes were authored.
4. Document this regeneration workflow (e.g. as a comment header inside `cmake_modules/arrow.diff` or in `CONTRIBUTING.md`) so future contributors follow the same process instead of hand-editing or appending patch fragments.
5. Optionally, note in the PR template/review guidance that `arrow.diff` changes should be regeneratable by the above steps and keep semantic changes reviewable.

### Anything else?

Related: commit `e776426` already reformatted the existing patch to the target layout without any semantic change (verified by comparing the multiset of all added/removed patch lines before and after). This issue tracks establishing the convention so the normalized state is preserved going forward.

lucasfang:
### Purpose

Linked issue: close #221

`cmake_modules/arrow.diff` (the patch applied to the bundled Arrow ExternalProject via `PATCH_COMMAND` in `cmake_modules/ThirdpartyToolchain.cmake`) had its patch sections in an ad-hoc order accumulated over time. This change reorganizes the file so the per-file patch sections are ordered consistently by file path (cmake modules first, then `arrow/io`, then `parquet/`), and splits hunks under their proper file headers (14 sections instead of 10, e.g. `parquet/arrow/reader.h`, `parquet/arrow/reader_internal.h`, `parquet/file_reader.cc`, `parquet/file_reader.h` now carry their own sections).

This is a pure reformat/reorder: every added and removed patch line was verified to be byte-for-byte identical to the previous version (compared with direction-aware line-by-line diff of all `+`/`-` lines), so the patch applied to the Arrow source is semantically unchanged and the Arrow build behavior is not affected.

### Tests

No behavior change, so no new UT/IT cases. Verified by comparing the old and new `arrow.diff`: the multiset of all added and removed patch lines is identical between the two versions, which guarantees the patch applies the same changes to the Arrow source. The Arrow ExternalProject rebuild in CI will additionally confirm the reformatted patch still applies cleanly.

### API and Format

No. This change only touches `cmake_modules/arrow.diff` (a build-time patch file); it does not affect any public API under `include/`, the storage format, or any protocol.

### Documentation

No new feature; no documentation change needed.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/cmake_modules/arrow.diff
+++ b/cmake_modules/arrow.diff
@@ -18,7 +18,7 @@ index 8cb3ec83f5..0765df8fa8 100644
 @@ -983,6 +983,11 @@ if(CMAKE_TOOLCHAIN_FILE)
    list(APPEND EP_COMMON_CMAKE_ARGS -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE})
  endif()
- 
+
 +# Compatibility with bundled dependencies that require old CMake versions.
 +if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.30")
 +  list(APPEND EP_COMMON_CMAKE_ARGS -DCMAKE_POLICY_VERSION_MINIMUM=3.5)
@@ -45,7 +45,7 @@ index b36c38c6d4..f974a33073 100644
    /// \return Status
 -  Status Advance(int64_t nbytes);
 +  virtual Status Advance(int64_t nbytes);
- 
+
    /// \brief Return zero-copy string_view to upcoming bytes.
    ///
 diff --git a/cpp/src/parquet/arrow/reader.cc b/cpp/src/parquet/arrow/reader.cc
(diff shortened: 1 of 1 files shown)


lxy-9602 (approved):
+1

SteNicholas (approved):
LGTM.

PR #226, opened 2026-08-20
feat: support PAGE_ROW_COUNT_LIMIT for parquet format

### LINKED ISSUE #225: [Feature] Support parquet page row count limit
[Feature] Support parquet page row count limit

## Search before asking

## Motivation

The Paimon-cpp Parquet writer currently splits data pages only by byte size (`parquet.page.size`, default 1MB), so a single page may contain hundreds of thousands or even millions of rows. This causes two problems:

1. **Read amplification**: the filtering granularity of the page index (ColumnIndex/OffsetIndex) is entirely determined by page size. When a predicate selects only a few rows, the reader still has to read and decode the whole large page, which dilutes the benefit of page-level filtering.
2. **Inconsistency with parquet-mr**: the Java side limits rows per page via `parquet.page.row.count.limit` (PARQUET-1414, default 20000). The C++ writer lacks the corresponding capability, so files written by C++ and Java differ significantly in page layout, making read-path behavior unpredictable in mixed read/write scenarios.

## Solution

Following PARQUET-1414 of parquet-mr, add a row count limit for Parquet data pages; a page is finished when either the row count or the byte size limit is reached first:

1. **Arrow patch (`cmake_modules/arrow.diff`)**: add a `data_page_row_count_limit` builder method, getter, and the constant `DEFAULT_DATA_PAGE_ROW_COUNT_LIMIT = 20000` (aligned with the parquet-mr default) to `parquet::WriterProperties`; in `column_writer.cc`, extend the page-split condition with a `num_buffered_rows_ >= properties_->data_page_row_count_limit()` check in addition to the byte-size check.
2. **Paimon-side wiring**: add the table option `parquet.page.row.count.limit` in `parquet_format_defs.h` (the key name is identical to parquet-mr's); `ParquetWriterBuilder::PrepareWriterProperties` reads the option and applies it to the arrow `WriterProperties`, falling back to the default 20000 when unset.
3. **Semantics**: the limit is checked at write batch granularity, so a page may exceed the limit by up to one write batch, consistent with parquet-mr; existing files are unaffected, only the page layout of newly written files changes.

## Anything else?

- This feature aligns with PR #459 (feat: support parquet page row count limit) of alibaba/paimon-cpp; its arrow patch parts (properties.h, column_writer.cc) are byte-for-byte identical.
- The default row-count page split changes writer memory behavior: pages of highly compressible data (e.g., all-null columns) finish early due to the row limit, so the writer peak memory stays well below the budget; the assertions in `ParquetFormatWriterTest.TestMemoryControl` are adjusted accordingly.
- Smaller pages bring slightly more page header / offset index metadata overhead, a known trade-off controllable via the row limit.

lucasfang:
### Purpose

Linked issue: close #225

Support splitting Parquet data pages by row count, aligned with parquet-mr's `parquet.page.row.count.limit` (PARQUET-1414). Previously the Parquet writer split pages only by byte size (`parquet.page.size`), so a single page could contain a very large number of rows, leading to coarse page-index filtering granularity, read amplification, and page layouts inconsistent with files written by the Java side. This PR introduces a row count limit (default 20000, same as parquet-mr); a page is finished when either the row count or the byte size limit is reached first.

Changes:

- `cmake_modules/arrow.diff`: patch arrow's `parquet::WriterProperties` to add the `data_page_row_count_limit` builder method, getter, and the constant `DEFAULT_DATA_PAGE_ROW_COUNT_LIMIT = 20000`; extend the page-split condition in `column_writer.cc` with a `num_buffered_rows_ >= properties_->data_page_row_count_limit()` check in addition to the byte-size check.
- `src/paimon/format/parquet/parquet_format_defs.h`: add the table option key `PARQUET_PAGE_ROW_COUNT_LIMIT = "parquet.page.row.count.limit"` (the key name is identical to parquet-mr's).
- `src/paimon/format/parquet/parquet_writer_builder.cpp`: `PrepareWriterProperties` reads the option and applies it to the arrow `WriterProperties`, falling back to the default 20000 when unset.
- `src/paimon/format/parquet/parquet_format_writer_test.cpp`: update the `TestMemoryControl` assertions — with the default row-count page split in effect, all-null pages are RLE-encoded and finish early, so the writer peak memory stays far below the budget; for the non-null case the budget upper bound is relaxed to 2.5x (the budget is mostly held as finished page buffers, which BufferedPageWriter copies into its in-memory sink when the row group is flushed, transiently doubling the footprint; ~2.2x measured).

Note: the limit is checked at write batch granularity, so a page may exceed the limit by up to one write batch (unlike parquet-mr, which enforces the limit exactly).

### Tests

- UT: `ParquetWriterBuilderTest.DefaultPrepareWriterProperties` adds a default-value assertion (`data_page_row_count_limit == 20000`); `ParquetWriterBuilderTest.PrepareWriterProperties` adds an assertion that the option `parquet.page.row.count.limit=40000` is propagated.
- UT: `ParquetFormatWriterTest.TestMemoryControl` adapts the memory assertions to the new default page-split behavior.
- Actual results: `paimon-parquet-format-test` 187/187 passed, `paimon-write-inte-test` 117/117 passed, `paimon-read-inte-test` 278/278 passed.

### API and Format

No change to the public API under `include/`. Adds the table option `parquet.page.row.count.limit` (default 20000). The storage format stays compatible: only the data page layout of newly written files changes (more pages, bounded rows per page); no new format feature is introduced and older readers can read the files normally.

### Documentation

New table option `parquet.page.row.count.limit`: a best-effort limit on the number of rows in a single data page, aligned with parquet-mr's option of the same name (PARQUET-1414), default 20000; a page is finished when either this row count or the byte size (`parquet.page.size`) is reached first. Unlike parquet-mr, which enforces the limit exactly, the limit here is only checked at write-batch granularity, so a page may exceed it by up to one write batch (e.g., with the default limit 20000 and the default write batch size 1024, a page holds at most 20480 rows).

### Generative AI tooling

Generated-by: Qoder


src/paimon/format/parquet/parquet_writer_builder.cpp
```diff
@@ -86,6 +86,11 @@ Result<std::shared_ptr<::parquet::WriterProperties>> ParquetWriterBuilder::Prepa
                            OptionsUtils::GetValueFromMap<int64_t>(options_, PARQUET_PAGE_SIZE,
                                                                   ::parquet::kDefaultDataPageSize));
     builder.data_pagesize(page_size);
+    PAIMON_ASSIGN_OR_RAISE(
+        int64_t page_row_count_limit,
+        OptionsUtils::GetValueFromMap<int64_t>(options_, PARQUET_PAGE_ROW_COUNT_LIMIT,
+                                               ::parquet::DEFAULT_DATA_PAGE_ROW_COUNT_LIMIT));
+    builder.data_page_row_count_limit(page_row_count_limit);
```
lxy-9602, line 105: `parquet.page.row.count.limit <= 0` is not validated, even though parquet-mr requires this value to be greater than zero. I suggest returning Invalid.

src/paimon/format/parquet/parquet_format_defs.h
```diff
@@ -40,6 +40,10 @@ namespace paimon::parquet {
 // write
 static inline const char PARQUET_BLOCK_SIZE[] = "parquet.block.size";
 static inline const char PARQUET_PAGE_SIZE[] = "parquet.page.size";
+// Max number of rows in a single data page, aligned with parquet-mr's
+// "parquet.page.row.count.limit" (PARQUET-1414). A page is finished when either
+// this row count or the page byte size (parquet.page.size) is reached first.
```
zjw1111: This comment overstates the guarantee. In parquet-mr the row count is a hard bound: `ColumnWriteStoreBase` pulls `rowCountForNextRowCountCheck` down to `rowsWrittenSoFar + pageRowCountLimit` (`// Do the check earlier if required to keep the row count limit`), so a page never exceeds the limit. Here the condition is only evaluated at write-batch granularity in `CommitWriteAndCheckPageLimit`, so the actual bound is `ceil(limit / write.batch-size) * write.batch-size` — 20480 rows with the default `write.batch-size=1024`, not 20000.

Please reword this to a best-effort limit and state that a page may exceed it by up to one write batch (arrow-rs describes its equivalent option that way). Same for the PR description, which currently claims page layout consistent with parquet-mr — per-page comparison across engines will not match exactly.

src/paimon/format/parquet/parquet_writer_builder.cpp
```diff
@@ -86,6 +86,16 @@ Result<std::shared_ptr<::parquet::WriterProperties>> ParquetWriterBuilder::Prepa
...
+    PAIMON_ASSIGN_OR_RAISE(
+        int64_t page_row_count_limit,
+        OptionsUtils::GetValueFromMap<int64_t>(options_, PARQUET_PAGE_ROW_COUNT_LIMIT,
+                                               ::parquet::DEFAULT_DATA_PAGE_ROW_COUNT_LIMIT));
+    // Aligned with parquet-mr's checkArgument(rowCount > 0) for parquet.page.row.count.limit.
+    if (page_row_count_limit <= 0) {
+        return Status::Invalid(fmt::format("Option '{}' should be greater than 0, but got {}",
+                                           PARQUET_PAGE_ROW_COUNT_LIMIT, page_row_count_limit));
+    }
+    builder.data_page_row_count_limit(page_row_count_limit);
```
zjw1111, line 105: Related: the effective granularity of this limit comes from `builder.write_batch_size(batch_size_)` at line 61, i.e. Paimon's `write.batch-size` (default 1024). The two options are semantically unrelated, so setting `write.batch-size` to e.g. 65536 silently turns `parquet.page.row.count.limit=20000` into a no-op — the first check already lands well past the limit, giving 65536 rows per page.

At minimum this coupling needs to be documented in the option description. A stronger fix is to bound the batch size used for the page-size check so the row limit stays meaningful regardless of `write.batch-size`.

zjw1111 (commented):
The `<= 0` validation in 54dda55 looks good. One remaining point on the semantics of the new option: the row limit is best effort here, not the hard bound parquet-mr gives, and its granularity is currently tied to an unrelated option. Details inline.

lucasfang:
Closing this PR as it negatively impacts write and batch scan/read performance.

PR #227, opened 2026-08-20
feat(core): support primary-key managed BLOBs and field-group compaction

### LINKED ISSUE #204: [Feature] Support managed BLOBs in primary-key tables and normal-file compaction across data-evolution field groups
### Search before asking

### Motivation

This tracks the roadmap item "Extend Data Evolution to primary-key tables and support compaction across evolved field groups" in #186.

One clarification on the wording first: neither Java nor C++ allows a row-tracking/data-evolution table to define primary keys (`SchemaValidation.validateRowTracking` on both sides). What the roadmap item maps to in Java is two capabilities:

1. **Managed BLOBs in primary-key tables** — [`[core][flink][spark] Support managed BLOBs in primary-key tables` (apache/paimon#8617)](https://github.com/apache/paimon/pull/8617), extended by [`[core] Allow first-row merge engine for primary-key managed BLOB tables` (apache/paimon#9201)](https://github.com/apache/paimon/pull/9201). BLOB payloads are externalized to shared `.managed.blob` pack files before entering the MergeTree write buffer, data files store small descriptors plus a `.blobref` reference sidecar, and compaction rewrites descriptors verbatim while rebuilding the exact per-data-file reference set.
2. **Compaction across data-evolution field groups** — [`[core] Introducer compaction for data-evolution table` (apache/paimon#6828)](https://github.com/apache/paimon/pull/6828), with the planning rules of [`[core] Optimize data evolution compaction planning` (apache/paimon#9177)](https://github.com/apache/paimon/pull/9177). Files sharing the same row-id range form one evolved field group; compaction merges the field groups of a contiguous row-id run into a single normal file without changing any row id.

paimon-cpp has neither today:

- A primary-key table with a BLOB column cannot even be created: `SchemaValidation::ValidateRowTracking` requires `data-evolution.enabled` for BLOB columns, which in turn forbids primary keys — an unsatisfiable pair.
- A data-evolution table is never compacted: auto compaction cannot run on `bucket = -1`, and `AppendCompactCoordinator::Run` only implements the plain append rewrite, which would reorder rows and break row ids.

### Solution

Port the two Java capabilities:

**1. Managed BLOB storage for primary-key tables** (apache/paimon#8617 + #9201)

- `PrimaryKeyBlobExternalizer`: externalize non-null blob values of insert rows into rolling `.managed.blob` packs (sealed by `blob.target-file-size`, copied through a `blob.copy-buffer-size` buffer) before they enter the write buffer; retract rows drop the payload. Uncommitted packs are deleted on abort/close; `PrepareCommit` seals and hands them over.
- `ManagedBlobReferenceFile`: the `.blobref` sidecar of each data file, byte-compatible with Java (magic/version/count/`writeUTF`/CRC32, sorted and deduplicated), plus `ManagedBlobReferenceCollector` on the key-value write path so a compacted file lists exactly the packs its surviving rows still reference.
- Lifecycle: the sidecar travels in the data file's extra files and dies with it (writer abort, uncommitted-file cleanup, snapshot expiration); pack files are shared and never deleted by table maintenance (orphan-files clean skips `.managed.blob`).
- Read path: `ManagedBlobResolvingBatchReader` resolves descriptors to payload bytes with one ranged read per surviving value after merging; `blob-as-descriptor` returns the serialized descriptors instead.
- Schema validation: only `deduplicate` / `partial-update` / `first-row` merge engines, `changelog-producer` must stay `none`, no `data-file.external-paths` (rejected on the raw option, even an empty string), managed blob fields cannot be primary/bucket/sequence keys or order a sequence group, and sequence-group-protected managed blob fields reject aggregate functions that need the retracted payload.
- Postpone-bucket writers externalize the same way, so `bucket = -2` tables hold descriptors too.

**2. Compaction across data-evolution field groups** (apache/paimon#6828 → #9177 planning rules)

- `DataEvolutionCompactCoordinator`: group files by exact row-id range into evolved field groups; bin weight is `sum(max(file_size, source.split.open-file-cost))`; a bin becomes a task once its weight strictly exceeds `target-file-size`; a heavier-than-target group is compacted alone; a row-id gap always cuts the bin; a bin needs at least `compaction.min.file-num` files.
- `DataEvolutionNormalCompactTask`: read the group through `DataEvolutionSplitRead` (newest field group wins per column), rewrite into exactly one output file, keep the input's first row id and the merged `[min, max]` sequence-number range so `_ROW_ID` stays stable.
- Entry point: `AppendCompactCoordinator::Run` plans with the data-evolution coordinator when `data-evolution.enabled` is set (scanning every live file instead of only small ones), and rejects the removed legacy mode `data-evolution.compaction.rewrite-row-ids=true` like Java does.

### Anything else?

Known scope cuts against current Java, disclosed in the user docs (`primary_key_table.rst`, `compaction.rst`):

- Only top-level scalar `BLOB` columns are managed; `ARRAY<BLOB>` / `MAP<K, BLOB>` and `blob-descriptor.source-table` (source-table FileIO credentials) are not supported yet.
- Compaction does not yet include: deletion-vector rewrite/materialization (tables with `deletion-vectors.enabled` keep being rejected, see #169 for the read side), the projected-manifest candidate planning with ~100k-file batches and multi-round commits of apache/paimon#9177, dedicated BLOB pack compaction, and vector-store file planning (tables holding vector-store files are rejected until a `VECTOR` type lands).

Follow-up regression tests that need the runtime to pin expectations: concurrent partial-update/append commits racing a planned compaction (Java `DataEvolutionTableTest`), and compaction across real schema ids after `ALTER TABLE`.

SteNicholas:
### Purpose

Linked issue: close #204

Adds two independent capabilities to the write and compaction paths.

**1. Managed BLOB storage for primary-key tables**

A BLOB column of a primary-key table can now keep its payload out of the data file
entirely. Payloads are externalized into rolling `.managed.blob` packs *before* they
reach the merge-tree write buffer, so neither the buffer nor its spill ever carries
payload bytes — only the fixed-size descriptor that replaces them.

- `PrimaryKeyBlobExternalizer` sits in front of the merge-tree writer, seals a pack when
  it reaches `blob.target-file-size`, and hands the packs it opened to the commit message
  its writer produces. A retract row drops its payload rather than storing one, and a
  batch of nothing but retracts opens no pack at all. An input that already holds a
  descriptor is re-materialized so the value is stored under this table's own packs.
- Each data file keeps a `.blobref` sidecar (`ManagedBlobReferenceFile`, a versioned
  binary format with a golden-bytes test against the Java writer) listing the packs its
  rows reference. `ManagedBlobReferenceCollector` writes it on close; it travels in the
  data file's `extra_files`, so snapshot expiration, orphan cleaning and abort all treat
  it as part of the data file.
- Reads go through `ManagedBlobResolvingBatchReader`, which resolves each stored
  descriptor back to payload bytes with one ranged read per value.
- Ownership is explicit: a pack belongs to the commit message whose writer created it, and
  a compaction that merely inherits a pack does not own it. `UncommittedFileCleaner` rolls
  back exactly the packs a failed commit — or a `PrepareCommit` that gives up partway —
  created, and leaves the inherited ones alone.
- `SchemaValidation` enforces the merge-engine, primary-key, sequence-group ordering and
  option rules the layout relies on, so an unsupported combination (`pk-clustering-override`,
  `blob-descriptor.source-table`, a managed BLOB in a key or ordering field) is rejected at
  schema time rather than silently changing semantics.

Only top-level scalar BLOB columns are managed; a BLOB nested in a ROW/MAP/ARRAY and the
existing inline descriptor/view fields are untouched.

**2. Compaction across the evolved field groups of a data-evolution table**

`AppendCompactCoordinator` previously refused a data-evolution table. It now plans such a
table with `DataEvolutionCompactPlanner`: files covering the exact same rows form one
evolved field group, the groups of a contiguous row-id run are bin-packed, and each task
rewrites one bin into a single normal file holding every non-dedicated column — preserving
row ids and the merged file-level sequence-number range.

- Because the rewrite keeps every input row, the deletions of the replaced groups are
  *re-keyed* onto the rewritten file and committed in the same snapshot rather than
  dropped. `DataEvolutionCompactDeletionVectorRewriter` performs the migration and
  `ConflictDetection` verifies it: a vector that went missing, shrank, or was left behind
  fails the commit instead of resurrecting deleted rows. Both bitmap32 and bitmap64 vectors
  are supported.
- `MaterializeDeletionVectors` is the heavy alternative entry point: it applies the
  deletions physically, which reassigns the row ids of the surviving rows and therefore
  drops the global indexes over the touched partitions in the same commit. It is never done
  automatically, and a range covered by a dedicated blob or vector-store file is rejected
  rather than corrupted. Its commit uses the new
  `RowIdCheckConflictForMaterializeDeletionVectors` range rule, since a rewrite of whole
  row ranges cannot rely on the narrower column-overlap check.
- Both `RunAndCommit` and `MaterializeDeletionVectors` split the row id space into bounded
  rounds and commit each round on its own, so only one round's file metadata is live at a
  time. The split is a soft target: a cut can only land where one data manifest's row id
  coverage ends before the next one's begins, and a snapshot without usable row id
  statistics falls back to a single round. Rounds are independent — a failed round leaves
  the committed ones committed and a later call re-plans the rest.
- Options that decide the compaction path or the physical layout (row tracking, data
  evolution, deletion vectors, bucket, blob layout) are now rejected when an override would
  contradict the persisted schema, and the legacy
  `data-evolution.compaction.rewrite-row-ids=true` mode fails instead of being ignored.

### Tests

**Managed BLOB — unit**

| Case group | Covers |
| --- | --- |
| `PrimaryKeyBlobExternalizerTest` (13) | descriptor replacement, per-field packs, pack rolling by target size, retract-only batches, `PrepareCommit` hand-over, seal failure still closing the pack stream, inline descriptor fields left alone, re-materializing descriptor input, IO failure |
| `ManagedBlobReferenceFileTest` (14) | round trip with sort + dedup, empty list, non-ASCII, **Java golden bytes**, and rejection of corruption, trailing bytes, bad magic, unsupported version, negative/oversized count, malformed surrogate, nested relative path |
| `ManagedBlobReferenceCollectorTest` (8) | collecting only declared columns, ignoring non-descriptor values, empty sidecar, abort deleting the sidecar, use-after-close and unknown-field rejection |
| `ManagedBlobResolvingBatchReaderTest` (7) | descriptor resolution preserving nulls, declared-columns-only, pass-through without managed fields, missing pack / non-struct / wrong-type / non-descriptor rejection |
| `UncommittedFileCleanerTest` (4) | rolling back data files, sidecars and *owned* packs; not touching an inherited pack; tolerating missing files; one unusable message not stranding the rest |
| `SchemaValidationTest.TestPrimaryKeyManagedBlob{,SequenceGroups}` | merge-engine, key, ordering and unsupported-option rules |
| `SingleFileWriterTest.TestAbortExecutorRemovesCompanionFiles` | an abort executor removes companion files and stays repeatable |

**Managed BLOB — integration (`PkBlobTableInteTest`, 10 cases)**

`TestWriteAndReadManagedBlob`, `TestReadWithPrefetchAndReadAheadCache`,
`TestCompactionRebuildsExactBlobReferences`, `TestFirstRowManagedBlobKeepsFirstValue`,
`TestPartialUpdateManagedBlob`, `TestDeleteDropsRow`,
`TestSnapshotExpirationRemovesSidecar`, `TestAbortDeletesTheManagedBlobPacksItRollsBack`,
`TestAbortOfACompactionKeepsTheHistoricalPacks`,
`TestAbortOfAWriteAndCompactionKeepsOnlyTheHistoricalPacks`.
Plus `CleanInteTest.TestOrphanFilesCleanKeepsCompanionFilesOfLiveDataFiles`.

**Compaction — unit**

| Case group | Covers |
| --- | --- |
| `DataEvolutionCompactPlannerTest` (19) | bin packing by target size and open-file cost, row-id gaps and large files cutting a bin, evolved groups packing together, mismatched group ranges / vector-store files rejected, blob files excluded, per-partition planning, `min-file-num`, and the round windowing (cut only at coverage gaps, file budget, fallback without row id statistics, delete-only manifests ignored) |
| `DataEvolutionNormalCompactTaskTest` (3), `DataEvolutionMaterializeDeletionCompactTaskTest` (5) | input union, discontiguous / misaligned / dedicated-storage rejection |
| `DataEvolutionCompactDeletionVectorRewriterTest` (6) | short-circuit without vectors, and rejection of a message that already changed the index, a bucketed message, a rewrite that moved its rows, several normal outputs |
| `DataEvolutionCompactGlobalIndexDropperTest` (2) | nothing dropped for a normal compaction or without replaced normal files |
| `DataEvolutionConflictDetectionTest` (10) | the migration check — a left-behind vector, lost deletions, a shrunken vector, two index files claiming one data file, an uncovered dropped group; accepting a materialized compaction and a fully deleted range |
| `ConflictDetectionTest` (+2), `AppendCompactCoordinatorTest` (+2) | plain append table still refuses dropping data files; row-id existence on compaction; `RunAndCommit` / `MaterializeDeletionVectors` on a plain append table |
| `RowIdRangeConflictCheckerTest` (2), `MaterializedIndexChangesProviderTest` (3) | range overlap rule; global indexes rescanned every attempt, other buckets kept |

**Compaction — integration (`DataEvolutionTableTest`, 33 new cases)**

Field-group compaction (`TestCompactAcrossEvolvedFieldGroups`, with partitions, with a
partition filter, across schema evolution, after a dropped column), deletion-vector
migration (`TestCompactKeepsDeletionVectorsOfOneRowRangeGroup`, bitmap64, shifts across
groups, only the touched index file rewritten, sibling vectors carried, out-of-group
deletion rejected), materialization (`TestMaterializeDeletionVectorsReassignsRowIds`,
no-op without deletions, untouched ranges left alone, global indexes dropped, other
partitions untouched, fully deleted range, option guards), bounded rounds
(`TestRunAndCommitSplitsRowIdSpaceIntoRounds`, with a partition filter, moving deletion
vectors every round for both bitmap kinds), and concurrency
(`TestStaleCompactMessagePreservesConcurrentPartialUpdate`,
`TestSmallFileCompactConflictsWithConcurrentPartialUpdate`,
`TestCompactKeepsConcurrentAppendForNextSmallFileMerge`).
Plus `PkCompactionInteTest.DeduplicateWith{,Bitmap64}DeletionVectors`.

### API and Format

Yes — additive public API, plus one new on-disk file kind.

- `include/paimon/append/append_compact_coordinator.h`: two new static entry points,
  `RunAndCommit` and `MaterializeDeletionVectors`, and `kDefaultCandidateFilesPerRound`.
  `Run` keeps its signature; its contract is documented more tightly — on a data-evolution
  table with deletion vectors the returned vector also holds index-only messages and the
  caller must commit the whole vector in one commit.
- `include/paimon/file_store_commit.h`: new pure virtual
  `RowIdCheckConflictForMaterializeDeletionVectors`. This is a source-breaking change for
  an out-of-tree implementer of the interface; there is none in this repository.
- `include/paimon/format/format_writer.h`: new `LastPayloadRange()` with a `std::nullopt`
  default, so the externalizer can point a descriptor at the bytes it just wrote without
  downcasting the writer the format factory handed it.
- `include/paimon/defs.h`: adds `BLOB_COPY_BUFFER_SIZE`, `BLOB_DESCRIPTOR_SOURCE_TABLE`,
  `PK_CLUSTERING_OVERRIDE` and `DATA_EVOLUTION_COMPACTION_REWRITE_ROW_IDS` (the last three
  documenting rejected configurations), and updates the `DELETION_VECTORS_ENABLED` and
  `DELETION_VECTOR_BITMAP64` notes now that a data-evolution table is compactable and
  bitmap64 is supported.
- **Format**: the `.managed.blob` pack is the existing blob file format. New is the
  `.blobref` sidecar, a versioned binary file with magic, version and a length-prefixed
  reference list; `ManagedBlobReferenceFileTest.TestJavaGoldenBytes` pins it against the
  Java bytes. The sidecar is referenced from `DataFileMeta::extra_files`, an existing field,
  so the manifest format is unchanged. Deletion vector index files, snapshots and manifests
  are unchanged.

### Documentation

Yes.

- `docs/source/user_guide/primary_key_table.rst` gains *Managed BLOB Storage*: how the
  payload is externalized and resolved, the `.blobref` sidecar, pack ownership versus
  reference, and the supported surface (top-level scalar BLOB columns only) with the option
  and merge-engine rules.
- `docs/source/user_guide/compaction.rst` gains *Data-Evolution Table Compaction*, covering
  evolved field groups, the deletion-vector migration and the commit check that verifies it,
  *Materializing Deletions* and why it is a separate opt-in, and *Bounded Rounds and
  Committing*. The previous note that such a table is never compacted is removed.
- `docs/source/user_guide/read.rst` is updated where it stated the same limitation.

### Generative AI tooling

Generated-by: Claude Opus 5 (Claude Code)

### DIFF
--- a/src/paimon/core/append/append_compact_coordinator.cpp
+++ b/src/paimon/core/append/append_compact_coordinator.cpp
@@ -19,18 +19,37 @@
 #include "paimon/append/append_compact_coordinator.h"
 
 #include <algorithm>
+#include <limits>
 #include <memory>
+#include <optional>
 #include <string>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
+#include "fmt/format.h"
+#include "fmt/ranges.h"
+#include "paimon/commit_context.h"
 #include "paimon/common/data/binary_row.h"
+#include "paimon/common/predicate/predicate_filter.h"
 #include "paimon/common/types/data_field.h"
 #include "paimon/common/utils/linked_hash_map.h"
+#include "paimon/common/utils/range_helper.h"
 #include "paimon/core/append/append_compact_task.h"
+#include "paimon/core/append/data_evolution_compact_deletion_vector_rewriter.h"
+#include "paimon/core/append/data_evolution_compact_global_index_dropper.h"
+#include "paimon/core/append/data_evolution_compact_planner.h"
+#include "paimon/core/append/data_evolution_materialize_deletion_compact_task.h"
+#include "paimon/core/append/data_evolution_normal_compact_task.h"
 #include "paimon/core/core_options.h"
+#include "paimon/core/deletionvectors/deletion_vectors_index_file.h"
+#include "paimon/core/index/deletion_vector_meta.h"
+#include "paimon/core/index/index_file_handler.h"
+#include "paimon/core/index/index_file_meta.h"
 #include "paimon/core/io/data_file_meta.h"
+#include "paimon/core/io/data_file_path_factory.h"
 #include "paimon/core/manifest/file_kind.h"
+#include "paimon/core/manifest/index_manifest_file.h"
 #include "paimon/core/manifest/manifest_entry.h"
 #include "paimon/core/manifest/manifest_file.h"
 #include "paimon/core/manifest/manifest_list.h"
@@ -40,11 +59,20 @@
 #include "paimon/core/schema/schema_manager.h"
 #include "paimon/core/schema/table_schema.h"
 #include "paimon/core/snapshot.h"
+#include "paimon/core/table/sink/commit_message_impl.h"
+#include "paimon/core/table/source/deletion_file.h"
 #include "paimon/core/utils/field_mapping.h"
 #include "paimon/core/utils/file_store_path_factory.h"
+#include "paimon/core/utils/index_file_path_factories.h"
 #include "paimon/core/utils/snapshot_manager.h"
+#include "paimon/defs.h"
 #include "paimon/executor.h"
+#include "paimon/file_store_commit.h"
+#include "paimon/fs/file_system.h"
+#include "paimon/logging.h"
 #include "paimon/memory/memory_pool.h"
+#include "paimon/utils/range.h"
+#include "paimon/utils/row_range_index.h"
 namespace paimon {
 
 namespace {
@@ -174,6 +202,51 @@ std::unique_ptr<AppendOnlyFileStoreWrite> CreateFileStoreWrite(
         /*realtime_context=*/nullptr, executor, pool);
 }
 
+/// Rejects caller options that change an immutable table option: they decide which
+/// compaction path runs and the physical layout the rewrite must preserve, so overriding
+/// them could route a data-evolution table into the plain append rewrite (reordering row
+/// ids) or slip a deletion-vector table past its rejection. Compared on the parsed effective
+/// values, so passing an explicit default is still allowed.
+Status ValidateImmutableOptions(const CoreOptions& schema_options,
+                                const CoreOptions& merged_options) {
+    std::vector<std::string> changed_options;
+    if (schema_options.RowTrackingEnabled() != merged_options.RowTrackingEnabled()) {
+        changed_options.emplace_back(Options::ROW_TRACKING_ENABLED);
+    }
+    if (schema_options.DataEvolutionEnabled() != merged_options.DataEvolutionEnabled()) {
+        changed_options.emplace_back(Options::DATA_EVOLUTION_ENABLED);
+    }
+    if (schema_options.DeletionVectorsEnabled() != merged_options.DeletionVectorsEnabled()) {
+        changed_options.emplace_back(Options::DELETION_VECTORS_ENABLED);
+    }
+    // The two deletion vector kinds serialize differently and refuse to merge into one
+    // another, so flipping this at the compaction entry point would either fail the rewrite
+    // or quietly replace a table's vectors with the other kind.
+    if (schema_options.DeletionVectorsBitmap64() != merged_options.DeletionVectorsBitmap64()) {
+        changed_options.emplace_back(Options::DELETION_VECTOR_BITMAP64);
+    }
+    if (schema_options.GetBucket() != merged_options.GetBucket()) {
+        changed_options.emplace_back(Options::BUCKET);
+    }
+    if (schema_options.GetBlobFields() != merged_options.GetBlobFields()) {
+        changed_options.emplace_back(Options::BLOB_FIELD);
+    }
+    if (schema_options.GetBlobDescriptorFields() != merged_options.GetBlobDescriptorFields()) {
+        changed_options.emplace_back(Options::BLOB_DESCRIPTOR_FIELD);
+    }
+    if (schema_options.GetBlobViewFields() != merged_options.GetBlobViewFields()) {
+        changed_options.emplace_back(Options::BLOB_VIEW_FIELD);
+    }
+    if (!changed_options.empty()) {
+        return Status::Invalid(
+            fmt::format("Compaction options must not change immutable table options [{}]: they "
+                        "decide the compaction path and the physical layout the rewrite must "
+                        "preserve.",
+                        fmt::join(changed_options, ", ")));
+    }
+    return Status::OK();
+}
+
 /// Load schema from table path and merge user options with schema options.
 Result<std::pair<std::shared_ptr<TableSchema>, CoreOptions>> LoadSchemaAndOptions(
     const std::string& table_path, const std::map<std::string, std::string>& options,
@@ -193,10 +266,13 @@ Result<std::pair<std::shared_ptr<TableSchema>, CoreOptions>> LoadSchemaAndOption
     }
     PAIMON_ASSIGN_OR_RAISE(CoreOptions core_options,
                            CoreOptions::FromMap(final_options, file_system));
+    PAIMON_ASSIGN_OR_RAISE(CoreOptions schema_core_options,
+                           CoreOptions::FromMap(table_schema->Options(), file_system));
+    PAIMON_RETURN_NOT_OK(ValidateImmutableOptions(schema_core_options, core_options));
     return std::make_pair(table_schema, std::move(core_options));
 }
 
-/// Validate that the table is an append-only unaware-bucket table without DV.
+/// Validate that the table is an append-only unaware-bucket table this coordinator can compact.
 Status ValidateTable(const std::shared_ptr<TableSchema>& table_schema,
                      const CoreOptions& core_options) {
     if (!table_schema->PrimaryKeys().empty() || core_options.GetBucket() != -1) {
@@ -205,10 +281,31 @@ Status ValidateTable(const std::shared_ptr<TableSchema>& table_schema,
             "with UNAWARE_BUCKET mode");
     }
     if (core_options.DeletionVectorsEnabled()) {
-        return Status::NotImplemented(
-            "AppendCompactCoordinator not support for dv in UNAWARE_BUCKET mode");
+        // A data-evolution rewrite preserves row ids, so its deletions can be re-keyed onto
+        // the rewritten files by DataEvolutionCompactDeletionVectorRewriter. The plain append
+        // rewrite reorders rows and has no such migration, so it still refuses.
+        if (!core_options.DataEvolutionEnabled()) {
+            return Status::NotImplemented(
+                "AppendCompactCoordinator does not support deletion vectors in UNAWARE_BUCKET "
+                "mode");
+        }
     }
     return Status::OK();
+}
+
+/// Build the index file handler used to read and rewrite a table's deletion-vector index.
+Result<std::shared_ptr<IndexFileHandler>> BuildIndexFileHandler(
+    const std::shared_ptr<FileStorePathFactory>& path_factory, const CoreOptions& core_options,
+    const std::shared_ptr<MemoryPool>& pool) {
+    PAIMON_ASSIGN_OR_RAISE(
+        std::unique_ptr<IndexManifestFile> index_manifest_file,
+        IndexManifestFile::Create(core_options.GetFileSystem(), core_options.GetManifestFormat(),
+                                  core_options.GetManifestCompression(), path_factory,
+                                  core_options.GetBucket(), pool, core_options));
+    return std::make_shared<IndexFileHandler>(
+        core_options.GetFileSystem(), std::move(index_manifest_file),
+        std::make_shared<IndexFilePathFactories>(path_factory),
+        core_options.DeletionVectorsBitmap64(), pool);
 }
 
 /// Build FileStorePathFactory from core options and table schema.
@@ -227,16 +324,25 @@ Result<std::shared_ptr<FileStorePathFactory>> BuildPathFactory(
         global_index_external_path, core_options.IndexFileInDataFileDir(), pool);
 }
 
-/// Scan the latest snapshot and collect small files grouped by partition.
-Result<LinkedHashMap<BinaryRow, std::vector<std::shared_ptr<DataFileMeta>>>> ScanSmallFiles(
+/// Scan the latest snapshot and collect files grouped by partition. When `small_files_only` is
+/// set, files at or above the compaction trigger size are dropped; a data-evolution plan needs
+/// every live file instead, since whether a file takes part depends on its field group, not
+/// only on its size.
+///
+/// `row_id_window`, when set, restricts the scan to the files whose rows fall in that row id
+/// window, which is how a batched data-evolution run keeps one round's file metadata bounded.
+/// `snapshot_out` receives the snapshot that was scanned, so a later deletion-vector rewrite
+/// reads the very same table state.
+Result<LinkedHashMap<BinaryRow, std::vector<std::shared_ptr<DataFileMeta>>>> ScanFiles(
     const std::shared_ptr<SnapshotManager>& snapshot_manager,
     const std::shared_ptr<SchemaManager>& schema_manager,
     const std::shared_ptr<TableSchema>& table_schema,
     const std::shared_ptr<arrow::Schema>& arrow_schema,
     const std::shared_ptr<arrow::Schema>& partition_schema, const CoreOptions& core_options,
     const std::shared_ptr<FileStorePathFactory>& path_factory,
-    const std::vector<std::map<std::string, std::string>>& partitions,
-    const std::shared_ptr<Executor>& executor, const std::shared_ptr<MemoryPool>& pool) {
+    const std::vector<std::map<std::string, std::string>>& partitions, bool small_files_only,
+    const std::optional<Range>& row_id_window, const std::shared_ptr<Executor>& executor,
+    const std::shared_ptr<MemoryPool>& pool, std::optional<Snapshot>* snapshot_out) {
     auto scan_filter = std::make_shared<ScanFilter>(
         /*predicate=*/nullptr, partitions, /*bucket_filter=*/std::nullopt);
 
@@ -247,8 +353,16 @@ Result<LinkedHashMap<BinaryRow, std::vector<std::shared_ptr<DataFileMeta>>>> Sca
     if (core_options.ManifestDeleteFileDropStats()) {
         scan->EnableDropStats();
     }
+    if (row_id_window.has_value()) {
+        PAIMON_ASSIGN_OR_RAISE(RowRangeIndex row_range_index,
+                               RowRangeIndex::Create({row_id_window.value()}));
+        scan->WithRowRangeIndex(row_range_index);
+    }
 
     PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<FileStoreScan::RawPlan> plan, scan->CreatePlan());
+    if (snapshot_out != nullptr) {
+        *snapshot_out = plan->GetSnapshot();
+    }
     std::vector<ManifestEntry> add_entries = plan->Files(FileKind::Add());
 
     int64_t compaction_file_size = core_options.GetCompactionFileSize(/*has_primary_key=*/false);
@@ -256,7 +370,7 @@ Result<LinkedHashMap<BinaryRow, std::vector<std::shared_ptr<DataFileMeta>>>> Sca
 
     for (const auto& entry : add_entries) {
         const auto& file = entry.File();
-        if (file->file_size < compaction_file_size) {
+        if (!small_files_only || file->file_size < compaction_file_size) {
             partition_files[entry.Partition()].push_back(file);
         }
     }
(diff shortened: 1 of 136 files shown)


src/paimon/core/operation/abstract_file_store_write.cpp
```diff
@@ -188,6 +190,22 @@ Result<std::vector<std::shared_ptr<CommitMessage>>> AbstractFileStoreWrite::Prep
...
     std::vector<std::shared_ptr<CommitMessage>> result;
+    // A writer hands its files over the moment it drains an increment, so once a message exists
+    // nothing the writer does removes its data files, sidecars or managed blob packs. Failing
+    // out of this loop never hands `result` to the caller, so the messages built so far would
+    // be stranded with no one left to clean them up.
+    ScopeGuard prepare_guard([this, &result]() {
+        if (result.empty()) {
+            return;
+        }
+        Status status = UncommittedFileCleaner::Delete(
```
wangyong9999, line 201: `result` may contain a metadata-only upgrade, where `CompactBefore` and `CompactAfter` reuse the same `file_name`. If a later bucket fails during prepare, this guard invokes the cleaner, which treats that file as an uncommitted output and deletes it even though the previous snapshot still references it, including its `.blobref`. Before deleting `CompactAfter`, exclude files also present in `CompactBefore`, as `MergeTreeWriter::DoClose` already does.

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -277,14 +280,31 @@ Status ExpireSnapshots::CleanUnusedDataFiles(const std::string& manifest_list_na
...
+            // reference sidecar). CollectFiles resolves their paths, honoring external data
+            // paths.
+            std::shared_ptr<DataFilePathFactory>& data_file_path_factory =
+                data_file_path_factories[entry.Partition()][entry.Bucket()];
+            if (data_file_path_factory == nullptr) {
+                PAIMON_ASSIGN_OR_RAISE(
+                    data_file_path_factory,
+                    path_factory_->CreateDataFilePathFactory(entry.Partition(), entry.Bucket()));
+            }
+            std::vector<std::string> delete_file_paths =
```
wangyong9999, line 300: Snapshot expiration deletes only the data file and its `.blobref` here, while the orphan cleaner always skips `.managed.blob` and does not support primary-key tables. After an update, first-row merge, or compaction drops the last descriptor reference, the pack remains forever, so managed BLOB storage grows without bound under normal workloads. Build a live set from sidecars reachable from retained snapshots, tags, and branches, then reclaim unreferenced packs.

src/paimon/core/io/primary_key_blob_externalizer.cpp
```diff
@@ -0,0 +1,353 @@
...
+        auto struct_type = checked_pointer_cast<arrow::StructType>(value_type_);
+        std::shared_ptr<arrow::StructArray> externalized_array;
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            externalized_array, arrow::StructArray::Make(new_children, struct_type->fields()));
+        ArrowArray c_array;
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportArray(*externalized_array, &c_array));
+        return std::make_unique<RecordBatch>(batch->GetPartition(), batch->GetBucket(), row_kinds,
+                                             &c_array);
+    }();
+    if (!result.ok()) {
```
wangyong9999, line 305: `Abort()` deletes every successfully written entry in `uncommitted_packs_`, even though the owning writer may already hold buffered or pending data that references those packs. Neither writer is poisoned after the error, so a later `PrepareCommit` can commit dangling descriptors; the same issue occurs after a seal failure because `new_files_` has already been flushed. The failure path must either roll back the pending data and sidecars and make the writer unusable, or delete only packs not referenced by pending files.

src/paimon/core/append/append_compact_coordinator.cpp
```diff
@@ -361,4 +859,210 @@ Result<std::vector<std::shared_ptr<CommitMessage>>> AppendCompactCoordinator::Ru
...
+        }
+        Status commit_status =
+            CommitRound(table_path, commit_user, core_options, messages, executor, pool);
+        if (!commit_status.ok()) {
+            // The round's rewritten data files are unreachable once its commit failed, and the
+            // rounds already committed stay committed: compaction is idempotent, a later run
+            // re-plans whatever is left. A deletion-vector index file the round may have
+            // written is left to orphan cleaning, which is what collects unreferenced index
+            // files.
+            CleanupCompactOutputs(messages, path_factory, core_options);
```
wangyong9999, line 962: `FileStoreCommitImpl::TryCommitOnce` deliberately treats an atomic snapshot-write error as an uncertain result and keeps outputs for recovery via `FilterAndCommit`. This branch instead deletes every `CompactAfter` on any commit error; if the snapshot was committed but the client timed out, the latest snapshot now references missing data files. Give each round a recoverable unique identifier and delete outputs only after confirming the commit did not land; otherwise keep them.

src/paimon/core/operation/commit/conflict_detection.cpp
```diff
@@ -434,11 +801,76 @@ Status ConflictDetection::CheckKeyRange(const std::vector<ManifestEntry>& merged
...
 Status ConflictDetection::CheckRowIdExistence(const std::vector<ManifestEntry>& base_entries,
                                               const std::vector<ManifestEntry>& delta_entries,
-                                              const std::optional<int64_t>& next_row_id) const {
+                                              const std::optional<int64_t>& next_row_id,
+                                              const Snapshot::CommitKind& commit_kind) const {
     if (!options_.DataEvolutionEnabled()) {
         return Status::OK();
     }
 
+    std::vector<const ManifestEntry*> existing_data_files;
```
wangyong9999, line 810: This path does not first verify that a delta DELETE still matches the current file with the same `Identifier` by `first_row_id` and `row_count`. Because those fields are not part of `Identifier`, a stale DELETE can cancel a concurrently reassigned ADD; the materialization output has no row ID and therefore skips the later range check, potentially replacing current data with a stale rewrite. Compare the base ADD and delta DELETE metadata first and reject the commit on any mismatch.

src/paimon/core/append/data_evolution_compact_planner.cpp
```diff
@@ -0,0 +1,319 @@
...
+                // nothing else is packed on top of it.
+                PAIMON_RETURN_NOT_OK(
+                    TriggerTask(bin.Drain(), partition, compact_min_file_num, tasks));
+                CompactBin single_group_bin;
+                PAIMON_RETURN_NOT_OK(single_group_bin.Add(std::move(field_group), weight));
+                PAIMON_RETURN_NOT_OK(
+                    TriggerTask(single_group_bin.Drain(), partition, compact_min_file_num, tasks));
+                continue;
+            }
+            PAIMON_RETURN_NOT_OK(bin.Add(std::move(field_group), weight));
```
wangyong9999, line 174: Blob files are dropped before packing (line 100-105), and a bin then packs several adjacent field groups, so the rewritten file spans a wider row range than the dedicated `.blob` files it left behind. The read side needs those ranges to match exactly, not just to be covered: `MergeRangesAndSort` groups files by *overlapping* closed ranges, and `CreateUnionReader` then requires every bunch of a group to report the same row count and the same first row id.

Concretely, on a data-evolution table with a blob column: commit 1 writes `{id}` for rows [0,2] (data file D1, no blob file); commit 2 writes `{id, payload}` for rows [3,5] (data file D2 + blob file B2 [3,5]). Both read fine today. A compaction with `compaction.min.file-num=2` packs the two field groups into one bin and rewrites them into a single file O [0,5]. The live set is then {O[0,5], B2[3,5]}: they overlap, land in one group, and `CreateUnionReader` fails with "All files in a field merge split should have the same row count." Every scan of that range fails from then on, and the commit is already durable.

`TestBlobFilesExcluded` only covers one field group whose blob file shares its exact range, which is why this shape is not caught.

Blob coverage needs to be part of the packing decision: pass the partition's blob files into `PlanPartition`, compute for each field group the set of blob field ids whose files cover its exact range, and drain the bin whenever the next group's coverage set differs — so a rewritten file never spans a change in dedicated-blob coverage.

src/paimon/core/operation/file_store_commit_impl.cpp
```diff
@@ -107,6 +111,26 @@ bool MatchPartitionSpec(const std::map<std::string, std::string>& partition,
...
+MaterializedBuckets CollectMaterializedBuckets(
+    const std::vector<std::shared_ptr<CommitMessage>>& commit_messages) {
+    MaterializedBuckets result;
+    for (const std::shared_ptr<CommitMessage>& message : commit_messages) {
+        auto message_impl = std::dynamic_pointer_cast<CommitMessageImpl>(message);
+        if (message_impl == nullptr) {
+            continue;
+        }
+        const CompactIncrement& compact_increment = message_impl->GetCompactIncrement();
+        if (DataEvolutionUtils::IsMaterializedCompaction(compact_increment.CompactBefore(),
```
wangyong9999, line 126: `IsMaterializedCompaction` only asks whether every output lacks a `first_row_id`, and `KeyValueDataFileWriter::CreateResult` always builds its `DataFileMeta` with `/*first_row_id=*/std::nullopt` (key_value_data_file_writer.cpp:152). So every ordinary primary-key compaction — which replaces normal files and produces a merged file without a row id — satisfies the predicate and lands in `materialized_buckets`.

The commit is then routed through `MaterializedIndexChangesProvider`, which re-scans the index manifest and emits `FileKind::Delete()` for *every* global-index entry of that (partition, bucket). A plain compaction of a primary-key table carrying a pk-sorted index therefore drops an index it never touched. Tables without a global index still pay an unfiltered index-manifest read on every compaction attempt.

The other two `IsMaterializedCompaction` call sites are already inside the data-evolution path; this one is on the generic commit path and needs its own gate — e.g. only collecting when `options_.DataEvolutionEnabled()`, or requiring the replaced `CompactBefore` files to actually carry a `first_row_id`.

src/paimon/core/append/append_compact_coordinator.cpp
```diff
@@ -281,6 +395,75 @@ std::vector<AppendCompactTask> GenerateCompactTasks(
...
+            path_factory->CreateDataFilePathFactory(message_impl->Partition(),
+                                                    message_impl->Bucket());
+        if (!data_file_path_factory.ok()) {
+            PAIMON_LOG_WARN(logger,
+                            "Skipping cleanup of compact outputs in partition %s bucket %d: %s",
+                            message_impl->Partition().ToString().c_str(), message_impl->Bucket(),
+                            data_file_path_factory.status().ToString().c_str());
+            continue;
+        }
+        for (const auto& file : message_impl->GetCompactIncrement().CompactAfter()) {
```
wangyong9999, line 425: This loop cleans up `CompactAfter()` data files only, so the deletion-vector index files `DataEvolutionCompactDeletionVectorRewriter` already wrote (`CompactIncrement::NewIndexFiles()`) stay on disk when the round's commit fails.

The comment on the failure path says they are "left to orphan cleaning, which is what collects unreferenced index files", but `OrphanFilesCleanerImpl::SupportToClean` accepts only `manifest-*`, `manifest-list-*`, `*.tmp`, and `data-*` with a data-format or `.blobref` suffix. An index file is named `index-<uuid>` (`IndexPathFactory::INDEX_PREFIX`), so it matches nothing and is never reclaimed.

Losing the optimistic-commit race is a normal outcome here — `CheckDeletionVectorMigrationIsComplete` explicitly tells the caller to "give up committing and plan again" — so every retry on a contended table leaves more index files behind for good. `RunMaterializeRound` has the same gap and does not even pass `index_messages` to the cleanup when `DropGlobalIndexes` fails after the rewrite succeeded.

`UncommittedFileCleaner::Delete` in this same PR already deletes data files, sidecars and index files for a message; calling it here (and passing the index messages on the materialize path) would cover this without new code.

PR #232, opened 2026-08-20
feat(benchmark): add format-level Parquet read/write micro-benchmarks

### LINKED ISSUE #228: [Feature] Add format-level Parquet read/write micro-benchmarks
### Search before asking

### Motivation

`benchmark/` currently only contains table-level end-to-end cases (`benchmark_case_read.cpp`, `benchmark_case_write.cpp`, `benchmark_case_pk_write.cpp`, `benchmark_case_mor_read.cpp`). There is no way to measure a change that lives entirely in the format layer — writer properties, encoding passthrough, per-batch overhead in the read loop — without the noise of catalog lookup, split planning, merge/sort and commit.

That makes format-level optimizations hard to justify and hard to review: a proposal either lands without evidence, or it is dropped because nobody can quantify it. Several concrete candidates below are currently blocked on exactly this.

For reference, the Velox Parquet writer effort ([facebookincubator/velox#17988](https://github.com/facebookincubator/velox/issues/17988)) landed a dedicated writer benchmark ([velox#18052](https://github.com/facebookincubator/velox/pull/18052)) alongside the optimizations precisely so each change could be attributed individually. The negative result in that same effort — parallel column writing plateauing at 1.3–1.6x ([velox#18128](https://github.com/facebookincubator/velox/pull/18128), closed unmerged) — was only identifiable because the measurement infrastructure existed.

### Solution

Add format-level micro-benchmarks for the Parquet reader and writer, driving `ParquetFormatWriter` / `ParquetFileBatchReader` directly instead of going through the table API.

Suggested coverage:

- **Writer**: flat primitives; low / medium / high cardinality `VARCHAR`; `DECIMAL` at precision 9 / 18 / 38; nested `STRUCT` / `LIST` / `VECTOR`; multi-batch writes (schema and property setup amortized over N batches); compression codecs already supported by `ParquetWriterBuilder`.
- **Reader**: full scan; projection; page-index filtered reads at varying selectivity; skip-heavy reads (many small surviving row ranges); nested column reads.
- **Metrics**: ns/row and bytes written for the writer, ns/row and bytes read for the reader, so encoding changes that trade CPU for size are visible.

Candidates to measure once this exists — listed here so they are not lost, all currently **unquantified**:

1. **Per-batch fixed cost in `ParquetFileBatchReader::NextBatch`**:
   - `array->Validate()` on every batch (`src/paimon/format/parquet/parquet_file_batch_reader.cpp:619`) — structural validation that arguably belongs behind a debug-only guard;
   - two `MetricsImpl::SetCounter` calls per batch (`:641`, `:642`), each taking a `std::mutex` and doing a `std::map<std::string, uint64_t>` lookup;
   - a fresh `ArrowSchema` export on every batch (`:637`) even though the type tree is constant for the whole file.

   This is the paimon-cpp analogue of the per-page fixed-overhead work in [velox#17989](https://github.com/facebookincubator/velox/pull/17989) (~100 ns/page). Whether it is worth touching depends entirely on batch size, and nothing here should change before there is a number.

2. **Intra-page skip cost in `PageFilteredRowGroupReader::ExecuteSkipReadPattern`** (`src/paimon/format/parquet/page_filtered_row_group_reader.cpp:243`): the data page read plan already skips fully-filtered pages at the IO layer, but inside a partially selected page arrow's `TypedColumnReaderImpl::Skip` decodes into a scratch buffer and discards (`cpp/src/parquet/column_reader.cc:1249` in arrow 17). The residual cost scales with the skipped-but-decoded run length and is unknown today.

### Anything else?

This issue came out of evaluating the Velox Parquet reader/writer performance issues ([velox#17994](https://github.com/facebookincubator/velox/issues/17994), [velox#17988](https://github.com/facebookincubator/velox/issues/17988)) against paimon-cpp. Most of the reader-side items there do not apply, because Velox optimizes its own decoders while paimon-cpp reads through arrow 17, which already has the equivalents:

- boolean batch decoding — arrow already has bulk paths (`CopyBitmap` for the null-free case, `BitBlockCounter` otherwise);
- `BYTE_STREAM_SPLIT` — arrow 17 supports FLOAT / DOUBLE / INT32 / INT64 / FLBA;
- `FIXED_LEN_BYTE_ARRAY` skip correctness — arrow's skip path is correct;
- PDEP/PEXT dispatch — arrow already gates BMI2 on `CpuInfo::HasEfficientBmi2()`, which excludes AMD entirely.

The items that *do* apply are tracked separately, and both need this benchmark to be evaluated properly.

SteNicholas:
### Purpose

Linked issue: close #228

`benchmark/` only had table-level cases, so a change confined to the format layer — writer properties, encoding passthrough, per-batch overhead in the read loop — could not be measured without the noise of catalog lookup, split planning, merge/sort and commit.

This adds `paimon-parquet-format-benchmark`, driving `ParquetWriterBuilder` and `ParquetFileBatchReader` directly. Every axis is swept on its own rather than as a combined matrix, so each case answers one question and a change can be attributed to it.

**Writer cases** — flat primitives; `VARCHAR` by cardinality with and without file-level dictionary encoding; dictionary-encoded `VARCHAR` / `INTEGER` input arrays against flat controls that hold value, cardinality and width fixed; `DECIMAL` at precision 9 / 18 / 38; nested `STRUCT` / `LIST` / `VECTOR` / `MAP`; null density 0–100%; rows per `AddBatch`; column count; row group size; the byte-triggered writer memory threshold; and every codec Parquet accepts.

**Reader cases** — full scan; projection; predicate-filtered reads with page-index filtering on and off; skip-heavy reads driven by a strided selection bitmap whose strides straddle the coalesce hole limit; null density; `DECIMAL`; `DOUBLE`; dictionary against plain files; rows per `NextBatch`; nested column reads.

**Metrics** — `ns_per_row` / `bytes_per_row` alongside `ns_per_input_row` / `bytes_per_input_row`, so settings that prune by different amounts stay comparable (pruning shrinks the per-materialized-row numerator and denominator together), plus the reader's own `row_groups` / `row_groups_after_filter` / `batches` counters. Read cases assert on the rows they materialized, and the process exits non-zero on any `SkipWithError`, so a case that stops measuring what it claims fails instead of looking fast.

Dictionary-encoded input gets cases of its own rather than being folded into the string ones. Paimon has no dictionary `FieldType`, but nothing in `ParquetWriterBuilder` rejects a dictionary Arrow type, so the writer does reach Arrow's decision about whether to pass indices through or materialize them — and each dictionary case is paired with a flat control that holds value, cardinality and width fixed, so the delta is the materialization alone.

### Tests

New `paimon-parquet-format-benchmark-test` (7 cases). The benchmark only builds under `PAIMON_BUILD_BENCHMARKS`, which CI does not set, so this test is what keeps the format-layer assumptions it rests on covered:

- `RegisteredCodecsWrite` — every registered codec writes and reads back with matching values; asserts `"lz4"` is still rejected (it resolves to Arrow's `LZ4_FRAME`, which `parquet::IsCodecSupported` rejects — this is why the sweep registers `lz4_raw` and `lz4_hadoop` instead).
- `DictionaryInputRoundTrip` — dictionary-encoded `VARCHAR` / `INT32` input over multiple `AddBatch` calls, compared value by value against the flat equivalent.
- `DecimalPrecisionRoundTrip` — precision 9 / 18 / 38, checking type and values (38 is the only one reaching `FIXED_LEN_BYTE_ARRAY`).
- `NestedRoundTrip` — `LIST` / `MAP` round trip, full row and per-column projection.
- `FilteredAndBitmapRowCounts` — predicate row counts bounded at the row-group grain, and selection-bitmap lower bound.
- `PlainAndDictionaryFilesAgree` — the two files the encoding case compares must decode identically.
- `MemoryThresholdFlushesRowGroups` — a small `parquet.writer.max.memory.use` cuts extra row groups on both plain and dictionary input, and row-group boundaries do not change the data.

### API and Format

No. Nothing under `include/` changes, and no storage format or protocol is affected. The new files are a benchmark binary and its test; the only existing files touched are `benchmark/CMakeLists.txt` and the benchmark docs.

### Documentation

Yes — `docs/source/examples/benchmark.rst` gains a `paimon-parquet-format-benchmark` section covering the case inventory, how to read each counter (including why filtered cases must be compared on `ns_per_input_row` rather than `ns_per_row`, and that `row_groups_after_filter` is row-group level and does not show page-level pruning), and how the local-FileSystem write path bounds what the absolute numbers mean.

### Generative AI tooling

Generated-by: Claude Opus 5 (Claude Code)

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -63,6 +63,20 @@ if(PAIMON_BUILD_BENCHMARKS)
                          ${PAIMON_BENCHMARK_LINK_TOOLCHAIN}
                          EXTRA_INCLUDES
                          ${CMAKE_SOURCE_DIR})
+
+    add_paimon_benchmark(parquet_format_benchmark
+                         SOURCES
+                         parquet_format_benchmark.cpp
+                         STATIC_LINK_LIBS
+                         arrow
+                         parquet
+                         ${PAIMON_BENCHMARK_STATIC_LINK_LIBS}
+                         test_utils_static
+                         Threads::Threads
+                         ${PAIMON_BENCHMARK_PLATFORM_LINK_LIBS}
+                         ${PAIMON_BENCHMARK_LINK_TOOLCHAIN}
+                         EXTRA_INCLUDES
+                         ${CMAKE_SOURCE_DIR})
 endif()
 
 if(PAIMON_BUILD_TESTS)
@@ -73,5 +87,22 @@ if(PAIMON_BUILD_TESTS)
                     ${CMAKE_SOURCE_DIR}
                     STATIC_LINK_LIBS
                     paimon_shared
+                    ${GTEST_LINK_TOOLCHAIN})
+
+    # Guards the format-layer assumptions parquet_format_benchmark.cpp is built on. The benchmark
+    # itself is only compiled under PAIMON_BUILD_BENCHMARKS, which CI does not set, so this test is
+    # what keeps those assumptions covered.
+    add_paimon_test(parquet_format_benchmark_test
+                    SOURCES
+                    parquet_format_benchmark_test.cpp
+                    EXTRA_INCLUDES
+                    ${CMAKE_SOURCE_DIR}
+                    STATIC_LINK_LIBS
+                    arrow
+                    parquet
+                    paimon_shared
+                    ${PAIMON_LOCAL_FILE_SYSTEM_SHARED_LINK_LIBS}
+                    ${PAIMON_PARQUET_FILE_FORMAT_STATIC_LINK_LIBS}
+                    test_utils_static
                     ${GTEST_LINK_TOOLCHAIN})
 endif()
--- a/benchmark/parquet_format_benchmark.cpp
+++ b/benchmark/parquet_format_benchmark.cpp
@@ -1,0 +1,1444 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+// Format-level micro-benchmarks for the Parquet reader and writer. These drive
+// ParquetWriterBuilder / ParquetFileBatchReader directly, so a format-layer change can be
+// attributed without the catalog lookup, split planning, merge/sort and commit that the
+// table-level read_write_benchmark includes. Each case comment says what that case answers.
+//
+// Every axis - type, cardinality, null density, batch size, encoding, selectivity - is swept on
+// its own rather than as a combined matrix, because the point is attributing one change rather
+// than describing a workload.
+//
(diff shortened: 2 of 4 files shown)


lucasfang:
+1

lxy-9602 (approved):
+1

PR #233, opened 2026-08-21
fix(executor): support destruction from worker threads

mrdrivingduck:
## What does this PR do?

While testing the OSS filesystem asynchronous read path, we found a lifecycle issue in `DefaultExecutor`. The issue is not OSS-specific: S3 asynchronous reads use the same ownership pattern and can trigger it as well.

The sequence is:

1. An asynchronous object-store request submits a task that captures a `shared_ptr` to its client.
2. The caller releases its client reference.
3. A worker completes the task and releases the final client reference.
4. The client and its executor are then destroyed on that worker thread.
5. Executor shutdown tries to `join()` the current worker thread.

Without this fix, joining the current thread throws `std::system_error` with `Resource deadlock avoided`. Since this happens during destruction, it can terminate the process.

This PR separates the executor's shared scheduling state from the executor object. Workers retain the shared state rather than accessing the executor through `this`. When destruction happens on a worker, shutdown stops the shared state, joins the other workers, and detaches the current one. The current worker then finishes its task and exits normally.

The PR adds a generic executor regression test and an S3 asynchronous range-read test covering this lifecycle.

### DIFF
--- a/src/paimon/common/executor/executor.cpp
+++ b/src/paimon/common/executor/executor.cpp
@@ -40,23 +40,26 @@ class DefaultExecutor : public Executor {
     uint32_t GetThreadNum() const override;
 
  private:
-    void WorkerThread();
+    struct State {
+        std::queue<std::function<void()>> tasks;
+        std::mutex mutex;
+        std::condition_variable condition;
+        bool stop = false;
+    };
+
+    static void WorkerThread(std::shared_ptr<State> state);
     void ShutdownInternal(bool wait_for_pending_tasks);
 
  private:
     uint32_t thread_count_;
     std::vector<std::thread> workers_;
-    std::queue<std::function<void()>> tasks_;
-    std::mutex queue_mutex_;
-    std::condition_variable condition_;
-    bool stop_ = false;
-    int32_t active_tasks_ = 0;
+    std::shared_ptr<State> state_ = std::make_shared<State>();
 };
 
 DefaultExecutor::DefaultExecutor(uint32_t thread_count) : thread_count_(thread_count) {
     assert(thread_count > 0);
     for (uint32_t i = 0; i < thread_count_; ++i) {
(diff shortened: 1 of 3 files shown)


lucasfang:
Thanks for your contribution. The core fix is correct and effective. Moving the scheduling state into a shared_ptr<State> so that workers never touch this, combined with detach() when the thread being shut down is the current one, is the right approach — both parts are necessary, since detaching without relocating the state would leave the surviving worker touching freed members.

src/paimon/common/executor/executor.cpp
```diff
@@ -66,21 +69,22 @@ uint32_t DefaultExecutor::GetThreadNum() const {
 
 void DefaultExecutor::ShutdownInternal(bool wait_for_pending_tasks) {
     {
-        std::unique_lock<std::mutex> lock(queue_mutex_);
-        if (stop_) {
-            return;
-        }
-        stop_ = true;
+        std::unique_lock<std::mutex> lock(state_->mutex);
+        state_->stop = true;
```
lucasfang, line 76: ShutdownInternal() dropped the early return that previously made shutdown idempotent:
if (stop_) {
    return;
}
Without it, two threads can both observe worker.joinable() == true and call join() on the same std::thread. That is a data race on the thread object, and in practice the second pthread_join() blocks forever on an already-reaped thread id.I reproduced this against the real build by adding a temporary probe test (4-thread executor, two threads calling ShutdownNow() concurrently, 50 iterations).

lucasfang: And please add an extra test

mrdrivingduck: Done.

lucasfang:
+1

lxy-9602 (approved):
+1 Thank you for your patience!

Issue #240, opened 2026-08-24
[Feature] Improve scan performance in hot read paths


gripleaf:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-cpp/issues) and found nothing similar.

### Motivation

Recent profiling of highly concurrent scans has revealed several performance bottlenecks caused by
  repeated operations on Arrow-returned `shared_ptr` objects in hot read loops.

  Two significant cases have been identified:

  1. Manifest readers repeatedly call `StructArray::fields()` while processing individual rows. This
     copies `shared_ptr<Array>` objects and, with GCC 8.3's libstdc++, can introduce substantial lock
     contention through `_Sp_locker`, `pthread_mutex_lock`, and futex waits when multiple workers read
     manifests concurrently.

  2. Avro decoding calls `ArrayBuilder::type()` for every integer and timestamp value. Because this
     method returns `std::shared_ptr<DataType>` by value, concurrent scans repeatedly modify reference
     counts on shared Arrow primitive data types, causing cache-line contention. Profiling showed
     `ArrayBuilder::type()` and shared-pointer release operations accounting for a large proportion of
     samples after the manifest bottleneck was removed.

  This issue tracks the broader effort to identify and eliminate similar shared-pointer operations
  from scan hot paths. The goal is to cache immutable Arrow metadata at an appropriate batch, reader,
  or builder lifetime, while ensuring caches are invalidated whenever the corresponding Arrow object
  tree is replaced.

  The expected outcome is lower synchronization and reference-counting overhead under concurrent
  scans, allowing CPU time to return to actual decoding, memory copying, and buffer management.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

gripleaf:
A follow-up PR has been opened for this issue: #242 

  After profiling the updated implementation, several additional sources of Arrow-side overhead remained.

  This follow-up PR focuses on three areas:

  1. Avoid redundant Arrow array validation

     Arrays produced by controlled Avro builders were validated again after `Finish()`. For nested Struct, List, and Map types, this validation recursively compares data types and computes type fingerprints.

     The PR skips this validation in release builds while retaining it in debug builds for invariant checking.

  2. Reuse `ColumnarRow` within a manifest batch

     The manifest bucket scan previously constructed and destroyed one `ColumnarRow` per row, although the underlying Arrow fields and memory pool remain unchanged throughout the batch.

[...]

PR #242, opened 2026-08-24
perf(read): reduce Arrow read-path overhead

gripleaf:
### Purpose

  Linked issue: #240

  Reduce avoidable CPU and memory-management overhead in the Avro and manifest read paths:

  - Skip the redundant recursive Arrow array validation in release builds while retaining it in debug builds.
  - Reuse a single `ColumnarRow` when scanning rows in the same manifest batch, updating only its row index.
  - Reserve Avro Builder capacity based on the configured batch size.
  - Reserve List and Map child Builder capacity using decoded Avro block sizes, including nested Struct builders.

  These changes reduce repeated type fingerprint calculation, row-view construction and destruction, memory reallocations, and data copying.

  ### Tests

  - `paimon-avro-format-test`: 68 tests passed.
  - `ManifestFileTest.*`: 8 tests passed.
  - `ColumnarRowTest.TestSimple`: passed.
  - Added coverage for:
    - Reusing `ColumnarRow` across different row indexes.
    - Recursive capacity reservation for Struct builders.
    - Correct capacity handling for nested List builders.
  - `pre-commit run --files <changed-files>`: passed.
  - `git diff --check`: passed.

  ### API and Format

  No public API changes under `include/paimon/`.

  No storage format or protocol changes.

  ### Documentation

  No documentation changes. This is an internal performance optimization and does not introduce new user-facing functionality.

  ### Generative AI tooling

### DIFF
--- a/src/paimon/core/manifest/manifest_file.cpp
+++ b/src/paimon/core/manifest/manifest_file.cpp
@@ -93,8 +93,9 @@ Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t buc
         file_name,
         [this, bucket, entries](const std::shared_ptr<arrow::StructArray>& batch) -> Status {
             const arrow::ArrayVector& fields = batch->fields();
+            ColumnarRow row(fields, pool_, /*row_id=*/0);
             for (int64_t i = 0; i < batch->length(); i++) {
-                ColumnarRow row(fields, pool_, i);
+                row.SetRowId(i);
                 PAIMON_RETURN_NOT_OK(ManifestEntrySerializer::ValidateVersion(row.GetInt(0)));
                 if (ManifestEntrySerializer::GetBucket(row) != bucket) {
                     continue;
--- a/src/paimon/format/avro/avro_direct_decoder.cpp
+++ b/src/paimon/format/avro/avro_direct_decoder.cpp
@@ -57,6 +57,20 @@ Status DecodeFieldToBuilder(const ::avro::NodePtr& avro_node,
                             ::avro::Decoder* decoder, arrow::ArrayBuilder* array_builder,
                             AvroDirectDecoder::DecodeContext* ctx);
 
+Status ReserveBuilderCapacityImpl(arrow::ArrayBuilder* array_builder, int64_t capacity) {
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(array_builder->Reserve(capacity));
+    if (array_builder->type()->id() != arrow::Type::STRUCT) {
+        return Status::OK();
+    }
+
+    auto* struct_builder = checked_cast<arrow::StructBuilder*>(array_builder);
+    for (int32_t i = 0; i < struct_builder->num_fields(); ++i) {
+        PAIMON_RETURN_NOT_OK(
+            ReserveBuilderCapacityImpl(struct_builder->field_builder(i), capacity));
+    }
+    return Status::OK();
+}
+
 /// \brief Skip an Avro value based on its schema without decoding
 Status SkipAvroValue(const ::avro::NodePtr& avro_node, ::avro::Decoder* decoder) {
     switch (avro_node->type()) {
@@ -193,6 +207,7 @@ Status DecodeListToBuilder(const ::avro::NodePtr& avro_node, ::avro::Decoder* de
     // Read array block count
     int64_t block_count = decoder->arrayStart();
     while (block_count != 0) {
+        PAIMON_RETURN_NOT_OK(ReserveBuilderCapacityImpl(value_builder, block_count));
         for (int64_t i = 0; i < block_count; ++i) {
             PAIMON_RETURN_NOT_OK(DecodeFieldToBuilder(element_node, /*projection=*/std::nullopt,
                                                       decoder, value_builder, ctx));
@@ -221,6 +236,8 @@ Status DecodeMapToBuilder(const ::avro::NodePtr& avro_node, ::avro::Decoder* dec
         // Read map block count
         int64_t block_count = decoder->mapStart();
         while (block_count != 0) {
+            PAIMON_RETURN_NOT_OK(ReserveBuilderCapacityImpl(key_builder, block_count));
+            PAIMON_RETURN_NOT_OK(ReserveBuilderCapacityImpl(item_builder, block_count));
             for (int64_t i = 0; i < block_count; ++i) {
                 PAIMON_RETURN_NOT_OK(DecodeFieldToBuilder(key_node, /*projection=*/std::nullopt,
                                                           decoder, key_builder, ctx));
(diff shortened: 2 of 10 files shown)


gripleaf:
### Flame graph comparison

  The following flame graphs were collected using the same workload and profiling configuration.

Before optimization:
image

After optimization:
image

src/paimon/format/avro/avro_direct_decoder.h
```diff
@@ -81,6 +81,12 @@ class AvroDirectDecoder {
                                       const std::optional<std::set<size_t>>& projection,
                                       ::avro::Decoder* decoder, arrow::ArrayBuilder* array_builder,
                                       DecodeContext* ctx);
+
+    /// Reserve slots for a builder and any struct children with the same cardinality.
+    /// @param array_builder Builder to reserve.
+    /// @param capacity Number of additional values to append.
+    /// @return Status::OK if all reservations succeed.
+    static Status ReserveBuilderCapacity(arrow::ArrayBuilder* array_builder, int64_t capacity);
 };
```
lxy-9602, line 90: Could we move the output parameter to the end?

src/paimon/format/avro/avro_file_batch_reader.cpp
```diff
@@ -123,7 +127,9 @@ Result<BatchReader::ReadBatch> AvroFileBatchReader::NextBatch() {
         }
         PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> array,
                                           array_builder_->Finish());
+#ifndef NDEBUG
         PAIMON_RETURN_NOT_OK_FROM_ARROW(array->Validate());
+#endif
```
lxy-9602, line 133: This looks like the first use of `#ifndef NDEBUG` in the production code. Could we use `assert` or a similar approach instead?

src/paimon/format/avro/avro_direct_decoder.cpp
```diff
@@ -57,6 +57,20 @@ Status DecodeFieldToBuilder(const ::avro::NodePtr& avro_node,
...
+        return Status::OK();
+    }
+
+    auto* struct_builder = checked_cast<arrow::StructBuilder*>(array_builder);
+    for (int32_t i = 0; i < struct_builder->num_fields(); ++i) {
+        PAIMON_RETURN_NOT_OK(
+            ReserveBuilderCapacityImpl(struct_builder->field_builder(i), capacity));
+    }
+    return Status::OK();
+}
```
lxy-9602, line 72: I didn’t notice an obvious allocation-related hotspot under ArrayBuilder::Append() in the flame graphs. Have we run any ablation tests to verify whether pre-reserving capacity actually improves scan latency?

gripleaf: before:
image

after:
image

image

> ReserveRowAlignedBuilderCapacity -> ReserveBuilderCapacityImpl

CPU Profile Comparison:

image

  The optimized implementation reserves builder capacity before decoding. This converts repeated reallocations into a small one-time allocation and reduces the major resizing and reallocation hotspots by approximately 70%.
  The profiles clearly demonstrate reduced memory-management CPU overhead.

src/paimon/format/avro/avro_file_batch_reader.cpp
```diff
@@ -123,7 +128,7 @@ Result<BatchReader::ReadBatch> AvroFileBatchReader::NextBatch() {
         }
         PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Array> array,
                                           array_builder_->Finish());
-        PAIMON_RETURN_NOT_OK_FROM_ARROW(array->Validate());
+        assert(array->Validate().ok());
```
zjw1111: One-Sentence Summary: Please revert to the #ifndef pattern. I've already discussed this with @lxy-9602 .

-----

Moving this into `assert` means that under `NDEBUG` the whole expression is elided and `Validate()` is never called at all. `CMakeLists.txt:37-42` defaults `CMAKE_BUILD_TYPE` to `Release`, and `cmake_modules/SetupCxxFlags.cmake:383-384` compiles Release with `-DNDEBUG`, so the check disappears from the default build. In debug builds the failure mode also changes from a recoverable `Status::Invalid` to `abort()`.

Two things make that a bit riskier than it looks:

- Arrow's `ExportArray`/`ImportArray` do not run `Array::Validate()` themselves, so this was the only structural check on the avro -> Arrow -> C Data Interface path. Downstream, `ColumnarRow::GetXxx()` indexes by `row_id` without bounds checks.
- The sibling readers still keep it: `parquet_file_batch_reader.cpp:619` and `orc_adapter.cpp:943` both use `PAIMON_RETURN_NOT_OK_FROM_ARROW(array->Validate())`, so avro becomes the only exception.

Also, the cost here scales with the number of schema nodes rather than the row count (element-wise scanning only happens in `ValidateFull`), and it runs once per batch (`read.batch-size` defaults to 1024) — so the saving may be smaller than expected.

Would it be possible to either keep the `Status`-returning check, or, if profiling shows it really is a hotspot, turn it into an explicit debug-only block with a short comment and apply the same treatment to the parquet/orc readers so the three stay consistent? If you already have measurements for this, adding them to the PR description would help.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -93,8 +93,9 @@ Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t buc
         file_name,
         [this, bucket, entries](const std::shared_ptr<arrow::StructArray>& batch) -> Status {
             const arrow::ArrayVector& fields = batch->fields();
+            ColumnarRow row(fields, pool_, /*row_id=*/0);
```
zjw1111, line 96: This reuse looks correct to me — `GetRow`/`GetArray`/`GetMap` snapshot `row_id_` at call time, nothing on this path calls `SetRowKind`, and it matches Java's `ColumnarRow.setRowId`.

One gap though: the exact same pattern lives in the base class at `src/paimon/core/utils/objects_file.h:139`, where `ObjectsFile<T>::Read()` still constructs a `ColumnarRow` per row inside its callback. `ManifestFile` is an `ObjectsFile<ManifestEntry>` and `ReadBucketEntries` is its narrower bucket-filtered variant, while `ObjectsFile<T>::Read` is the shared path for full manifest reads, manifest lists and index manifests (`file_store_scan.cpp:266` and `:491`, `manifest_file_merger.cpp:178`, `commit_scanner.cpp:122`, `expire_snapshots.cpp:301`). Those two are the only per-row `ColumnarRow` constructions in non-test code.

Could you apply the same one-line change there as well? Otherwise the broader read path misses the optimization and the two sibling loops diverge.

zjw1111 (commented):
Thanks for working on the read-path overhead here — the `ColumnarRow` reuse and the builder capacity reservation both look like the right direction, and I verified the row-reuse change is safe.

Two points I'd like to raise before this lands, left inline.

src/paimon/core/manifest/manifest_file.cpp
gripleaf: You are right that Validate() is O(k) rather than row-wise. However, our workload scans thousands of small manifest files. Since batching does not cross file boundaries and most files produce only one underfilled batch, Validate() effectively runs once per file.

Each file also creates new Arrow DataType objects, so type fingerprints are recomputed instead of reused. This is why Validate → TypeEquals → LoadFingerprint accounts for about 25.9% of total CPU in the profile, despite Validate() not scanning every row.

gripleaf: Avro can skip Validate() because the arrays are built entirely through Arrow Builders from a controlled schema, without manually reconstructing ArrayData.

  ORC and Parquet are different because both paths may manually transform or reconstruct Arrow arrays. In those cases, Validate() is still useful for catching invalid Arrow invariants before they cause an internal abort, so we should keep it there. @zjw1111

zjw1111 (approved):
+1

lxy-9602 (approved):
+1

PR #243, opened 2026-08-24
feat(reader): support late materialization with probe/payload two-phase reads

zhf999:
This PR is inspired by #196, thanks @gripleaf for the initial work.

### Purpose

This PR introduces a **late-materializing file batch reader** (`LateMaterializingFileBatchReader`)
that performs probe/payload two-phase reads when a predicate is pushed down through
`SetReadSchema`. Without a predicate the reader degrades to a plain passthrough, so it is safe to
install unconditionally.

**Motivation.** In a standard read, all projected columns are materialized together, even though a
pushed-down predicate may filter out the vast majority of rows. Late materialization splits the
read into two passes:

1. **Probe pass** — read only the predicate columns, evaluate the predicate batch by batch, and
   build a `matched_bitmap_` of surviving rows.
2. **Payload pass** — read the remaining (payload) columns, compacted to only the matched rows,
   then assemble the probe and payload columns into a single struct array in `full_schema_` field
   order.

This avoids decoding and holding the wide payload columns for rows that will be discarded.

**Architecture.** The reader is installed **below the prefetch layer** by wrapping the
format-specific `ReaderBuilder` with a `LateMaterializingReaderBuilder`
(`AbstractSplitRead::CreateFileBatchReader`). Each parallel reader under the prefetch layer thus
gets its own late-materializing wrapper. The reader implements the full
`PrefetchFileBatchReader` interface, forwarding schema, row-count, seek, read-range, and metric
queries to the inner reader while interposing the two-phase logic in `SetReadSchema` /
`NextBatch`.

The feature is controlled by `ReadContextBuilder::EnableLateMaterializing(bool)` (default `false`).
When disabled, no wrapping occurs and the read path is unchanged.

### TODOs
- 

**Key files:**

- `src/paimon/common/reader/late_materializing_file_batch_reader.{h,cpp}` — core reader with a
  state machine (`kInit → kProbing → kRunning | kNoLatMat → kEOF`), probe data filtering, payload
  batch reading with bitmap compaction, and full-batch assembly.
- `src/paimon/common/reader/late_materializing_reader_builder.h` — `ReaderBuilder` wrapper that
  installs the late-materializing reader around each format reader produced by the inner builder.
- `src/paimon/core/operation/abstract_split_read.{h,cpp}` — integration: the reader builder is
  wrapped with `LateMaterializingReaderBuilder` inside `CreateFileBatchReader`.
- `src/paimon/format/parquet/parquet_file_batch_reader.{h,cpp}` — fixed read-range loss after
  reentrant `SetReadSchema` by caching and re-applying read ranges.
- `src/paimon/common/utils/arrow/arrow_utils.{h,cpp}` — added `ArrowUtils::NormalizeArrayOffsets()`
  helper for post-slice Arrow array offset normalization.
- `src/paimon/testing/mock/mock_file_batch_reader.h` — enhanced to faithfully emulate a real
  format reader: `SetReadSchema` now resets position and read ranges; `NextBatchWithBitmap`
  honours assigned read ranges; a `ProjectBatch` helper supports column re-selection.
- `src/paimon/CMakeLists.txt` — registers the new source and test files.

**Safety guards:**
- If probe-filter binding fails (e.g. type mismatch between predicate and schema), `SetReadSchema`
  returns `Status::Invalid` to surface the configuration error early.
- If `SetReadSchema` has not been called before `NextBatch`, the reader transitions directly to
  `kNoLatMat`, matching the `FileBatchReader` contract for schema-less reads.

### Tests

**Unit tests.** A dedicated test suite `LateMaterializingFileBatchReaderTest` is added in
`src/paimon/common/reader/late_materializing_file_batch_reader_test.cpp` (16 cases):

| Test case | Verifies |
|---|---|
| `PassThroughWhenNoPredicate` | No predicate → passthrough, single batch returned |
| `PassThroughWhenPayloadEmpty` | All fields are probe fields → passthrough |
| `ContiguousSubsetAcrossBatches` | Contiguous matched rows spanning multiple batches |
| `ScatteredAlternatingMatch` | Non-contiguous alternating matches compacted correctly |
| `MatchedIntersectsSelection` | Predicate bitmap intersected with `selection_bitmap` |
| `EmptyMatchReturnsEof` | Zero matched rows → immediate EOF |
| `SeekToRowRealignsProbeCursor` | `SeekToRow` realigns the probe cursor and state |
| `ReadRangesForwardedAcrossPhases` | Read ranges forwarded to inner reader across phases |
| `ReentrantSetReadSchema` | Repeated `SetReadSchema` resets state cleanly |
| `ForwardsRowCountAndFileSchema` | `GetNumberOfRows` / `GetFileSchema` forwarded |
| `MultiFieldPreservesColumnOrder` | Multi-field schema preserves field order in assembled batch |
| `NestedPayloadColumn` | Nested-type payload column handled correctly |
| `FailsOnPredicateTypeMismatch` | Predicate type mismatch returns error instead of silent fallback |
| `WorksAsInnerOfPrefetchReader` | Works as inner reader of `PrefetchFileBatchReaderImpl` |
| `PrefetchInnerReentrantSetReadSchema` | Reentrant `SetReadSchema` under prefetch |
| `PrefetchInnerParallelReadersWithSeek` | Parallel prefetch readers with seek interleave correctly |

**Integration tests:**
- `test/inte/read_inte_test.cpp` — new LM-enabled read tests with and without predicates.
- `test/inte/scan_and_read_inte_test.cpp` — new LM-enabled scan-and-read tests.
- `test/inte/read_inte_with_index_test.cpp` — `TestBitmapIndexWithLateMaterializing`.
- `src/paimon/core/operation/merge_file_split_read_test.cpp` — `TestReadWithPredicateAndLateMaterializing`.
- Existing tests incompatible with LM (blob table, global index) explicitly disable it via
  `EnableLateMaterializing(false)`.

### Known Limitations / TODOs

1. **ORC `SetReadRanges` is a no-op** — the ORC reader currently ignores read ranges and always
   reads the entire file, causing read amplification in the payload pass.
2. **`PrefetchFileBatchReader` only pre-buffers probe columns** — the prefetch layer currently
   issues `PreBufferRange` only for the probe schema; payload columns are read synchronously on
   demand without pre-buffering.
3. **Predicate evaluation not SIMD-optimized** — row filtering currently uses
   `PredicateFilter::Test`; a future optimization will switch to `arrow::compute::Filter` which
   supports SIMD-accelerated batch filtering.

### API and Format

No storage format or protocol change.

**Public API changes (`include/paimon/`):**

- `PrefetchFileBatchReader::GetNextRowToRead()` — return type changed from `uint64_t` to
  `Result<uint64_t>` to propagate errors from seek operations. All implementors updated
  (Parquet, ORC, PrefetchImpl, LateMat).
- `ReadContext` — added `EnableLateMaterializing() const` getter and
  `ReadContextBuilder::EnableLateMaterializing(bool)` builder method (default: `false`).

**Internal signature changes** in the split-read path:

- `AbstractSplitRead::CreateFileBatchReader` — `const ReaderBuilder*` →
  `std::unique_ptr<ReaderBuilder>` (ownership transfer to allow wrapping).
- `AbstractSplitRead::CreateFieldMappingReader` — same change for the `reader_builder` parameter.

Both are private methods of `AbstractSplitRead`; all call sites within the class are updated in the
same change.

### Documentation

No user-facing documentation change yet. User-facing documentation for the late-materialization
read mode will be added in a follow-up once the feature stabilizes.

### Generative AI tooling

Generated-by: Qoder


src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -152,13 +153,17 @@ Result<std::unique_ptr<ReaderBuilder>> AbstractSplitRead::PrepareReaderBuilder(
 
 Result<std::unique_ptr<FileBatchReader>> AbstractSplitRead::CreateFileBatchReader(
     const std::string& file_format_identifier, const std::string& data_file_path,
-    int64_t data_file_size, const ReaderBuilder* reader_builder) const {
+    int64_t data_file_size, std::unique_ptr<ReaderBuilder> reader_builder) const {
     if (context_->EnablePrefetch() && file_format_identifier != "blob" &&
         file_format_identifier != "avro") {
+        if (context_->EnableLateMaterializing()) {
```
gripleaf: Could we also support late materialization when prefetch is disabled?

Our main read path runs with EnablePrefetch(false). With the current integration, LateMaterializingReaderBuilder is only installed inside the prefetch branch, so those reads always fall back to the regular format reader and never benefit from late materialization.

Late materialization and prefetch should ideally be independent capabilities: the non-prefetch path should also wrap a compatible format reader, or the two-phase orchestration should be moved above the prefetch/non-prefetch split. PR #196 supports both modes by implementing the probe/payload orchestration at the RawFileSplitRead layer.

zhf999: Nice proposal, I will change this logic.

src/paimon/common/reader/late_materializing_file_batch_reader.h
```diff
@@ -0,0 +1,149 @@
...
+        std::unique_ptr<PrefetchFileBatchReader> inner, std::shared_ptr<MemoryPool> pool);
+
+    Result<FileBatchReader::ReadBatch> NextBatch() override;
+
+    std::shared_ptr<Metrics> GetReaderMetrics() const override {
+        return inner_->GetReaderMetrics();
+    };
+
+    void Close() override {
+        inner_->Close();
```
lxy-9602, line 55: `Close()` should also clean up internal data, such as `probe_data_` and similar state.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
```diff
@@ -0,0 +1,352 @@
...
+                return Status::Invalid(
+                    fmt::format("probe cursor {} does not match probe data length {}",
+                                probe_cursor_, probe_data_->length()));
+            }
+            return MakeEofBatch();
+        }
+        auto& [batch, bitmap] = batch_with_bitmap;
+        if (bitmap.IsEmpty()) {
+            ReaderUtils::ReleaseReadBatch(std::move(batch));
+            continue;
```
lxy-9602: In theory, no reader should return a batch with `bitmap.IsEmpty()` to the upper layer. It seems it would be better to return a bad status or simply assert in this case.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
```diff
@@ -0,0 +1,352 @@
...
+                probe_fields.push_back(field);
+            } else {
+                payload_fields.push_back(field);
+            }
+        }
+        // probing only pays off when the predicate fields are a strict subset of the read schema
+        if (!probe_fields.empty() && !payload_fields.empty()) {
+            probe_schema_ = arrow::schema(probe_fields, full_schema_->metadata());
+            payload_schema_ = arrow::schema(payload_fields, full_schema_->metadata());
+            PAIMON_RETURN_NOT_OK(PredicateValidator::ValidatePredicateWithSchema(
```
lxy-9602, line 279: In what scenarios would schema metadata be set here?

src/paimon/common/reader/late_materializing_reader_builder.h
```diff
@@ -0,0 +1,74 @@
...
+    ReaderBuilder* WithReadHints(const std::optional<ReadHints>& hints) override {
+        inner_->WithReadHints(hints);
+        return this;
+    }
+
+    Result<std::unique_ptr<FileBatchReader>> Build(
+        const std::shared_ptr<InputStream>& stream) const override {
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<FileBatchReader> base, inner_->Build(stream));
+        auto* prefetch = dynamic_cast<PrefetchFileBatchReader*>(base.get());
+        if (prefetch == nullptr) {
```
lxy-9602: This feels a bit awkward — does enabling late materialization have to depend on the format supporting prefetch? If the format doesn’t support prefetch, does that mean late materialization can’t be enabled at all?

test/inte/read_inte_test.cpp
```diff
@@ -3109,6 +3194,7 @@ TEST_P(ReadInteTest, TestPkReadSnapshot6WithSchemaEvolutionWithPredicateOnlyPush
     context_builder.SetReadAheadCacheEnabled(param.read_ahead_cache_enabled);
     context_builder.SetPredicate(predicate);
     context_builder.EnablePrefetch(param.enable_prefetch)
+        .EnableLateMaterializing(false)
         .AddOption("test.enable-adaptive-prefetch-strategy",
```
lxy-9602, line 3197: I remember `EnableLateMaterializing` defaults to `false`. Was this change made because some tests were failing, or was it simply adjusted to keep the results stable since the test involves predicates?

test/inte/scan_and_read_inte_test.cpp
```diff
@@ -744,6 +744,47 @@ TEST_P(ScanAndReadInteTest, TestWithPKWithDvBatchScanSnapshot6WithPredicate) {
...
+    ASSERT_OK_AND_ASSIGN(auto predicate, PredicateBuilder::And({not_equal, greater_than}));
+    scan_context_builder.SetPredicate(predicate);
+    ASSERT_OK_AND_ASSIGN(auto scan_context, FinishScanContext(scan_context_builder));
+    ASSERT_OK_AND_ASSIGN(auto table_scan, TableScan::Create(std::move(scan_context)));
+
+    ReadContextBuilder read_context_builder(table_path);
+    AddReadOptionsForPrefetch(&read_context_builder);
+    read_context_builder.SetPredicate(predicate)
+        .EnablePredicateFilter(true)
+        .EnableLateMaterializing(true);
```
lxy-9602: From the integration test, it looks like if `EnablePredicateFilter` is also enabled, the final output will always reflect the filtered result anyway. Could that end up masking the effect of `EnableLateMaterializing`?

src/paimon/common/reader/late_materializing_file_batch_reader_test.cpp
```diff
@@ -0,0 +1,658 @@
...
+                         const std::shared_ptr<Predicate>& predicate,
+                         const std::optional<RoaringBitmap32>& selection) {
+        ::ArrowSchema c_schema;
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportSchema(*schema, &c_schema));
+        return reader->SetReadSchema(&c_schema, predicate, selection);
+    }
+
+    // Collect all output rows as a single concatenated struct array (for schema/nested checks).
+    Result<std::shared_ptr<arrow::StructArray>> CollectStruct(FileBatchReader* reader) {
+        arrow::ArrayVector chunks;
```
lxy-9602: What’s the difference between this and `read_result_collector.h`? It seems like we may be duplicating the same logic here.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+    if (state_ == kNoLatMat) {
+        return inner_->NextBatch();
+    } else if (state_ == kRunning) {
+        return ReadPayloadBatch();
+    } else if (state_ == kEOF) {
+        return MakeEofBatch();
+    }
+    return Status::Invalid("invalid state when calling NextBatch: " + std::to_string(state_));
+}
```
lxy-9602, line 91: Read-ahead cache does not explicitly cover payload columns

`PrefetchFileBatchReaderImpl::Workloop()` initializes the shared read-ahead cache only once. At that point, this reader is configured with `probe_schema_`, so `inner_->PreBufferRange()` only returns byte ranges for the probe columns.

After probing, `SetInnerReadSchema(payload_schema_, ...)` switches to the payload columns, but no additional pre-buffer ranges are collected or registered. `ReadAheadCache` does not dynamically add ranges on a miss, so payload reads fall back to the underlying input stream unless they happen to be covered by a coalesced probe range.

This is especially relevant for Parquet because its native pre-buffering is disabled when the shared read-ahead cache is enabled. Could we initialize the cache with the union of probe and payload byte ranges, or otherwise add the payload ranges when switching phases?

src/paimon/common/reader/late_materializing_file_batch_reader.h
```diff
@@ -0,0 +1,149 @@
...
+
+    Status SeekToRow(uint64_t row_number) override;
+
+    uint64_t GetNextRowToRead() const override {
+        return inner_->GetNextRowToRead();
+    }
+
+    Result<std::vector<std::pair<uint64_t, uint64_t>>> GenReadRanges(
+        bool* need_prefetch) const override {
+        return inner_->GenReadRanges(need_prefetch);
```
lxy-9602: ORC read ranges are not enforced during late materialization

The ORC implementation of `SetReadRanges()` currently returns `OK` without applying the supplied ranges. Normally, the prefetch reader still respects range boundaries by calling `EnsureReaderPosition()` before each batch and slicing the returned batch.

Late materialization bypasses that control point: a single `NextBatch()` call drains the probe reader until EOF. Therefore, each parallel ORC reader may probe from its first assigned range to the end of the file, including ranges assigned to other readers.

The payload phase can amplify this further: `SetReadSchema()` recreates the ORC row reader at row 0, reapplying `SetReadRanges()` has no effect, and ORC does not apply the selection bitmap precisely. The payload reader may consequently scan from the beginning again.

This can cause substantial duplicated I/O and decoding when late materialization and parallel prefetch are enabled together. Could we enforce the assigned ranges in the ORC reader?

test/inte/read_inte_test.cpp
zhf999: In there tests, result would diff from expected because LM reader returns a precise reuslt set. I set `EnableLateMaterializing` to false to avoid test failure in the future when late materialization is set to `true` by default.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
zhf999: In most case of our code, metadata is `nullptr`, `schema` is contructed with metadata here is simply a defensive passthrough.

src/paimon/common/reader/late_materializing_file_batch_reader.h
zhf999: Done.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
zhf999: Done.

test/inte/scan_and_read_inte_test.cpp
zhf999: Done.

src/paimon/common/reader/late_materializing_file_batch_reader_test.cpp
zhf999: Done.

src/paimon/common/reader/late_materializing_file_batch_reader.h
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    Status SetReadRanges(const std::vector<std::pair<uint64_t, uint64_t>>& read_ranges) override;
+
+    Result<std::vector<std::pair<uint64_t, uint64_t>>> PreBufferRange() override {
+        // TODO(zhouhongfeng.zhf): PrebufferRange (called by PrefetchFileBatchReader) only read the
```
gripleaf, line 95: In point-lookup scenarios, especially when the payload is large, pre-buffering payload columns before the probe result is available may introduce significant read amplification, because most prefetched payload data may never be used. I suggest keeping probe-only pre-buffering as the default behavior. 

If payload pre-buffering is introduced later, it would be better to make it configurable so users can choose between minimizing I/O and reducing payload-read latency based on their workload.

src/paimon/common/reader/late_materializing_file_batch_reader_test.cpp
```diff
@@ -0,0 +1,647 @@
...
+    ASSERT_OK_AND_ASSIGN(
+        auto reader, LateMaterializingFileBatchReader::Create(std::move(mock), GetDefaultPool()));
+    // probe = {k}; payload = {arr (list<int64>), tag}
+    auto predicate =
+        PredicateBuilder::GreaterOrEqual(/*field_index=*/0, "k", FieldType::BIGINT, Literal(5l));
+    ASSERT_OK(SetReadSchema(reader.get(), arrow::schema(type->fields()), predicate, std::nullopt));
+
+    ASSERT_OK_AND_ASSIGN(std::shared_ptr<arrow::StructArray> result, CollectStruct(reader.get()));
+    ASSERT_TRUE(result);
+    ASSERT_EQ(result->length(), 3);  // k = 5,6,7
```
lxy-9602, line 501: How about just writing the JSON result by hand? It’s only three lines anyway.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
zhf999: Remain it for future work.

src/paimon/common/reader/late_materializing_file_batch_reader.h
zhf999: Remain it for future work.

src/paimon/common/reader/late_materializing_file_batch_reader_test.cpp
zhf999: The result record values are validated in later loop.

lxy-9602 (approved):
+1

PR #244, opened 2026-08-24
feat(rest): support DLF authentication

JingsongLi:
### Purpose

Linked issue: #164 

Add Java-compatible DLF authentication to the REST Catalog:

- Support DLF4-HMAC-SHA256 and DlfNext OpenAPI HMAC-SHA1 request signing.
- Support static AK/SK/STS credentials, refreshable local token files, and ECS RAM role credentials.
- Match Java credential precedence, endpoint-based signer selection, token parsing, and one-hour refresh window.
- Disable automatic redirects for DLF-signed requests to prevent STS header leakage and invalid redirected signatures.
- Apply the Java-compatible three-minute timeout to ECS metadata requests and preserve explicit empty-role direct token URL semantics.

### Tests

- `cmake --build build-dlf-tdd --target paimon-rest-test -j2`
- `paimon-rest-test --gtest_filter=Dlf*` (13 tests passed)
- 31 non-mock REST/DLF regression tests passed
- `pre-commit run --files <changed files>` (all hooks passed)

The full `unittest` build is blocked on this macOS environment by an existing `literal_converter_test.cpp` incompatibility with the SDK `INFINITY`/`NAN` macros. The redirect transport regression test compiles, but local completion is blocked by the existing macOS `MockRestServer::Stop()` accept-thread issue; Linux CI can execute it.

### API and Format

Adds public `CatalogOptions` keys for DLF REST authentication. No storage-format or REST protocol changes.

### Documentation

Documents the optional OpenSSL build dependency, DLF configuration, credential sources, signing algorithms, and refresh behavior.

### Generative AI tooling

Generated-by: Codex (GPT-5)


leaves12138 (commented):
I found one endpoint-normalization issue in the DLF OpenAPI path.

src/paimon/rest/dlf_auth.cpp
```diff
@@ -0,0 +1,807 @@
...
+}
+
+std::string DlfAuthProvider::ParseSigningAlgorithmFromUri(const std::string& uri) {
+    std::string lower_uri = StringUtils::ToLowerCase(uri);
+    return lower_uri.find("dlfnext") == std::string::npos ? DlfDefaultSigner::kIdentifier
+                                                          : DlfOpenApiSigner::kIdentifier;
+}
+
+Result<std::string> DlfAuthProvider::ExtractHost(const std::string& uri) {
+    std::string host = uri;
```
leaves12138: `RestHttpClient::NormalizeUri` trims surrounding whitespace before transport, but this helper reads the raw `CatalogOptions::URI`. For a supported value such as `" https://dlfnext.cn-hangzhou.aliyuncs.com "`, this returns `" https:"`; OpenAPI signing then sends and signs that invalid `Host`. Please normalize/trim the URI here (or pass the normalized URI into the auth provider) and add a regression test so DLF uses the same endpoint normalization as the transport.

JingsongLi: Fixed in 6920897. `DlfAuthProvider::ExtractHost` now reuses `RestHttpClient::NormalizeUri`, so the signing host follows the same trimming, trailing-slash, and scheme normalization as transport. Added `DlfAuthProviderTest.NormalizesSigningHostLikeTransport`; verified DLF 14/14 and the non-mock REST/DLF set 32/32.

leaves12138 (dismissed):
The endpoint normalization issue is fixed in 6920897, and the regression test covers the transport/signing consistency. I rechecked the incremental change and the original DLF authentication flow. LGTM.

src/paimon/rest/dlf_auth_test.cpp
```diff
@@ -0,0 +1,431 @@
...
+    ASSERT_EQ("UNSIGNED-PAYLOAD", headers.at("x-dlf-content-sha256"));
+    ASSERT_EQ("v1", headers.at("x-dlf-version"));
+    ASSERT_EQ("securityToken", headers.at("x-dlf-security-token"));
+
+    ASSERT_OK_AND_ASSIGN(std::string authorization,
+                         signer.Authorization(parameter, token, "unused", headers));
+    ASSERT_EQ(
+        "DLF4-HMAC-SHA256 Credential=YourAccessKeyId/20250416/cn-beijing/"
+        "DlfNext/aliyun_v4_request,Signature="
+        "22594f8bbb8bb0ec296ced6003b7ffdf7022a8ca3815da5b53090daa11a06558",
```
SteNicholas, line 162: Please add Java's golden vector so cross-language compatibility is actually pinned. The expected strings here are newly minted for this PR. I verified out-of-band that they are consistent with Java's algorithm, but nothing in the suite locks that: a canonicalization regression (signed-header set, the empty-query newline, parameter trimming) would keep these tests green while every real request started coming back 403.

Porting `DLFAuthSignatureTest#testGetAuthorization` verbatim would fix that, and it passes against this implementation as-is:

- method `POST`, path `/v1/paimon/databases`
- parameters `k1=v1`, `k2=v2`
- body `{"name":"database","options":{"a":"b"}}`
- region `cn-hangzhou`, `x-dlf-date` `20231203T121212Z`
- AK `access-key-id` / `access-key-secret`, security token `securityToken`

expected:

```
DLF4-HMAC-SHA256 Credential=access-key-id/20231203/cn-hangzhou/DlfNext/aliyun_v4_request,Signature=c72caf1d40b55b1905d891ee3e3de48a2f8bebefa7e39e4f277acc93c269c5e3
```

A matching golden case for `DlfOpenApiSigner` would be good too, though Java has no published vector for it, so the one here is fine as the reference.

src/paimon/rest/dlf_auth.cpp
```diff
@@ -0,0 +1,808 @@
...
+    Result<HttpResponse> response =
+        http_client_->Execute(request, [&body](const char* data, int64_t size) {
+            if (size < 0 || body.size() + static_cast<size_t>(size) > kMaxTokenResponseBytes) {
+                return Status::Invalid("DLF ECS metadata response is too large");
+            }
+            body.append(data, static_cast<size_t>(size));
+            return Status::OK();
+        });
+    if (!response.ok()) {
+        return Status::IOError("failed to request DLF credentials from ECS metadata service");
```
SteNicholas: The underlying transport error is dropped, which makes ECS failures very hard to diagnose. DNS failure, connection refused, TLS error and the three-minute timeout all collapse into this one message, so an operator on an ECS instance has no way to tell "no RAM role attached" from "metadata service unreachable" from "timed out". The secret in this exchange is the *response body*, not the URL or libcurl's error string — the metadata endpoint is a fixed link-local address, and `Get()` already refuses to echo the body. Java keeps the message here (`DLFECSTokenLoader.getRole` wraps it as `"get role failed, error : " + e.getMessage()`). Suggest propagating the failure detail, e.g. `return Status::IOError("failed to request DLF credentials from ECS metadata service: ", response.status().message());`

src/paimon/rest/dlf_auth.h
```diff
@@ -0,0 +1,210 @@
...
+    static Result<std::unique_ptr<DlfAuthProvider>> FromTokenLoader(
+        std::unique_ptr<DlfTokenLoader> token_loader, const std::string& uri,
+        const std::string& region, const std::string& signing_algorithm,
+        Clock clock = std::chrono::system_clock::now);
+
+    Result<std::map<std::string, std::string>> MergeAuthHeader(
+        const std::map<std::string, std::string>& base_header,
+        const RestAuthParameter& parameter) const override;
+
+    bool AllowsRedirects() const override {
```
SteNicholas, line 185: A 3xx from a DLF endpoint now surfaces as a misleading error. With redirects disabled, a `302` is no longer followed, so it reaches `RestApi::ErrorToStatus`, misses every `case` and falls through to `default`, producing:

```
rest request failed with code 302: unparsable error response body (http status 302)
```

That reads like a malformed server response, when the actual cause is that we deliberately refused to follow a redirect. Since this PR is what introduces the non-following behaviour, it would be good to add an explicit branch in `ErrorToStatus` (or a check on the caller side) for a 3xx when the provider reports `AllowsRedirects() == false`, saying something like "the endpoint returned a redirect, which is not followed for signed requests".

The behaviour itself is right — just the diagnostics need to catch up with it.

src/paimon/rest/dlf_auth.cpp
```diff
@@ -0,0 +1,808 @@
...
+    if (stream.fail() || stream.peek() != std::char_traits<char>::eof()) {
+        return Status::Invalid("invalid DLF token expiration");
+    }
+    int32_t year = utc.tm_year;
+    int32_t month = utc.tm_mon;
+    int32_t day = utc.tm_mday;
+    int32_t hour = utc.tm_hour;
+    int32_t minute = utc.tm_min;
+    int32_t second = utc.tm_sec;
+    std::time_t timestamp = timegm(&utc);
```
SteNicholas, line 174: The `_WIN32` branches in this file are dead code and cannot work as written. `timegm` is called unconditionally here, but it does not exist on MSVC (the equivalent is `_mkgmtime`). So the `#if defined(_WIN32)` / `gmtime_s` branches at L128-136 and L184-192 cannot actually make this file compile on Windows — they only give the impression that Windows is handled. Meanwhile `docs/source/building.rst` states "Windows is not supported for now", and `include/paimon/visibility.h` is the only other `_WIN32` user in the tree, so this is not an established pattern here either. Suggest dropping the `_WIN32` branches and just using `gmtime_r` + `timegm`, consistent with the rest of the repository.

SteNicholas (commented):
@JingsongLi, I cross-checked this against the Java implementation in `apache/paimon` (`paimon-api/.../rest/auth/DLF*.java`). The signing math is correct, and I verified that independently rather than by reading: I re-implemented Java's `DLFDefaultSigner` / `DLFOpenApiSigner` from scratch and it reproduces both Java's golden signature from `DLFAuthSignatureTest` (`c72caf1d40b55b1905d891ee3e3de48a2f8bebefa7e39e4f277acc93c269c5e3`) and all three expected strings in `dlf_auth_test.cpp`. Canonical-request newline handling for empty query strings, signed-header ordering, the four-stage `aliyun_v4` key derivation, and the ROA `x-acs-*` canonicalization all match. Other parity checks that came out clean: the nine option keys match `RESTCatalogOptions` exactly; credential precedence (`token-loader` > `token-path` > AK/SK) matches `DLFAuthProviderFactory`; the one-hour refresh window matches `RESTApi.TOKEN_EXPIRATION_SAFE_TIME_MILLIS`; the explicit-empty-ECS-role semantics are preserved. `dlf.access-key-secret` and `dlf.security-token` are already fully masked by the existing `SensitiveConfigUtils` markers, and `dlf.signing-algorithm` correctly is not.

src/paimon/rest/dlf_auth_test.cpp
JingsongLi: Fixed in 440ae62. Added `DlfDefaultSignerTest.MatchesJavaGoldenAuthorization`, mirroring Java `DLFAuthSignatureTest#testGetAuthorization` with the exact method/path/parameters/body/time/credentials and pinning signature `c72caf1d40b55b1905d891ee3e3de48a2f8bebefa7e39e4f277acc93c269c5e3`.

src/paimon/rest/dlf_auth.cpp
JingsongLi: Fixed in 440ae62. `DlfEcsTokenLoader::Get` now preserves the underlying transport status message while still excluding the response body. Added a regression test that verifies `connection refused` remains visible with the DLF ECS context.

src/paimon/rest/dlf_auth.h
JingsongLi: Fixed in 440ae62. `RestApi` now passes the actual `follow_redirects` decision into `ErrorToStatus`; a 3xx received with redirects disabled reports that redirect status explicitly and explains that signed requests do not follow it. Added coverage for status 302 and retained `RestErrorDetail`.

src/paimon/rest/dlf_auth.cpp
JingsongLi: Fixed in 440ae62. Removed both non-functional `_WIN32` branches and consistently use `gmtime_r` plus `timegm`, matching the repository supported-platform policy.

SteNicholas (dismissed):
LGTM.

src/paimon/rest/dlf_auth.h
```diff
@@ -0,0 +1,210 @@
...
+
+#include "paimon/common/utils/http_client.h"
+#include "paimon/rest/rest_auth.h"
+
+namespace paimon {
+
+/// Access key credentials used to sign DLF REST requests.
+class DlfToken {
+ public:
+    DlfToken(const std::string& access_key_id, const std::string& access_key_secret,
```
zjw1111, line 40: We avoid default values for function parameters (e.g., `= std::nullopt`); instead, explicitly pass all arguments at the call site, including `std::nullopt`.

cmake_modules/arrow.diff
```diff
@@ -15,6 +15,13 @@ diff --git a/cpp/cmake_modules/ThirdpartyToolchain.cmake b/cpp/cmake_modules/Thi
 index 8cb3ec83f5..0765df8fa8 100644
 --- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
 +++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
+@@ -814,5 +814,6 @@ if(DEFINED ENV{ARROW_THRIFT_URL})
+   set(THRIFT_SOURCE_URL "$ENV{ARROW_THRIFT_URL}")
+ else()
+   set_urls(THRIFT_SOURCE_URL
++           "https://archive.apache.org/dist/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
```
zjw1111, line 22: Just curious, why were the URLs in these two patches changed?

CMakeLists.txt
```diff
@@ -79,6 +80,9 @@ endif()
 if(PAIMON_ENABLE_S3 OR PAIMON_ENABLE_REST)
     find_package(CURL REQUIRED)
 endif()
+if(PAIMON_ENABLE_REST)
+    find_package(OpenSSL REQUIRED)
```
zjw1111:
```suggestion
    find_package(OpenSSL 1.1.0 REQUIRED)
```

src/paimon/rest/dlf_auth.cpp
```diff
@@ -0,0 +1,797 @@
...
+}
+
+Result<std::string> RequiredNonEmptyOption(const std::map<std::string, std::string>& options,
+                                           const std::string& key) {
+    std::optional<std::string> value = FindOption(options, key);
+    if (!value || value->empty()) {
+        return Status::Invalid(fmt::format("option '{}' must be configured for DLF auth", key));
+    }
+    return value.value();
+}
```
zjw1111: Could you reuse with `src/paimon/common/utils/options_utils.h`, If the semantics of some of these functions still fall short of the requirements (e.g., RequiredNonEmptyOption), you can extract new utility functions.

src/paimon/rest/dlf_auth.h
JingsongLi: Fixed in c3ad6d4. Removed the default arguments from all newly added DLF auth APIs and updated every call site to pass `std::nullopt`, retry settings, and the clock explicitly.

cmake_modules/arrow.diff
JingsongLi: These URL changes fixed the dependency-download failures seen in the first CI run: ORC format 1.0.0 is no longer available from the active Apache mirror selected by the dynamic closer, so the build failed while downloading it. The Apache Archive is the canonical location for that retired release. The Arrow patch only prepends the Archive URL as a fallback for the older Thrift artifact and retains the existing mirror URLs. After bb1f37d, the affected CI builds passed.

CMakeLists.txt
JingsongLi: Fixed in c3ad6d4. `find_package` now explicitly requires OpenSSL 1.1.0; the EVP APIs used by the DLF signers are available from that version.

src/paimon/rest/dlf_auth.cpp
JingsongLi: Fixed in c3ad6d4. DLF option parsing now reuses `OptionsUtils`. I added `GetOptionalValueFromMap` because DLF must distinguish a missing option from an explicitly configured empty string, notably for the ECS role-name semantics, and added focused coverage for present, missing, empty, and invalid values.

zjw1111 (approved):
+1

lxy-9602 (approved):
+1

PR #245, opened 2026-08-25
feat(pk-index): maintain BTree indexes during compaction

### LINKED ISSUE #291: [Feature] Maintain source-backed primary-key BTree indexes during compaction
### Search before asking

### Motivation

#192 and #194 added the source-backed primary-key BTree read path. Paimon C++ writers still need the corresponding maintenance path: after compaction changes the active source files of a data level, a missing or stale payload leaves that level uncovered and queries fall back to normal file scans.

Paimon C++ should maintain these payloads during fixed-bucket primary-key writes and compaction, using the existing Java-compatible source metadata, BTree payload format, and index manifests.

### Solution

Add the source-backed primary-key BTree maintenance lifecycle for fixed-bucket primary-key tables:

1. Validate the Java-equivalent table and index prerequisites.
2. Restore committed source-backed payload metadata into bucket writers without mixing Data Evolution payloads.
3. Build one payload per indexed field and positive data level from physical source rows, then commit matching index additions and deletions in the same snapshot as the data changes.
4. Reconcile missing, stale, duplicate, replaced, removed-definition, and empty-level payloads during compaction.
5. Isolate build failures to the affected field and level so reads safely fall back to normal scans and a later maintenance attempt can rebuild the payload.
6. Retain live index files during snapshot expiration and orphan cleanup, including tag and branch safety and external-file deletion retries.

Reuse the existing storage formats and internal reader, writer, sort-buffer, path, manifest, and commit abstractions. Do not introduce a new index family or storage protocol.

The implementation is in #245. It keeps maintenance synchronous; Java asynchronous scheduling, manual rebuild actions, realtime writers, and postpone-bucket writers remain outside this scope.

### Anything else?

This is a maintenance-path follow-up to the read-path work in #192 and #194. It ports an existing Java capability, so no separate PIP is proposed.

wangyong9999:
### Purpose

Closes #291

This completes the Java-compatible source-backed primary-key BTree index lifecycle for fixed-bucket primary-key tables. #194 added the read path; this PR adds the missing maintenance path:

- validate the Java-equivalent table/index prerequisites and BTree options;
- restore committed source files and primary-key source-backed payload metadata into each bucket writer, without mixing Java data-evolution (`DEIX`) payloads;
- read physical source rows without deletion-vector filtering, sort `(value, group row id)` with the existing bounded/spill-capable buffers, and build one payload per indexed field and positive data level;
- isolate synchronous build failures to one field and level, retain successful payloads, leave only failed levels uncovered for normal scan fallback, and rebuild them on a later maintenance attempt;
- reconcile missing, stale, duplicate, replaced, removed-definition, and empty-level payloads during compaction, and commit matching index ADD/DELETE entries in the same snapshot as the data changes;
- retain files referenced by current-branch live tags during snapshot expiration, reject expiration while another branch exists, and preserve the retry anchor when an external payload deletion fails; and
- document the synchronous C++ maintenance model and its current boundaries.

The implementation reuses the source metadata, BTree payload format, index manifests, commit messages, sort buffers, path factories, and reader/writer abstractions already in the repository. It adds no storage protocol, index family, background scheduler, retry policy, or manual action.

Java's asynchronous scheduling, retry/fairness policy, and manual rebuild actions are not ported. Realtime and postpone-bucket writers do not automatically build source-backed payloads. Removed-definition cleanup is emitted when a bucket is subsequently opened by write or compaction.

### Tests

- Added coverage for option parsing, failed-increment state preservation, schema validation, deterministic payload construction, field/level build-failure fallback and repair, writer restore and compaction reconciliation, removed definitions, data-evolution payload separation, current-branch tag retention, branch-safe expiration, external-delete retry, orphan cleanup, and Parquet/ORC lifecycle behavior. The ORC case forces dictionary encoding and lazy decoding. Regression coverage also verifies user columns named `_PK_INDEX_ROW_ID` with in-memory and spill sorting, and preservation of tagged data files together with their extra files.
- GCC 8.3 Release `paimon-core-test` and `paimon-primary-key-sorted-index-inte-test` targets built with the project's `-Wall -Werror` configuration (local x86_64 build, reusing the patched bundled dependencies).
- `paimon-core-test`: 2,230/2,230 passed.
- `paimon-primary-key-sorted-index-inte-test`: 12/12 passed.
- `pre-commit run --all-files`: all hooks passed.
- Apache RAT 0.16.1: 0 unknown licenses after excluding the linked-worktree `.git` pointer and local build directory.
- `git diff --check`: passed.

### API and Format

- No new supported table option or query/build API. `Options::PK_CLUSTERING_OVERRIDE` centralizes the existing Java-compatible key, which remains rejected by the unsupported C++ commit path.
- No new storage format or protocol. Payloads continue to use the existing BTree/global-index files, `GlobalIndexMeta`, primary-key source metadata v1, index manifests, and commit-message format.
- The internal restore contract explicitly requests primary-key source-backed payload scanning and excludes Java data-evolution source metadata.
- `OrphanFilesCleaner` supports primary-key tables and retains reachable index manifests, index payloads, and data-file extra files. It still does not enumerate `global-index.external-path`; snapshot expiration deletes retired external payloads by exact committed path and retains metadata for retry on failure.

### Documentation

Updated `docs/source/user_guide/primary_key_global_index.rst` with automatic build and compaction maintenance, build-failure fallback, cleanup ownership, tag/branch expiration boundaries, spill requirements, and the synchronous/realtime/postpone limitations.

### Generative AI tooling

Generated-by: Codex (GPT-5)


src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -185,6 +189,7 @@ Result<int32_t> ExpireSnapshots::ExpireUntil(int64_t earliest_snapshot_id,
             continue;
         }
         PAIMON_ASSIGN_OR_RAISE(Snapshot snapshot, snapshot_manager_->LoadSnapshot(id));
+        PAIMON_RETURN_NOT_OK(CleanUnusedIndexManifest(snapshot.IndexManifest(), &skipping_sets));
```
wangyong9999, line 243: `skipping_sets` only covers the current retained snapshot. Tags (and branches sharing the table root) can still reference an older index manifest, so this can delete a live payload; a tagged DV read then fails or loses its deletion semantics. Add all live tag/branch index manifests and payloads to the retention set, or keep expiration rejected until that traversal exists.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
```diff
@@ -95,24 +139,95 @@ Result<std::shared_ptr<IndexFileMeta>> PkSortedIndexFile::Build(
...
+            }
+            ordinals.push_back(ordinal);
+        }
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            std::shared_ptr<arrow::StructArray> values,
+            arrow::StructArray::Make({record_batch->column(2)}, {field.Name()}));
+        ArrowArray c_array;
+        PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportArray(*values, &c_array));
+        ScopeGuard array_guard([&]() { ArrowArrayRelease(&c_array); });
+        PAIMON_RETURN_NOT_OK(writer->AddBatch(&c_array, std::move(ordinals)));
```
wangyong9999, line 219: These batches do not bound writer memory: `BTreeGlobalIndexWriter` retains every row id for the current key until the key changes. A valid low-cardinality field such as BOOL can therefore hold a whole level across all spill batches, and `Flush()` allocates/copies another encoded buffer, so prepare-commit can OOM even with spill enabled. Spill/chunk each posting list, or enforce a controlled per-key limit.

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -290,6 +295,50 @@ Status ExpireSnapshots::CleanUnusedManifests(const std::string& manifest_list_na
...
+
+    for (const auto& [file_name, file_path] : index_files_to_delete) {
+        skipping_manifest_set->insert(file_name);
+        auto delete_status = fs_->Delete(file_path);
+        // Index payload cleanup is best effort, consistent with data file expiration.
+        (void)delete_status;
+    }
+
+    skipping_manifest_set->insert(index_manifest.value());
+    index_manifest_file_->DeleteQuietly(index_manifest.value());
```
wangyong9999, line 395: The payload delete result is ignored, then the only manifest that records it is removed. Orphan cleanup does not enumerate the global-index external path, so a transient delete failure there becomes permanent and repeated expirations can exhaust external storage. Keep the manifest/retry record until every external payload is deleted or confirmed absent.

src/paimon/core/operation/abstract_file_store_write.cpp
```diff
@@ -208,6 +211,29 @@ Result<std::vector<std::shared_ptr<CommitMessage>>> AbstractFileStoreWrite::Prep
...
+                                PAIMON_ASSIGN_OR_RAISE(
+                                    std::string index_path,
+                                    dv_maintainer_factory_->GetIndexFileHandler()->FilePath(
+                                        partition, bucket, index_file));
+                                [[maybe_unused]] Status cleanup_status =
+                                    options_.GetFileSystem()->Delete(index_path);
+                            }
+                        }
+                    }
+                    return index_status;
```
wangyong9999, line 237: Returning here discards `increment` — which `writer->PrepareCommit()` already drained out of the writer — and also every `CommitMessage` collected for the buckets processed earlier in this loop. A retry gets empty increments from those writers, so data files that were written and would have been committed silently drop out of the commit and stay on disk as orphans.

That window now holds a heavyweight, failure-prone step: a full re-read of a data level plus an external sort, which hard-fails when the sort quota is exhausted (`pk_sorted_index_builder.cpp`, "external-sort quota is exhausted"). A table with a level larger than `write-buffer-size` and no spill directory fails every commit.

The read path is already built to tolerate this — `PkSortedBucketIndexState` documents that levels without a valid group stay uncovered and are scanned normally, and Java builds asynchronously for the same reason. Suggest degrading a maintainer failure to "this level has no payload" (log, delete the payloads built in this attempt, keep committing), or at least running the build before the writer increment is drained.

src/paimon/core/io/key_value_in_memory_record_reader.cpp
```diff
@@ -110,6 +115,21 @@ void KeyValueInMemoryRecordReader::Close() {
 
 Result<std::shared_ptr<arrow::NumericArray<arrow::UInt64Type>>>
 KeyValueInMemoryRecordReader::SortBatch() const {
+    if (sort_comparator_ != nullptr) {
+        std::vector<uint64_t> indices(static_cast<size_t>(value_struct_array_->length()));
+        std::iota(indices.begin(), indices.end(), 0);
+        std::stable_sort(indices.begin(), indices.end(), [&](uint64_t left, uint64_t right) {
+            ColumnarRowRef left_row(value_ctx_, left);
```
wangyong9999, line 122: `ColumnarRowRef`'s constructor takes `std::shared_ptr<ColumnarBatchContext>` by value, so each comparison copies two shared pointers — about 2·n·log₂n atomic refcount round-trips, ~40M for a one-million-row level, none of which does any comparing.

Build the row refs once and index into them:

```cpp
std::vector<ColumnarRowRef> rows;
rows.reserve(indices.size());
for (uint64_t i = 0; i < indices.size(); ++i) rows.emplace_back(value_ctx_, i);
std::stable_sort(indices.begin(), indices.end(), [&](uint64_t left, uint64_t right) {
    return sort_comparator_->CompareTo(rows[left], rows[right]) < 0;
});
```

src/paimon/core/index/pksorted/pk_sorted_data_file_reader.cpp
```diff
@@ -0,0 +1,173 @@
...
+                                   reader->GetPreviousBatchFileRowId(static_cast<uint64_t>(index)));
+            if (physical_position > static_cast<uint64_t>(std::numeric_limits<int64_t>::max()) ||
+                static_cast<int64_t>(physical_position) != rows_read + index) {
+                return Status::Invalid(fmt::format(
+                    "Source file {} returned non-contiguous physical row position {} at row {}.",
+                    file->file_name, physical_position, rows_read + index));
+            }
+            positions.push_back(static_cast<int64_t>(physical_position));
+        }
+        PAIMON_RETURN_NOT_OK(consumer(struct_array, positions));
```
wangyong9999: `positions` never reaches a consumer that uses it. The loop above already asserts `physical_position == rows_read + index` for every row, and `physical_row_count == file->row_count`, `bitmap.Cardinality() == length`, and the final `rows_read == file->row_count` pin the same property from three other directions — so the vector is always exactly `[rows_read, rows_read + length)`. The one consumer (`PkSortedIndexBuilder::Build`) only compares its `size()` and derives the ordinal from its own `rows_buffered + index`.

So this materializes an int64 vector per batch and widens `BatchConsumer` to two parameters to carry a value nobody reads. Either use it for the ordinal (`file_base + positions[i]`, which is what the group-row-id contract suggests) or drop the parameter and keep the contiguity assertion local to the reader.

src/paimon/core/index/pksorted/pk_sorted_index_builder.cpp
```diff
@@ -0,0 +1,276 @@
...
+            return Status::Invalid(
+                "A sorted index can only cover compacted files from one positive data level.");
+        }
+        source_metas.emplace_back(file->file_name, file->row_count);
+    }
+
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<FieldsComparator> unique_comparator,
+                           FieldsComparator::Create({field_}, /*is_ascending_order=*/true));
+    auto comparator = std::shared_ptr<FieldsComparator>(std::move(unique_comparator));
+    DataField row_id_field(std::numeric_limits<int32_t>::max(),
```
wangyong9999: This column carries the same number the sort buffer already assigns. With `last_sequence_number = -1` and rows written in group order, `KeyValue.sequence_number` *is* the group row id — that is exactly what `BuildFromSortedReader` reads back from `record_batch->column(0)`.

The cost of the duplicate: an extra int64 per row, which doubles the sort-buffer footprint for a 4-byte indexed type on the one path that hard-fails when the buffer quota is exhausted, plus three comparators (`comparator`, `sequence_comparator`, `in_memory_comparator`) where one would do.

Dropping it should work directly: `stable_sort` on the value alone keeps insertion order for ties, and `SortMergeReaderWithMinHeap` already falls back to `sequence_number` when the user-defined sequence comparator is null.

src/paimon/core/schema/schema_validation.cpp
```diff
@@ -341,6 +403,71 @@ Status SchemaValidation::ValidateForDeletionVectors(const CoreOptions& options)
...
+
+        auto definition_iter = std::find_if(
+            definitions.Definitions().begin(), definitions.Definitions().end(),
+            [&column](const PrimaryKeyIndexDefinition& definition) {
+                return definition.GetFamily() == PrimaryKeyIndexDefinition::Family::BTREE &&
+                       definition.Column() == column;
+            });
+        if (definition_iter == definitions.Definitions().end()) {
+            return Status::Invalid(
+                fmt::format("Failed to resolve primary-key BTree index column '{}'.", column));
```
wangyong9999: Unreachable: `PrimaryKeyIndexDefinitions::Create` builds BTREE definitions by iterating `schema.Fields()` and testing membership in the btree column list, so any column that passed the "must reference an existing column" check a few lines above necessarily has a definition here.

Also, `kPkClusteringOverride` duplicates the same literal already defined in `file_store_commit_impl.cpp:98`.

src/paimon/core/mergetree/in_memory_sort_buffer.h
```diff
@@ -57,7 +57,8 @@ class InMemorySortBuffer : public SortBuffer {
                        const std::vector<std::string>& user_defined_sequence_fields,
                        bool sequence_fields_ascending,
                        const std::shared_ptr<FieldsComparator>& key_comparator,
-                       uint64_t write_buffer_size, const std::shared_ptr<MemoryPool>& pool);
+                       uint64_t write_buffer_size, const std::shared_ptr<MemoryPool>& pool,
+                       const std::shared_ptr<FieldsComparator>& sort_comparator = nullptr);
```
wangyong9999, line 61: Defaulting to nullptr leaves the main write path on the divergence this parameter exists to fix: `WriteBuffer` sorts its in-memory runs with Arrow `SortIndices`, while `ExternalSortBuffer` merges the spilled runs with `FieldsComparator`. The two orderings must agree or a spilled flush produces a run that is not globally sorted.

The `CompareFloatingPoint` change in this PR makes them definitively opposite for NaN — Arrow places NaN with nulls under `NullPlacement::AtStart`, `FieldsComparator` now places NaN last — so a float/double sort field containing NaN merges incorrectly once spilling kicks in.

Separately, that change also makes `-0.0 < +0.0` where the old comparator returned 0, which changes key equality for every existing table with a float/double key or sequence field. Worth splitting out, or at least calling out in the description.

src/paimon/core/operation/file_system_write_restore.h
```diff
@@ -84,9 +84,19 @@ class FileSystemWriteRestore : public WriteRestore {
...
         }
 
+        std::vector<std::shared_ptr<IndexFileMeta>> primary_key_index_payloads;
+        if (scan_primary_key_indexes) {
+            if (index_file_handler_ == nullptr) {
+                return Status::Invalid("Primary-key index restore requires an index file handler.");
+            }
+            PAIMON_ASSIGN_OR_RAISE(
+                primary_key_index_payloads,
+                index_file_handler_->Scan(snapshot.value(), "btree", partition, bucket));
```
wangyong9999: Third copy of this literal — `BtreeDefs::kIdentifier` and `kBTreeIndexType` (primary_key_index_definitions.cpp:38) already exist, and the value has to stay in sync with `PrimaryKeyIndexDefinition::IndexType()` or the restore scan silently returns nothing and every level gets rebuilt on each commit. A generic `WriteRestore` hard-coding one index family is also the wrong layer; the index type could come from the maintainer factory, which already supplies the `IndexFileHandler`.

While here: `scan_primary_key_indexes = false` is a default argument on a virtual function (declared on both the pure virtual in `write_restore.h` and this override). Defaults bind statically, so a future override with a different default silently changes behavior depending on the static type. Better to make it required and update the two call sites.

src/paimon/core/mergetree/in_memory_sort_buffer.h
wangyong9999: Fixed. Java floating-point ordering is now scoped to the PK BTree in-memory and spill/merge comparators. The generic merge-tree comparator keeps its previous behavior; tests cover NaN and signed zero.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: This matches Java expiration: payload deletion is best-effort, followed by index-manifest cleanup. Retaining the manifest as a retry journal would change that lifecycle contract, so this remains unchanged.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: Fixed for the supported scope. Expiration now uses current-branch live tags to retain their data files, manifests, and index payloads, matching Java. Cross-branch traversal remains unchanged.

src/paimon/core/index/pksorted/pk_sorted_index_builder.cpp
wangyong9999: Kept to match the Java (value, rowId) sort contract and preserve row-id quota accounting. Removing it is an optimization rather than a correctness fix.

src/paimon/core/operation/abstract_file_store_write.cpp
wangyong9999: Fixed. BTree build failures now keep the drained data transition, clean payloads created by that attempt, and leave the level uncovered for fallback and later rebuild. Structural increment errors still propagate. Added a forced-failure regression test.

src/paimon/core/io/key_value_in_memory_record_reader.cpp
wangyong9999: The refcount traffic is real, but prebuilding owning row refs adds O(n) retained row-view memory to the spill path. This is a performance tradeoff rather than a correctness issue, so it remains unchanged here.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
wangyong9999: Confirmed: spill does not bound one posting list. Java uses the same one-key/one-posting-list BTree format; chunking or a limit would change the shared format or behavior, so this parity PR keeps it unchanged.

src/paimon/core/index/pksorted/pk_sorted_data_file_reader.cpp
wangyong9999: Fixed. The reader keeps the contiguous-position validation locally, and the unused position vector and callback parameter are removed.

src/paimon/core/operation/file_system_write_restore.h
wangyong9999: Fixed. Restore now scans source-backed entries by source_meta instead of a hard-coded index type, virtual defaults are removed, and BtreeDefs::kIdentifier is canonical.

src/paimon/core/schema/schema_validation.cpp
wangyong9999: Fixed. The unreachable definition lookup is removed, resolved BTree definitions are validated directly, and pk-clustering-override is centralized under Options.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,261 @@
...
+    next_payloads.reserve(active_payloads_.size() + new_payloads.size());
+    for (const std::shared_ptr<IndexFileMeta>& payload : active_payloads_) {
+        if (deleted_identities.count(PayloadIdentity(payload)) == 0) {
+            next_payloads.push_back(payload);
+        }
+    }
+    next_payloads.insert(next_payloads.end(), new_payloads.begin(), new_payloads.end());
+    active_payloads_ = std::move(next_payloads);
+
+    bool has_compaction_transition =
```
wangyong9999, line 282: Two writers based on the same snapshot can each publish a different payload for the same uncovered level. The second commit does not compare source metadata, and GlobalFileNameCombiner keeps both filenames; the next restore rejects the whole level because it has two candidates and falls back to a full scan. Deduplicate or conflict-check by partition/bucket/type/field/source set during commit retry.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,261 @@
...
+        PAIMON_LOG_WARN(GetLogger(),
+                        "Failed to build primary-key BTree index for column %s; committing data "
+                        "files without new index payloads. %s",
+                        failed_column.c_str(), build_status.ToString().c_str());
+        return Status::OK();
+    }
+
+    std::vector<std::shared_ptr<IndexFileMeta>> next_payloads;
+    next_payloads.reserve(active_payloads_.size() + new_payloads.size());
+    for (const std::shared_ptr<IndexFileMeta>& payload : active_payloads_) {
```
wangyong9999, line 274: active_payloads_ can contain payloads for a field removed from pk-btree.index.columns, but only current fields_ populate deleted_identities, so this loop retains the old payload forever; removing the last definition skips creating the maintainer entirely. Either reject that schema change, or emit DELETEs for restored BTree payloads with no current owner, including the zero-definition case.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: createBranch(branch, tag) copies the tag into the branch, but deleting the source-branch tag does not check branch ownership. The branch snapshot and copied tag still reference the same table-root payload, and main-branch expiration can delete it. Please traverse branch metadata or reject expiration while another branch exists.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: Java parity does not restore a retry path here: this PR explicitly keeps global-index external paths out of orphan cleanup. Once this manifest is deleted, a transient payload-delete failure is no longer discoverable. Keep a durable retry record or retain the manifest until external deletes succeed or are confirmed absent.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
wangyong9999: The shared format does not require accepting process OOM. A hard memory-budget check can return Status before current_row_ids_ and the encoding buffer allocate; the new maintainer fallback then leaves the level uncovered. Please add that bound even if chunking is deferred.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,261 @@
...
+                break;
+            }
+            AddUniquePayload(std::move(build_result).value(), &new_identities, &new_payloads);
+        }
+        if (!build_status.ok()) {
+            break;
+        }
+    }
+
+    if (!build_status.ok()) {
```
wangyong9999: The `break` a few lines above leaves the field loop on the first failure, and this branch then deletes **every** payload built in this round — including the ones that already succeeded for earlier fields and for earlier levels of the failing field.

With a persistent failure that turns into permanent per-commit amplification. The likely persistent failure is exactly the one this path exists for: `PkSortedIndexBuilder::Build` returns Invalid when the sort quota is exhausted, so a level that does not fit the write buffer fails every time. Each commit then re-reads and re-sorts the levels that *can* build, writes their payloads, deletes them again, and the table ends up with no index on any level.

Make the failure granularity (field, level): keep the payloads that built plus the rejected payloads they replace, and leave only the failing level uncovered. That is also what the new class comment promises — "the affected level remains uncovered", not "every level stays uncovered".

src/paimon/core/index/index_file_handler.cpp
```diff
@@ -73,4 +73,22 @@ Result<std::vector<std::shared_ptr<IndexFileMeta>>> IndexFileHandler::Scan(
     return std::vector<std::shared_ptr<IndexFileMeta>>{};
 }
 
+Result<std::vector<std::shared_ptr<IndexFileMeta>>> IndexFileHandler::ScanSourceIndexes(
+    const Snapshot& snapshot, const BinaryRow& partition, int32_t bucket) const {
+    std::function<Result<bool>(const IndexManifestEntry&)> filter =
+        [&partition, bucket](const IndexManifestEntry& entry) -> bool {
+        const std::optional<GlobalIndexMeta>& global_index_meta =
+            entry.index_file->GetGlobalIndexMeta();
```
wangyong9999: `entry.index_file` is dereferenced here without a null check, and so are `expire_snapshots.cpp:357` and `:533` — while `orphan_files_cleaner_impl.cpp:304`, added in the same PR, treats a null `index_file` as an error worth failing the scan for.

Both can't be right. If a deserialized `IndexManifestEntry` can carry a null `index_file`, these three new sites segfault on a manifest the cleaner would have rejected; if it can't, that check is dead weight. Worth picking one and applying it consistently.

src/paimon/common/utils/fields_comparator.cpp
```diff
@@ -128,21 +144,24 @@ Result<FieldsComparator::FieldComparatorFunc> FieldsComparator::CompareField(
...
             return FieldsComparator::FieldComparatorFunc(
-                [field_idx](const InternalRow& lhs, const InternalRow& rhs) -> int32_t {
+                [field_idx, use_java_floating_point_order](const InternalRow& lhs,
+                                                           const InternalRow& rhs) -> int32_t {
                     float lvalue = lhs.GetFloat(field_idx);
                     float rvalue = rhs.GetFloat(field_idx);
-                    return lvalue == rvalue ? 0 : (lvalue < rvalue ? -1 : 1);
+                    return use_java_floating_point_order
+                               ? CompareFloatingPoint(lvalue, rvalue)
+                               : (lvalue == rvalue ? 0 : (lvalue < rvalue ? -1 : 1));
```
wangyong9999: Scoping the Java order to the index build is the right call, but note what the `false` branch is: `compare(NaN, x)` and `compare(x, NaN)` both return 1, so it is not a strict weak ordering, and every `std::sort` / `std::stable_sort` / heap comparator that consumes a default `FieldsComparator` is UB the moment a NaN shows up.

Leaving that behavior alone in this PR is fine, but the TODO that documented it ("nan cannot be compared") was deleted in the previous revision, so the branch now reads as deliberate and correct. Worth restoring a note here.

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -144,7 +156,33 @@ Result<int32_t> ExpireSnapshots::ExpireUntil(int64_t earliest_snapshot_id,
...
+                                "Skip cleaning data files of snapshot #%ld because the data files "
+                                "referenced by tag snapshot #%ld could not be loaded. %s",
+                                id, previous_tag->Id(),
+                                tagged_data_files_result.status().ToString().c_str());
+                tagged_data_files.reset();
+            } else {
+                tagged_data_files = std::move(tagged_data_files_result).value();
+            }
+        }
+        if (previous_tag != nullptr && !tagged_data_files) {
```
wangyong9999, line 192: Two notes on this skip:

- `tagged_data_files` is only recomputed when the tag changes, so one failed read is sticky for every remaining snapshot under the same tag, not just this one. Java retries `tryReadDataFiles` per expiring snapshot and only caches successes.
- The skip covers data-file cleanup only. The second loop still deletes this snapshot's manifests and the snapshot file itself, so the data files that were not cleaned end up unreachable — no snapshot references them and no delta manifest lists them any more. They are recoverable only through orphan cleanup.

The conservative direction is right (better an orphan than a deleted tagged file); worth saying the second part in the warning so the leak is diagnosable.

src/paimon/core/schema/schema_validation.cpp
```diff
@@ -341,6 +402,64 @@ Status SchemaValidation::ValidateForDeletionVectors(const CoreOptions& options)
...
+        bool deletion_vectors_merge_on_read,
+        OptionsUtils::GetValueFromMap<bool>(schema.Options(), kDeletionVectorsMergeOnRead, false));
+    if (deletion_vectors_merge_on_read) {
+        return Status::Invalid(
+            "Primary-key BTree indexes require deletion-vectors.merge-on-read = false.");
+    }
+    PAIMON_ASSIGN_OR_RAISE(bool pk_clustering_override,
+                           OptionsUtils::GetValueFromMap<bool>(
+                               schema.Options(), Options::PK_CLUSTERING_OVERRIDE, false));
+    if (pk_clustering_override) {
```
wangyong9999, line 469: `FileStoreCommitImpl::ValidateCommitOptions` already rejects `pk-clustering-override` for every table (`file_store_commit.cpp:129` runs it on every commit path), so this branch only fires for a schema whose first commit would fail anyway, and the message reads as if the option works when BTree indexes are off. Either drop it or point it at the global restriction.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Confirmed as temporary coverage loss, not result corruption: duplicate candidates invalidate that level and queries fall back to normal scanning. Java has the same source-backed behavior. Commit-time semantic arbitration would add a new conflict policy, so this parity PR leaves it unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Fixed. Restored PK BTree payloads with no current field owner are deleted when the bucket is next opened, including the zero-definition case. Added schema-evolution coverage.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: Fixed. Snapshot expiration now rejects tables while another branch exists, until cross-branch retention is supported.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: Fixed. If an external payload delete fails and the file still exists, expiration returns an error and keeps the manifest/snapshot retry anchor; confirmed absence is accepted. Added retry coverage.

src/paimon/core/index/pksorted/pk_sorted_index_file.cpp
wangyong9999: Confirmed, but neither implementation defines a posting-list memory budget. Reusing the sort-buffer limit would add C++-only rejection semantics, so this parity PR leaves it unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Fixed. Build failures are now isolated to one field and level; successful payloads and matching deletions are retained, and only the failed level remains uncovered.

src/paimon/core/index/index_file_handler.cpp
wangyong9999: `index_file` is non-null by the index-manifest serializer/deserializer contract. Removed the inconsistent dead guard.

src/paimon/common/utils/fields_comparator.cpp
wangyong9999: Restored the note. PK BTree construction continues to opt into the Java floating-point order.

src/paimon/core/operation/expire_snapshots.cpp
wangyong9999: Current Java also builds one skipper per tag per expiration, so the failure is sticky there as well. Updated the warning to state that metadata expiration continues and skipped files may remain for orphan cleanup.

src/paimon/core/schema/schema_validation.cpp
wangyong9999: Updated the message to point to the global C++ commit-path restriction while retaining early schema validation.

wangyong9999:
cc @lxy-9602 @SteNicholas  could you please take a look when convenient? Thanks~

wangyong9999:
Merged the latest main and resolved the conflicts. cc @lxy-9602 @SteNicholas, could you please take a look when convenient? Thanks!

src/paimon/core/operation/expire_snapshots.cpp
```diff
@@ -372,16 +479,94 @@ Status ExpireSnapshots::GetManifestSkippingSet(const std::vector<Snapshot>& reta
...
+        tagged_snapshots.begin(), right, begin_inclusive_id,
+        [](int64_t snapshot_id, const Snapshot& snapshot) { return snapshot_id < snapshot.Id(); });
+    if (left != tagged_snapshots.begin()) {
+        --left;
+    }
+    return std::vector<Snapshot>(left, right);
+}
+
+Result<std::set<std::string>> ExpireSnapshots::GetTaggedDataFiles(
+    const Snapshot& tagged_snapshot) const {
```
lszskye: seems familiar

wangyong9999:
Rebased onto the latest main and resolved the conflicts. cc @lszskye, could you please take a look when you have a chance? Thanks!

include/paimon/defs.h
```diff
@@ -396,6 +396,10 @@ struct PAIMON_EXPORT Options {
     /// a table is never compacted.
     static const char DELETION_VECTORS_ENABLED[];
 
+    /// "pk-clustering-override" - Whether primary-key clustering columns override the primary
```
lszskye: please keep the same comment with Java paimon

lszskye: Why not add a `PkClusteringOverrideEnabled()` api in `CoreOptions`

src/paimon/core/schema/schema_validation.cpp
```diff
@@ -137,6 +144,58 @@ Status ValidatePerLevelOption(
     return Status::OK();
 }
 
+std::vector<std::string> PrimaryKeyBTreeIndexColumns(
+    const std::map<std::string, std::string>& options) {
+    auto iter = options.find(Options::PK_BTREE_INDEX_COLUMNS);
```
lszskye: Also add `GetXXX` api in `CoreOptions` please

lszskye: this function is supposed to be moved to  class `CoreOptions`

include/paimon/defs.h
```diff
@@ -396,6 +396,10 @@ struct PAIMON_EXPORT Options {
...
+    /// keys when clustering data. Default value is false.
+    static const char PK_CLUSTERING_OVERRIDE[];
```
lszskye, line 403: require more check for `PK_CLUSTERING_OVERRIDE`, e.g., 

Image

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,294 @@
...
+            return false;
+        }
+        group_index++;
+    }
+    return true;
+}
+
+}  // namespace
+
+Result<std::shared_ptr<BucketedPrimaryKeyIndexMaintainer::Factory>>
```
lszskye: Factory::Create currently has no fallible operation or error return path. Why use Result

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,294 @@
...
+        return Status::Invalid("Primary-key index commit increment is null.");
+    }
+    auto previous_data_files = active_data_files_;
+    const DataIncrement& data_increment = increment->GetNewFilesIncrement();
+    const CompactIncrement& compact_increment = increment->GetCompactIncrement();
+    PAIMON_RETURN_NOT_OK(ValidateAppendFiles(data_increment.NewFiles()));
+    RemoveDataFiles(compact_increment.CompactBefore(), &active_data_files_);
+    Status update_status = AddSourceFiles(compact_increment.CompactAfter(), &active_data_files_);
+    if (!update_status.ok()) {
+        active_data_files_ = std::move(previous_data_files);
```
lszskye: Instead of mutating active_data_files_ first and rolling it back on failure, could we apply the changes to a temporary copy and assign it back only after AddSourceFiles succeeds? This would make the transactional semantics clearer and guarantee that a failed PrepareCommit leaves the maintainer state unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,294 @@
...
+    bool has_compaction_transition =
+        !compact_increment.CompactBefore().empty() || !compact_increment.CompactAfter().empty();
+    if (has_compaction_transition) {
+        increment->GetCompactIncrement().AddNewIndexFiles(std::move(new_payloads));
+        increment->GetCompactIncrement().AddDeletedIndexFiles(std::move(deleted_payloads));
+    } else {
+        increment->GetNewFilesIncrement().AddNewIndexFiles(std::move(new_payloads));
+        increment->GetNewFilesIncrement().AddDeletedIndexFiles(std::move(deleted_payloads));
+    }
+    return Status::OK();
```
lszskye, line 227: The function is too long. Consider splitting it into smaller, focused functions to improve readability and maintainability.

src/paimon/core/operation/abstract_file_store_write.cpp
```diff
@@ -495,6 +521,8 @@ Result<std::shared_ptr<RestoreFiles>> AbstractFileStoreWrite::ScanExistingFileMe
     std::shared_ptr<IndexFileHandler> index_file_handler;
     if (dv_maintainer_factory_) {
         index_file_handler = dv_maintainer_factory_->GetIndexFileHandler();
+    } else if (primary_key_index_maintainer_factory_) {
+        index_file_handler = primary_key_index_maintainer_factory_->GetIndexFileHandler();
     }
```
lszskye, line 528: Both factories are expected to share the same IndexFileHandler, but the current if/else if logic makes this invariant implicit. Could we pass the shared IndexFileHandler directly to AbstractFileStoreWrite, or at least assert that both factories hold the same instance when they are present? This would make the dependency explicit and avoid silently selecting the wrong handler if the construction logic changes in the future.

src/paimon/core/index/pksorted/pk_sorted_data_file_reader.cpp
```diff
@@ -0,0 +1,170 @@
...
+    ReadContextBuilder builder(root_path);
+    builder.SetReadFieldIds({field_id})
+        .SetOptions(read_options)
+        .WithBranch(branch)
+        .WithFileSystem(options.GetFileSystem())
+        .WithExecutor(executor)
+        .WithMemoryPool(pool)
+        .EnablePrefetch(false)
+        .EnablePredicateFilter(false);
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<ReadContext> read_context, builder.Finish());
```
lszskye: just `PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<ReadContext> read_context, builder.Finish());`

src/paimon/core/index/pksorted/pk_sorted_index_builder.cpp
```diff
@@ -0,0 +1,270 @@
...
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<IndexPathFactory> index_path_factory,
+                           path_factory->CreateIndexFileFactory(partition, bucket));
+    return std::unique_ptr<PkSortedIndexBuilder>(new PkSortedIndexBuilder(
+        partition, bucket, std::move(field), definition,
+        std::shared_ptr<PkSortedDataFileReader>(std::move(data_file_reader)),
+        options.GetFileSystem(), std::shared_ptr<IndexPathFactory>(std::move(index_path_factory)),
+        options, io_manager, enable_multi_thread_spill, pool));
+}
+
+Result<std::shared_ptr<IndexFileMeta>> PkSortedIndexBuilder::Build(
```
lszskye, line 118: The function is too long. Consider splitting it into smaller, focused functions to improve readability and maintainability.

include/paimon/defs.h
wangyong9999: Updated the description to match Java and added `CoreOptions::PkClusteringOverrideEnabled()`. Schema validation now uses the accessor.

src/paimon/core/schema/schema_validation.cpp
wangyong9999: Moved this into `CoreOptions::GetPrimaryKeyBTreeIndexColumns()` and added parsing coverage.

include/paimon/defs.h
wangyong9999: The Java condition allows first-row + deletion vectors when clustering override is enabled. C++ still rejects that mode in the commit path, so I kept the restriction rather than enabling an unsupported combination.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Removed `Result` from `Factory::Create()` and updated its callers. `CreateMaintainer()` remains fallible.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Changes now go into a temporary copy, assigned back only after validation succeeds. Added a regression test verifying that an invalid increment leaves the original state unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Extracted per-field reconciliation into `ReconcileField()`, leaving source updates and commit publication in `PrepareCommit()`.

src/paimon/core/operation/abstract_file_store_write.cpp
wangyong9999: Added an assertion that both factories share the same handler when present.

src/paimon/core/index/pksorted/pk_sorted_data_file_reader.cpp
wangyong9999: Done; both context objects now use the shared-pointer target directly in the macro.

src/paimon/core/index/pksorted/pk_sorted_index_builder.cpp
wangyong9999: Extracted sort-buffer creation and source loading. The build-scoped memory pool and cleanup remain unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.h
```diff
@@ -0,0 +1,144 @@
...
+///
+/// This initial implementation builds synchronously during prepare-commit. Synchronous execution
+/// deliberately omits Java's scheduling and retry policy while preserving the same source-level
+/// reconciliation and payload format. A build failure does not block the data-file commit; readers
+/// fall back to scanning the committed data files until a later maintenance attempt succeeds.
+class BucketedPrimaryKeyIndexMaintainer {
+ public:
+    class Factory {
+     public:
+        static std::shared_ptr<Factory> Create(
```
lszskye: We can remove `Create` here and construct the object directly using the constructor.

wangyong9999: Removed Factory::Create and updated both callers to use std::make_shared. BTree definition filtering and ordering are unchanged.

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
```diff
@@ -0,0 +1,296 @@
...
+        if (file != nullptr) {
+            active->erase(file->file_name);
+        }
+    }
+}
+
+Status AddSourceFiles(const std::vector<std::shared_ptr<DataFileMeta>>& files,
+                      std::map<std::string, std::shared_ptr<DataFileMeta>>* active) {
+    for (const std::shared_ptr<DataFileMeta>& file : files) {
+        if (file == nullptr) {
```
lszskye: Actually, there is no need to check if `DataFileMeta` or `IndexFileMeta` is nullptr

include/paimon/defs.h
lszskye: make sense

src/paimon/core/index/pk/bucketed_primary_key_index_maintainer.cpp
wangyong9999: Removed the redundant metadata null checks from the maintainer and build path. AddSourceFiles now returns void; the state-preservation test uses an invalid append instead of a null file.

lszskye:
+1

lxy-9602 (approved):
+1

Issue #174, opened 2026-08-03
[Feature] Support changelog production


lxy-9602:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-cpp/issues) and found nothing similar.

### Motivation

Java Paimon supports producing changelog files for primary-key tables through the
`changelog-producer` option, including `input`, `lookup`, and `full-compaction`
modes.

Paimon C++ already contains part of the required infrastructure, such as:

- Parsing the `changelog-producer` option.
- Changelog file metadata in `DataIncrement` and `CompactIncrement`.
- Changelog manifest and snapshot commit support.
- Lookup changelog merge functions and changelog compaction rewriters.
- Streaming changelog scans and the `audit_log` and `binlog` system tables.

However, schema validation currently rejects every value except `none`, and the
writer does not produce changelog files. As a result, primary-key tables written
by Paimon C++ cannot provide the same changelog behavior as tables written by
Java Paimon.

Supporting Java-compatible changelog production would enable incremental
downstream consumption and improve interoperability between Java Paimon and
Paimon C++.

### Solution

Implement changelog production for Paimon C++ primary-key table writes, aligned
with Java Paimon semantics and storage format.

The implementation should cover the following producer modes:

- `input`
  - Preserve input row kinds and double-write input changes to changelog files.
  - Attach generated changelog files to the corresponding data increment.

- `lookup`
  - Look up the previous value of each primary key when necessary.
  - Produce the correct `INSERT`, `UPDATE_BEFORE`, `UPDATE_AFTER`, and `DELETE`
    records.
  - Generate changelog files while compacting level-0 data.

- `full-compaction`
  - Produce changelog files during full compaction by comparing the compacted
    result with the previous table state.

The generated changelog files should:

- Use the same file schema, row-kind encoding, metadata, and directory layout as
  Java Paimon.
- Be committed through changelog manifests and referenced by the snapshot.
- Be readable through streaming scans, `audit_log`, and `binlog`.
- Work correctly with supported merge engines, sequence fields, partitions, and
  bucket modes.
- Either support deletion vectors consistently with Java Paimon or reject
  unsupported combinations with a clear validation error.
- Be included in commit metrics, expiration, rollback, and orphan-file cleanup.

The feature can be delivered incrementally, with separate pull requests for each
changelog producer mode.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

PR #248, opened 2026-08-26
feat(core): support input and lookup changelog producers

lszskye:
### Purpose

Linked issue: #174 

Support `input` and `lookup` changelog producers for primary-key tables.

This change:

- Writes raw input changelog records into changelog files while write process.
- Rolls input changelog files by target file size, consistently with Java Paimon.
- Produces lookup changelogs during compaction by comparing level-0 records with existing records.
- Supports lookup changelog with:
  - several merge engines;
  - deletion vectors enabled;
  - `ignore_previous_files`;
  - row deduplication and ignored comparison fields;
  - schema evolution and external data paths.
- Persists keys in first-row lookup SST files so excluded higher-level records can be detected.
- Supports shared shredding when writing input and compaction changelog files.
- Supports configurable changelog file prefix, format, and compression.
- Reads changelog manifests during streaming scans.
- Excludes level-0 files from the initial full scan for lookup changelog tables to avoid duplicate emission.
- Adds order-independent equality comparison for MAP values.
- Keeps `full-compaction` changelog producer unsupported.

### Tests
Added and updated unit and integration coverage for:
  - input and lookup changelog stream reads;
  - insert, update, and delete changelogs;
  - first-row and deletion-vector lookup;
  - row deduplication and ignored fields;
  - schema evolution and external paths;
  - configurable changelog format, compression, and prefix;
  - shared-shredding changelog files;
  - nested row and order-independent MAP equality;
  - changelog manifest scanning.

### API and Format
This change adds public option constants for:
- `changelog-producer.row-deduplicate`
- `changelog-producer.row-deduplicate-ignore-fields`
- `changelog-file.prefix`
- `changelog-file.format`
- `changelog-file.compression`


include/paimon/defs.h
```diff
@@ -378,14 +378,34 @@ struct PAIMON_EXPORT Options {
...
     static const char DELETION_VECTOR_BITMAP64[];
 
-    ///  @note `CHANGELOG_PRODUCER` currently only support `none`
-    ///
     /// "changelog-producer" - Whether to double write to a changelog file. This changelog file
     /// keeps the details of data changes, it can be read directly during stream reads. This can be
     /// applied to tables with primary keys. Values can be "none", "input", "lookup",
     /// "full-compaction". Default value is "none".
+    /// @note C++ Paimon currently supports "none" and "input".
     static const char CHANGELOG_PRODUCER[];
```
lxy-9602, line 394: lookup is also support?

src/paimon/core/table/source/snapshot/changelog_follow_up_scanner.h
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright 2026-present Alibaba Inc.
```
lxy-9602: check license

src/paimon/core/operation/key_value_file_store_scan.cpp
```diff
@@ -176,9 +176,10 @@ Result<bool> KeyValueFileStoreScan::IsValueFilterEnabled() const {
         case ScanMode::ALL:
             return value_filter_force_enabled_;
         case ScanMode::DELTA:
+        case ScanMode::CHANGELOG:
             return false;
```
lxy-9602, line 179: This path uses `ScanMode::CHANGELOG`, but `KeyValueFileStoreScan::IsValueFilterEnabled()` disables value-stat pruning for all changelog scans. Java enables it for `LOOKUP` and `FULL_COMPACTION`. Could we align this behavior, or document it as a known performance gap?

src/paimon/core/io/async_key_value_producer_and_consumer.h
```diff
@@ -48,24 +79,39 @@ class AsyncKeyValueProducerAndConsumer {
...
     }
 
     Result<R> NextBatch();
 
+    Result<AsyncKeyValueResultBatch<R>> NextBatchWithType();
+
     std::shared_ptr<Metrics> GetReaderMetrics() const {
-        return sort_merge_reader_->GetReaderMetrics();
+        return sort_merge_reader_ ? sort_merge_reader_->GetReaderMetrics() : nullptr;
     }
```
lxy-9602, line 126: Otherwise, could we create an empty metrics object here instead? It’d be better to avoid returning nullptr if possible.

src/paimon/core/io/async_key_value_producer_and_consumer.cpp
```diff
@@ -227,7 +292,12 @@ void AsyncKeyValueConsumer<T, R>::CleanUp() {
     if (consumer_future_.valid()) {
         [[maybe_unused]] Status status = consumer_future_.get();
     }
-    key_value_consumer_->CleanUp();
+    if (key_value_consumer_) {
+        key_value_consumer_->CleanUp();
+    }
+    if (changelog_consumer_) {
+        changelog_consumer_->CleanUp();
+    }
```
lxy-9602, line 289: This overload introduces two mutually exclusive internal modes: one uses `sort_merge_reader_` and `batch_size_`, while the other uses `producer_` and leaves the former members inactive. As a result, `ProduceLoop`, `GetReaderMetrics`, and `Close` all need to branch on how the object was constructed. Could we wrap `SortMergeReader` in an `AsyncKeyValueBatchProducer` and keep a single producer abstraction? The reader-specific batch size, metrics, and close behavior could then live in that adapter.

lxy-9602: Could we simplify this pipeline by keeping the responsibilities separated as follows?

- The producer decides which output types are enabled and tags each emitted batch as DATA or CHANGELOG.
- The consumer only converts `KeyValue` rows into `KeyValueBatch` and preserves the tag in the converted result.
- `ChangelogMergeTreeRewriter` uses the preserved tag to route the converted batch to either the compact data writer or the changelog writer.

DATA and CHANGELOG currently use the same `KeyValue` input type, write schema, and `create_consumer` factory, so maintaining separate `key_value_consumer_` and `changelog_consumer_` instances and selecting between them here does not appear necessary. Could we use a single converter and keep the DATA/CHANGELOG distinction only for downstream writer routing? This would also remove the nullable consumer state and the need to synchronize producer output flags with consumer availability.

src/paimon/core/mergetree/compact/changelog_merge_tree_rewriter.cpp
```diff
@@ -17,7 +17,152 @@
...
+    std::function<Result<std::shared_ptr<MergeFunctionWrapper<ChangelogResult>>>()>;
+using KeyComparator = std::function<int32_t(const InternalRow&, const InternalRow&)>;
+using CancellationChecker = std::function<bool()>;
+
+class ChangelogCompactionBatchProducer : public AsyncKeyValueBatchProducer {
+ public:
+    ChangelogCompactionBatchProducer(
+        const std::vector<std::vector<SortedRun>>& sections,
+        std::vector<std::unique_ptr<SortMergeReader>>& reader_holders, int32_t write_batch_size,
+        SortMergeReaderFactory reader_factory,
```
lxy-9602: `std::vector<std::unique_ptr<SortMergeReader>>& reader_holders`? Use `std::vector &&` or just `std::vector`

src/paimon/core/mergetree/compact/first_row_merge_function_wrapper.h
```diff
@@ -62,17 +69,26 @@ class FirstRowMergeFunctionWrapper : public MergeFunctionWrapper<KeyValue> {
...
-        // TODO(xinyu.lxy) support changelog
+        if (value_serializer_) {
+            PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<Bytes> bytes,
+                                   value_serializer_->SerializeToBytes(*result->value));
+            PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<InternalRow> changelog_value,
+                                   value_serializer_->Deserialize(bytes));
+            changelog_result.changelogs.emplace_back(result->value_kind, result->sequence_number,
+                                                     result->level, result->key,
+                                                     std::move(changelog_value));
+        }
```
lxy-9602, line 83: Could we add a TODO here? There seems to be a clear hotspot at the moment, and we’re still doing row-by-row data copying.

src/paimon/core/mergetree/compact/internal_row_equalizer.h
```diff
@@ -0,0 +1,245 @@
...
+#include "paimon/result.h"
+#include "paimon/status.h"
+
+namespace paimon {
+
+/// Creates equality functions for internal rows, including nested values.
+class InternalRowEqualizer {
+ public:
+    using Equalizer = std::function<bool(const InternalRow&, const InternalRow&)>;
+
```
lxy-9602: `std::function<bool(const InternalRow&, const InternalRow&)>` is actually already defined as `FieldComparatorFunc` in `fields_comparator.h`. It looks like we’ve redefined it in several places now.

src/paimon/core/mergetree/compact/internal_row_equalizer.h
```diff
@@ -0,0 +1,245 @@
...
+                return true;
+            });
+    }
+
+ private:
+    using ValueEqualizer =
+        std::function<bool(const DataGetters&, int32_t, const DataGetters&, int32_t)>;
+
+    static bool EqualAt(const DataGetters& lhs, int32_t lhs_pos, const DataGetters& rhs,
+                        int32_t rhs_pos, const ValueEqualizer& equalizer) {
```
lxy-9602, line 78: The Java side also compares `rowkind`. It may not affect the current path, but could we add a comment to clarify that for now?

src/paimon/core/mergetree/compact/lookup_changelog_merge_function_wrapper.h
```diff
@@ -105,30 +119,80 @@ class LookupChangelogMergeFunctionWrapper : public MergeFunctionWrapper<KeyValue
...
-          comparator_(CreateSequenceComparator(user_defined_seq_comparator)) {}
+          comparator_(CreateSequenceComparator(user_defined_seq_comparator)),
+          value_serializer_(std::move(value_serializer)),
+          value_equalizer_(std::move(value_equalizer)) {}
+
+    Result<KeyValue> CloneKeyValue(const KeyValue& from, const RowKind* value_kind) {
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<Bytes> bytes,
+                               value_serializer_->SerializeToBytes(*from.value));
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<InternalRow> value,
+                               value_serializer_->Deserialize(bytes));
```
lxy-9602, line 166: Similarly, could we add a comment here to call out the copy overhead and a TODO for future optimization?

src/paimon/core/mergetree/compact/lookup_changelog_merge_function_wrapper.h
```diff
@@ -105,30 +119,80 @@ class LookupChangelogMergeFunctionWrapper : public MergeFunctionWrapper<KeyValue
...
+        return KeyValue(value_kind, from.sequence_number, from.level, from.key, std::move(value));
```
lxy-9602: Java's `KeyValue.replace` resets the generated changelog record's level to `UNKNOWN_LEVEL`, while `CloneKeyValue` preserves from.level. The current Arrow converter does not serialize this field, so this may not affect files today, but the intermediate KeyValue semantics differ from Java. Could we use `KeyValue::UNKNOWN_LEVEL` here?

src/paimon/core/mergetree/compact/lookup_merge_tree_compact_rewriter.h
```diff
@@ -50,16 +53,19 @@ class LookupMergeTreeCompactRewriter : public ChangelogMergeTreeRewriter {
         return lookup_levels_->Close();
     }
 
-    static std::shared_ptr<MergeFunctionWrapper<KeyValue>> CreateFirstRowMergeFunctionWrapper(
-        std::unique_ptr<FirstRowMergeFunction>&& merge_func, int32_t output_level,
-        LookupLevels<bool>* lookup_levels);
+    static std::shared_ptr<MergeFunctionWrapper<ChangelogResult>>
+    CreateFirstRowMergeFunctionWrapper(std::unique_ptr<FirstRowMergeFunction>&& merge_func,
+                                       int32_t output_level, LookupLevels<bool>* lookup_levels,
+                                       std::unique_ptr<RowCompactedSerializer>&& value_serializer);
```
lxy-9602: How about moving `value_serializer` before `lookup_levels` instead?

src/paimon/core/mergetree/compact/merge_tree_compact_manager_factory_test.cpp
```diff
@@ -377,13 +379,12 @@ TEST_F(MergeTreeCompactManagerFactoryWriteTest,
...
-                                          {Options::DELETION_VECTORS_ENABLED, "true"},
-                                          {Options::CHANGELOG_PRODUCER, "lookup"}},
-                                         /*with_io_manager=*/true),
-        "C++ Paimon does not support changelog-producer yet");
+       TestCreateFileStoreWriteShouldSucceedWhenLookupChangelogConfigured) {
+    ASSERT_OK(CreateSingleStringFileStoreWrite({{"bucket", "1"},
+                                                {Options::DELETION_VECTORS_ENABLED, "true"},
+                                                {Options::CHANGELOG_PRODUCER, "lookup"}},
+                                               /*with_io_manager=*/true)
+                  .status());
```
lxy-9602: ASSERT_OK(Func())

src/paimon/core/mergetree/compact/merge_tree_compact_manager_factory.cpp
```diff
@@ -241,29 +270,37 @@ MergeTreeCompactManagerFactory::CreateLookupRewriterWithDeletionVector(
...
                                        data_schema, trimmed_primary_keys, options, pool));
             PAIMON_ASSIGN_OR_RAISE(
-                std::shared_ptr<MergeFunctionWrapper<KeyValue>> merge_function_wrapper,
+                std::unique_ptr<RowCompactedSerializer> value_serializer,
+                CreateChangelogValueSerializer(data_schema, should_produce_changelog, pool));
+            PAIMON_ASSIGN_OR_RAISE(
+                InternalRowEqualizer::Equalizer value_equalizer,
+                CreateChangelogValueEqualizer(data_schema, options, should_produce_changelog));
+            PAIMON_ASSIGN_OR_RAISE(
+                std::shared_ptr<MergeFunctionWrapper<ChangelogResult>> merge_function_wrapper,
```
lxy-9602: Could we refactor this a bit? It looks like very similar logic has been implemented multiple times now.

src/paimon/core/mergetree/compact/merge_tree_compact_rewriter.cpp
```diff
@@ -146,11 +146,40 @@ MergeTreeCompactRewriter::CreateRollingRowWriter(int32_t level) {
...
+            /*is_changelog=*/false, pool_);
+    }
+    return std::make_unique<MergeTreeCompactRewriter::KeyValueRollingFileWriter>(
+        options_.GetTargetFileSize(/*has_primary_key=*/true),
+        /*target_file_row_num=*/std::numeric_limits<int64_t>::max(), factory);
+}
+
+Result<std::unique_ptr<MergeTreeCompactRewriter::KeyValueRollingFileWriter>>
+MergeTreeCompactRewriter::CreateRollingChangelogWriter(int32_t level) {
+    std::shared_ptr<FileFormat> format = options_.GetChangelogFileFormat();
```
lxy-9602, line 147: `CreateRollingRowWriter` and `CreateRollingChangelogWriter` duplicate the path-factory setup, shredding-plan selection, data-file-writer factory construction, and rolling-writer construction. The same shredding-vs-normal factory selection is also repeated in `MergeTreeWriter` and `PostponeBucketWriter`. Could we extract a shared `CreateKeyValueDataFileWriterFactory(...)` helper, while keeping format/path, FileSource, and rolling row-limit decisions at the callers? At minimum, these two methods could delegate to a private `CreateRollingWriter(level, is_changelog)` helper.

src/paimon/core/mergetree/merge_tree_writer.cpp
```diff
@@ -329,15 +390,38 @@ MergeTreeWriter::CreateRollingRowWriter() const {
...
     }
     return std::make_unique<RollingFileWriter<KeyValueBatch, std::shared_ptr<DataFileMeta>>>(
         options_.GetTargetFileSize(/*has_primary_key=*/true), options_.GetTargetFileRowNum(),
         factory);
 }
 
+Result<std::unique_ptr<RollingFileWriter<KeyValueBatch, std::shared_ptr<DataFileMeta>>>>
+MergeTreeWriter::CreateRollingChangelogWriter() const {
+    std::shared_ptr<SingleFileWriterFactory<KeyValueBatch, std::shared_ptr<DataFileMeta>>> factory;
+    PAIMON_ASSIGN_OR_RAISE(
```
lxy-9602, line 400: This is very similar to `CreateRollingRowWriter`. Could we extract the shared parts?

src/paimon/core/mergetree/merge_tree_writer_test.cpp
```diff
@@ -295,6 +300,182 @@ TEST_P(MergeTreeWriterTest, TestSimple) {
...
+                     expected_changelog, "parquet");
+}
+
+TEST_P(MergeTreeWriterTest, TestInputChangelogIgnoresTargetFileRowNum) {
+    ASSERT_OK_AND_ASSIGN(CoreOptions options,
+                         CoreOptions::FromMap({{Options::FILE_FORMAT, "orc"},
+                                               {Options::CHANGELOG_PRODUCER, "input"},
+                                               {Options::TARGET_FILE_ROW_NUM, "1"},
+                                               {Options::WRITE_BATCH_SIZE, "1"}}));
+
```
lxy-9602, line 377: I’m a bit curious: since `TargetFileRowNum` is already defined at batch granularity, if we only write a single batch, wouldn’t it end up as one file regardless of whether `ignore` is enabled?

test/inte/write_and_read_inte_test.cpp
```diff
@@ -630,6 +676,652 @@ TEST_P(WriteAndReadInteTest, TestPKSimple) {
...
+
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<RecordBatch> initial_batch,
+                         TestHelper::MakeRecordBatch(arrow::struct_(fields), R"([["Alice", 10]])",
+                                                     /*partition_map=*/{}, /*bucket=*/0, {}));
+    ASSERT_OK(helper->WriteAndCommit(std::move(initial_batch), /*commit_identifier=*/0,
+                                     /*expected_commit_messages=*/std::nullopt));
+
+    std::string table_path = PathUtil::JoinPath(test_dir_, "foo.db/bar");
+    auto compact_and_commit = [this, &options, &table_path](int64_t commit_identifier) -> Status {
+        WriteContextBuilder write_context_builder(table_path, "commit_user");
```
lxy-9602: Why not use `CompactAndCommit`?

test/inte/write_and_read_inte_test.cpp
```diff
@@ -630,6 +676,652 @@ TEST_P(WriteAndReadInteTest, TestPKSimple) {
...
+        file_store_write->PrepareCommit(/*wait_compaction=*/true, /*commit_identifier=*/1));
+    ASSERT_OK(file_store_write->Close());
+    ASSERT_FALSE(compact_messages.empty());
+    CommitContextBuilder commit_context_builder(table_path, "commit_user");
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<CommitContext> commit_context,
+                         commit_context_builder.SetOptions(options).Finish());
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<FileStoreCommit> file_store_commit,
+                         FileStoreCommit::Create(std::move(commit_context)));
+    ASSERT_OK(file_store_commit->Commit(compact_messages, /*commit_identifier=*/1));
+
```
lxy-9602, line 798: Why not use `CompactAndCommit`?

lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #250, opened 2026-08-26
fix(compaction): persist keys in lookup SST files while first-row merge function

### LINKED ISSUE #251: [Bug] persist keys in SST files while first-row merge function
### Search before asking

### Paimon-cpp version

- an existing `key=1` in L2;
- `key=1` and `key=2` in L0;
- compaction of only L0 into L1;
- the output is supposed to be only `key=2`.

The previous implementation special-cased `bool` by creating and immediately closing the lookup writer. This produced a valid but empty SST file. As a result, a compaction could incorrectly retain a newer row even though the first row for that key already existed.

### Minimal reproduce step

- an existing `key=1` in L2;
- `key=1` and `key=2` in L0;
- compaction of only L0 into L1;
- the output is supposed to be only `key=2`.

The previous implementation special-cased `bool` by creating and immediately closing the lookup writer. This produced a valid but empty SST file. As a result, a compaction could incorrectly retain a newer row even though the first row for that key already existed.

### Anything else?

_No response_

lszskye:
### Purpose

Linked issue: close #251 

First-row lookup uses `LookupLevels<bool>` because it only needs to determine whether a primary key already exists.

The previous implementation special-cased `bool` by creating and immediately closing the lookup writer. This produced a valid but empty SST file. As a result, a compaction could incorrectly retain a newer row even though the first row for that key already existed.

This change:

- removes the `bool`-specific empty SST shortcut;
- writes serialized primary keys with empty value payloads;
- uses a key-only read schema for first-row lookup, avoiding unnecessary value-column reads.

### Tests

Added `LookupMergeTreeCompactRewriterTest.TestFirstRowLooksUpExistingKeys`, covering:

- an existing `key=1` in L2;
- `key=1` and `key=2` in L0;
- compaction of only L0 into L1;
- verification that the output contains only `key=2`.

### API and Format

No public API or protocol changes.

### Documentation

No documentation changes are required.

### Generative AI tooling

Generated-by: Codex (GPT-5)

### DIFF
--- a/src/paimon/core/mergetree/compact/lookup_merge_tree_compact_rewriter_test.cpp
+++ b/src/paimon/core/mergetree/compact/lookup_merge_tree_compact_rewriter_test.cpp
@@ -484,6 +484,45 @@ TEST_F(LookupMergeTreeCompactRewriterTest, TestFirstRowRewrite) {
     CheckResult(compact_file_name, table_schema, "orc", expected_array);
 }
 
+TEST_F(LookupMergeTreeCompactRewriterTest, TestFirstRowLooksUpExistingKeys) {
+    std::map<std::string, std::string> options = {{Options::MERGE_ENGINE, "first-row"},
+                                                  {Options::FILE_FORMAT, "orc"}};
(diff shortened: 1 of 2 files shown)


lxy-9602 (approved):
+1

PR #253, opened 2026-08-27
perf(index): reuse primary-key index groups with retired source files

### LINKED ISSUE #252: [Improvement] Reuse primary-key sorted index after partial source retirement
### Search before asking

### Paimon-cpp version

Apache Paimon C++ `main` at `4027601`.

### Minimal reproduce step

1. Create a primary-key table and compact data files `A` and `B` into a positive level.
2. Build a source-backed scalar index payload whose immutable source metadata is `[A, B]`.
3. Commit an update which retires `A`, keeps `B` active with a deletion vector, and adds an uncovered replacement file `C`. The original payload remains active.
4. Scan an indexed predicate.

### What does not meet expectations?

The reader currently validates a payload only when its complete source list equals the current active source set of the whole level. It therefore rejects `[A, B]` against active `[B, C]` and turns both surviving indexed data and the new file into ordinary scans.

The payload still has a valid immutable ordinal namespace for `B`. The reader should reuse it for the active intersection, preserve the deletion vector on `B`, and leave only uncovered `C` on the ordinary scan path. Multiple retained payload groups should be accepted only when their active source intersections are disjoint; ambiguous overlap must continue to fall back.

### Anything else?

The fix can remain read-side only: no public API, storage format, or index protocol change is required.

- [x] I am willing to submit a PR.

wangyong9999:
### Purpose

Linked issue: close #252

A source-backed primary-key index payload uses one immutable ordered source list as its ordinal namespace. After an update retires only part of that list, the current reader requires the payload to equal the complete active source set of its data level. It rejects the still-valid payload and degrades every surviving indexed file to an ordinary scan. Query results remain correct; this PR only avoids that performance fallback.

The same optimization is tracked for Java and Python in [apache/paimon#9458](https://github.com/apache/paimon/issues/9458), with the aligned change in [apache/paimon#9460](https://github.com/apache/paimon/pull/9460).

**Key code and data-flow changes**

- Keep the payload's complete source list for group-ordinal localization, but compute coverage from its intersection with eligible active files at `source_meta.DataLevel()`.
- Preserve the existing one-group-per-level invariant. Multiple valid payload candidates for one level are rejected and fall back to an ordinary scan.
- Map an accepted group only to files allowed by `PrimaryKeyIndexSourcePolicy::ShouldRead`, using `data level + file name + row count` as the scan-side source identity.
- Preserve retired prefix or middle source offsets and deletion files when converting surviving files to `IndexedSplit`.
- Leave uncovered, newly written, or cross-level files on the existing ordinary-scan path.

This restores index reuse for surviving compact files without introducing a new writer lifecycle or allowing an uncovered file to depend on stale index metadata.

**Error and compatibility boundaries**

- Source names remain canonically ordered; an active source must match the payload's recorded row count.
- A payload with no active source at its metadata-declared level is rejected. A file with the same name and row count at another level cannot inherit the group.
- Invalid field/type metadata, row range, source metadata, or multiple candidates at one level fail closed to the existing ordinary scan.
- No public include API, storage format, index protocol, write path, or concurrency behavior changes.

### Tests

- Current x86_64 GCC 8 internal branch: focused `PkSortedBucketIndexStateTest.*:PrimaryKeySortedIndexScanTest.*` — 37/37 passed.
- Current x86_64 GCC 8 internal branch: full `core_test` — 1,747 passed, 100 skipped, 0 failed (1,847 total).
- Follow-up `DataLevel` binding regression set — 3/3 passed: cross-level source fallback, retired-source ordinal preservation, and wrong-level payload rejection.
- Added/retained coverage for partial active intersection, one group per level, metadata level mismatch, retired prefix/middle ordinal offsets, row-count mismatch, a mixed update with deletion-vector preservation, a non-eligible APPEND source, and cross-level source movement.
- `clang-format` dry-run and `git diff --check` pass.

### API and Format

No public include API, storage format, or index protocol changes. The change only broadens safe read-side reuse of one existing source-backed payload per level.

### Documentation

Updated the primary-key global-index user guide to describe immutable source groups, partial active coverage, the one-group-per-level boundary, declared-level validation, and fallback behavior.

### Generative AI tooling

### DIFF
--- a/src/paimon/core/index/pksorted/pk_sorted_bucket_index_state.cpp
+++ b/src/paimon/core/index/pksorted/pk_sorted_bucket_index_state.cpp
@@ -36,7 +36,6 @@ struct PayloadCandidate {
     std::shared_ptr<IndexFileMeta> payload;
     std::shared_ptr<PkSortedIndexGroup> group;
     std::vector<PrimaryKeyIndexSourceFile> active_sources;
-    bool conflicted = false;
 };
 }  // namespace
 
@@ -59,20 +58,9 @@ PkSortedBucketIndexState PkSortedBucketIndexState::FromActiveDataFiles(
             });
     }
 
-    std::map<std::string, int64_t> active_sources;
-    std::set<std::string> ambiguous_active_source_names;
-    for (const auto& level_sources : sources_by_level) {
-        for (const PrimaryKeyIndexSourceFile& source : level_sources.second) {
-            if (!active_sources.emplace(source.file_name, source.row_count).second) {
-                ambiguous_active_source_names.insert(source.file_name);
-            }
-        }
-    }
-
     // Keep the complete immutable source group for ordinal localization, but only claim the
-    // sources which are still active in this snapshot. Several disjoint groups may coexist
-    // after an update; no active source file may be claimed by two groups.
-    std::vector<PayloadCandidate> candidates;
+    // sources which are still active at the metadata-declared level in this snapshot.
+    std::map<int32_t, std::vector<PayloadCandidate>> candidates_by_level;
     std::vector<std::shared_ptr<IndexFileMeta>> rejected;
     for (const std::shared_ptr<IndexFileMeta>& payload : active_payloads) {
         if (payload == nullptr) {
@@ -91,21 +79,30 @@ PkSortedBucketIndexState PkSortedBucketIndexState::FromActiveDataFiles(
             continue;
         }
         PrimaryKeyIndexSourceMeta source_meta = std::move(source_meta_result).value();
+        auto active_level = sources_by_level.find(source_meta.DataLevel());
+        if (active_level == sources_by_level.end()) {
+            rejected.push_back(payload);
+            continue;
+        }
         const std::vector<PrimaryKeyIndexSourceFile>& payload_sources = source_meta.SourceFiles();
         bool valid_candidate = !payload_sources.empty();
         std::vector<PrimaryKeyIndexSourceFile> active_intersection;
+        size_t active_source_index = 0;
         for (size_t i = 0; valid_candidate && i < payload_sources.size(); i++) {
             const PrimaryKeyIndexSourceFile& source = payload_sources[i];
             if (i > 0 && payload_sources[i - 1].file_name >= source.file_name) {
                 valid_candidate = false;
                 break;
             }
-            auto active_source = active_sources.find(source.file_name);
-            if (active_source == active_sources.end()) {
+            while (active_source_index < active_level->second.size() &&
+                   active_level->second[active_source_index].file_name < source.file_name) {
+                active_source_index++;
+            }
+            if (active_source_index == active_level->second.size() ||
+                active_level->second[active_source_index].file_name != source.file_name) {
                 continue;
             }
-            if (active_source->second != source.row_count ||
-                ambiguous_active_source_names.count(source.file_name) > 0) {
+            if (active_level->second[active_source_index].row_count != source.row_count) {
                 valid_candidate = false;
                 break;
             }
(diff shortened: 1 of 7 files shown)


src/paimon/core/index/pksorted/pk_sorted_bucket_index_state.cpp
```diff
@@ -49,10 +59,20 @@ PkSortedBucketIndexState PkSortedBucketIndexState::FromActiveDataFiles(
...
-    // Match payloads against the expected level sources; anything that does not decode or
-    // does not exactly cover its level is rejected.
-    std::map<int32_t, std::vector<std::shared_ptr<IndexFileMeta>>> payloads_by_level;
-    std::map<int32_t, std::vector<PrimaryKeyIndexSourceMeta>> payload_metas_by_level;
+    std::map<std::string, int64_t> active_sources;
+    std::set<std::string> ambiguous_active_source_names;
+    for (const auto& level_sources : sources_by_level) {
+        for (const PrimaryKeyIndexSourceFile& source : level_sources.second) {
+            if (!active_sources.emplace(source.file_name, source.row_count).second) {
+                ambiguous_active_source_names.insert(source.file_name);
```
lxy-9602: Is ignoring `source_meta.DataLevel()` intentional to support metadata-only file upgrades? If so, could we add a regression test and document this behavior?

lxy-9602:
The current changes seem reasonable to me. That said, to avoid `PkSortedBucketIndexState` in C++ gradually diverging too much from the Java implementation and becoming harder to align later — for example, if additional features or refactors are introduced on the Java side in the future — I’d suggest merging this logic into `paimon-java` first, and then bringing it into C++.

src/paimon/core/index/pksorted/pk_sorted_bucket_index_state.cpp
```diff
@@ -71,39 +91,86 @@ PkSortedBucketIndexState PkSortedBucketIndexState::FromActiveDataFiles(
...
             rejected.push_back(payload);
             continue;
         }
-        payloads_by_level[source_meta.DataLevel()].push_back(payload);
-        payload_metas_by_level[source_meta.DataLevel()].push_back(std::move(source_meta));
+        candidates.push_back(
+            {payload, std::move(group), std::move(active_intersection), /*conflicted=*/false});
+    }
+
+    std::map<std::pair<std::string, int64_t>, std::vector<size_t>> candidates_by_source;
```
wangyong9999: The Java maintainer still keeps exactly one sorted-index group per level and rejects multiple units, so no current writer can produce this state. This conflict graph adds a C++-only lifecycle model beyond the partial-retirement fix. Could we keep the one-group-per-level invariant here and limit this PR to reusing that group’s active intersection?

wangyong9999: Updated. The same-level conflict graph and its multi-group test have been removed. The reader again accepts at most one payload group per level and only broadens reuse to that group’s eligible active intersection.

src/paimon/core/index/pksorted/pk_sorted_bucket_index_state.cpp
wangyong9999: Not intentional. The current revision now resolves the active set from `source_meta.DataLevel()` and rejects a payload whose declared level has no eligible active intersection. `RejectsPayloadForDifferentActiveDataLevel` covers the metadata-level mismatch.

wangyong9999:
Thanks. I narrowed the revision to preserve the existing one-group-per-level lifecycle and added strict `source_meta.DataLevel()` validation; it does not add a writer/protocol state or a C++-only multi-group model. The remaining change fixes C++ read-side validation of an already-written immutable source group after partial retirement. Because no Java behavior or persisted format changes, could this narrow reader correction proceed independently while keeping the Java/C++ lifecycle aligned?

wangyong9999:
I checked the current Paimon Java and Python readers and confirmed the same exact-active-set rejection there. I filed apache/paimon#9458 and opened the aligned Java/Python fix in apache/paimon#9460. This C++ branch now also binds scan-side source matching to the group's data level, with a regression test for a source moving to another level.

lxy-9602:
@wangyong9999 I’m not sure this is best described as a bug, since it doesn’t affect correctness and seems more like a performance fallback when the index is not applied. With that in mind, the current title may be slightly misleading.

Also, I think it may be better to wait for the Java side to land first before moving this forward, since the Java implementation covers more complex scenarios and the C++ path is currently a subset of that functionality. From what I can tell, the optimization on the C++ side looks reasonable, but I’d feel more comfortable waiting for full validation on the Java side before treating the overall behavior as fully confirmed. Thanks again for the patience here.

wangyong9999:
Agreed. This is a performance fallback, not a correctness issue. I changed the PR title to `perf`, changed both tracking issues to `Improvement`, and clarified that query results remain correct. The aligned Java/Python change is apache/paimon#9460; I will leave this PR pending until that change completes validation and lands, then update here. No further C++ behavior change is planned.

lxy-9602 (approved):
+1 Thanks for the adjustment!

PR #254, opened 2026-08-27
feat(build): make TBB dependency configurable

gripleaf:
### Purpose

  Linked issue: N/A

  Make the oneTBB dependency optional through the `PAIMON_USE_TBB` CMake option.

  - Keep `PAIMON_USE_TBB=ON` by default to preserve existing behavior.
  - Skip resolving, building, including, and linking TBB when `PAIMON_USE_TBB=OFF`.
  - Encapsulate TBB concurrent containers under Paimon utility headers.
  - Provide C++17 standard-library implementations for builds without TBB.
  - Select the implementation at compile time:
    - `concurrent_hash_map.h` exposes the TBB implementation when TBB is enabled and includes `std_concurrent_hash_map.h` otherwise.
    - `concurrent_bounded_queue.h` exposes the TBB implementation when TBB is enabled and includes `std_concurrent_bounded_queue.h` otherwise.
  - Keep the class names, template parameters, and exposed methods consistent between the TBB and standard-library implementations.
  - Keep the `ConcurrentBoundedQueue` method names and relevant type aliases compatible with the corresponding `tbb::concurrent_bounded_queue` API.
  - Remove direct exposure of TBB queue types from the asynchronous key-value producer and consumer.

  ### Tests

  - Built `paimon_objlib` with `PAIMON_USE_TBB=ON`.
  - Built `paimon_objlib` with `PAIMON_USE_TBB=OFF`.
  - Ran the concurrent bounded queue tests with TBB enabled: 2 tests passed.
  - Ran the concurrent bounded queue tests with TBB disabled: 2 tests passed.
  - Compiled the concurrent hash map test translation unit with TBB enabled and disabled.
  - Ran pre-commit checks for all changed files.
  - Ran `git diff --check`.

  The complete unit and integration test suites were not run.

  ### API and Format

  This change does not affect the public API under `include/paimon/`.

  It does not change the storage format or protocol.

  The new concurrent container implementations are internal utilities under `src/paimon/common/utils/`.

  ### Documentation

  Updated `docs/source/building.rst` to document:

  - The `PAIMON_USE_TBB` CMake option.
  - TBB being enabled by default.
  - Building Paimon without resolving, building, or linking TBB.
  - `TBB_SOURCE` only being considered when TBB is enabled.

  ### Generative AI tooling


src/paimon/common/utils/concurrent_hash_map.h
```diff
@@ -22,56 +22,173 @@
...
+    bool equal(const Key& lhs, const Key& rhs) const {
+        return lhs == rhs;
+    }
+};
+
+template <typename Key, typename HashCompare>
+class HashCompareHasher {
+ public:
+    size_t operator()(const Key& key) const {
+        return HashCompare{}.hash(key);
```
wangyong9999: HashCompare{} creates a fresh comparator for every hash/equality call, unlike the TBB map which owns one comparator instance. With a stateful comparator (for example, one whose default constructor chooses a seed), PAIMON_USE_TBB=OFF can insert into one bucket and immediately search another; an insert-then-find repro returns missing. Please keep one comparator per backend and initialize both unordered-map adapters from that same state, with a regression test.

src/paimon/common/utils/std_concurrent_hash_map.h
```diff
@@ -0,0 +1,129 @@
...
+    }
+};
+
+namespace detail {
+
+template <typename Key, typename HashCompare>
+class HashCompareHasher {
+ public:
+    size_t operator()(const Key& key) const {
+        return HashCompare{}.hash(key);
```
wangyong9999: The previous issue is still present after this refactor: HashCompare{} is rebuilt for every hash/equality call. On this head, the same stateful-comparator repro still returns missing immediately after Insert when TBB is off (three comparator instances versus one with TBB). Please retain one comparator state per map and initialize both unordered-map adapters from it.

src/paimon/common/utils/std_concurrent_bounded_queue.h
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    void push(const T& value) {
+        T copied_value = value;
+        push(std::move(copied_value));
+    }
+
+    void push(T&& value) {
+        std::unique_lock<std::mutex> lock(mutex_);
+        capacity_available_.wait(lock, [this]() { return queue_.size() < capacity_; });
```
wangyong9999, line 63: try_pop() can observe the queue empty before producers woken from this wait reacquire the mutex and enqueue. AsyncKeyValueProducerAndConsumer::CleanUp() drains once and then waits for every consumer, so an early Close() with 9+ configured row-to-batch threads can leave consumers blocked here forever. A cleanup-order repro with 32 consumers times out 5/5 with this backend and completes 5/5 with oneTBB. Please explicitly cancel/unblock pending pushes or keep draining until all consumers exit, and add an early-close regression test.

src/paimon/common/utils/std_concurrent_hash_map.h
gripleaf: The std backend now creates one HashCompare instance per ConcurrentHashMap and shares that same instance between the unordered_map hasher and equality adapters. It no longer reconstructs HashCompare{} for each hash/equality call.
I also added a stateful-comparator regression test covering insert, find, and erase, and verified it with both PAIMON_USE_TBB=ON and PAIMON_USE_TBB=OFF.

src/paimon/common/utils/std_concurrent_bounded_queue.h
gripleaf: CleanUp() now continues draining both the result queue and the key-value queue until the producer and all consumer futures have finished. This unblocks pending bounded-queue pushes and avoids relying on a single drain observing items pushed by newly awakened workers.

I also added an early-close regression test that synchronizes 32 consumers before they push into the bounded result queue, ensuring multiple consumers are blocked when Close() is called. The regression test passed in 10 consecutive runs with the non-TBB backend.

src/paimon/common/utils/std_concurrent_hash_map.h
wangyong9999: Verified on f481ffa. Both adapters now share the same HashCompare instance, and the stateful-comparator repro passes with TBB enabled and disabled. Fixed, thanks.

src/paimon/common/utils/std_concurrent_bounded_queue.h
wangyong9999: Verified on f481ffa. Cleanup now keeps draining until the producer and all consumers finish, and the early-close stress case completes with both backends, including under ThreadSanitizer. Fixed, thanks.

lxy-9602:
Before starting the review, I’d like to better understand the specific motivation behind this PR — for example, which platform or downstream project is unable to use TBB, and what the performance trade-offs are when disabling it. BTW, the current CI also doesn’t seem to cover the `PAIMON_USE_TBB=OFF` branch.

gripleaf:
> Before starting the review, I’d like to better understand the specific motivation behind this PR — for example, which platform or downstream project is unable to use TBB, and what the performance trade-offs are when disabling it. BTW, the current CI also doesn’t seem to cover the `PAIMON_USE_TBB=OFF` branch.

The motivation is not platform compatibility. We integrate Paimon C++ as a DSO to isolate its large dependency graph from the existing service. However, both the service and the DSO bring their own TBB runtime and worker pool, which can cause thread oversubscription. PAIMON_USE_TBB=OFF allows the DSO to avoid introducing the second TBB runtime.

You are also right that CI currently does not cover PAIMON_USE_TBB=OFF; I will add a dedicated non-TBB build/test configuration.

zjw1111:
@gripleaf Thanks for the clarification. If the main concern is that the Paimon DSO embeds a second TBB runtime, could this be addressed by building Paimon against the same shared TBB used by the host service, for example with:

```bash
-DTBB_SOURCE=SYSTEM -DTBB_ROOT=/path/to/the/host/tbb/prefix
```

The existing dependency resolver already supports `TBB_SOURCE=SYSTEM`. If both the host and the Paimon DSO resolve to the same compatible `libtbb.so`, this should avoid embedding the bundled static TBB into the DSO and should let them share one runtime. Of course, this would need to be verified from the final `DT_NEEDED` entries / `ldd` output, since `TBB_SOURCE=SYSTEM` by itself does not guarantee that the selected library is shared or identical to the host copy.

Also, Paimon currently appears to use TBB only for `concurrent_hash_map` and `concurrent_bounded_queue`; the producer/consumer threads themselves are created with `std::async`. Could you provide profiling or thread/runtime evidence showing that the second TBB copy actually initializes another worker pool and causes measurable oversubscription? That would help establish whether a non-TBB implementation is necessary, rather than simply reusing the host shared TBB.

gripleaf:
> @gripleaf Thanks for the clarification. If the main concern is that the Paimon DSO embeds a second TBB runtime, could this be addressed by building Paimon against the same shared TBB used by the host service, for example with:
> 
> ```shell
> -DTBB_SOURCE=SYSTEM -DTBB_ROOT=/path/to/the/host/tbb/prefix
> ```
> 
> The existing dependency resolver already supports `TBB_SOURCE=SYSTEM`. If both the host and the Paimon DSO resolve to the same compatible `libtbb.so`, this should avoid embedding the bundled static TBB into the DSO and should let them share one runtime. Of course, this would need to be verified from the final `DT_NEEDED` entries / `ldd` output, since `TBB_SOURCE=SYSTEM` by itself does not guarantee that the selected library is shared or identical to the host copy.
> 
> Also, Paimon currently appears to use TBB only for `concurrent_hash_map` and `concurrent_bounded_queue`; the producer/consumer threads themselves are created with `std::async`. Could you provide profiling or thread/runtime evidence showing that the second TBB copy actually initializes another worker pool and causes measurable oversubscription? That would help establish whether a non-TBB implementation is necessary, rather than simply reusing the host shared TBB.

@zjw1111  Thanks for the suggestion. We’ve addressed this issue by having Paimon dynamically link against the TBB provided by the host side, so both now share the same TBB runtime instead of introducing a separate bundled copy.

PR #257, opened 2026-08-28
perf(parquet): forward dictionary encoding through append compaction rewrite

### LINKED ISSUE #230: [Feature] Support dictionary passthrough for Parquet compaction rewrite
### Search before asking

### Motivation

`AppendOnlyFileStoreWrite::CompactRewrite` (`src/paimon/core/operation/append_only_file_store_write.cpp:141`) is a straight batch passthrough loop: `reader->NextBatch()` → `ImportArray` → drop `_VALUE_KIND` → `ExportArray` → `rewriter->Write()` → `ParquetFormatWriter::AddBatch`. Nothing between the reader and the writer inspects or transforms the values.

For a dictionary-encoded `VARCHAR` column, that loop currently does the following per compaction:

1. arrow's Parquet reader expands the dictionary into a flat `StringArray`, writing one full copy of the string for every row;
2. the flat strings are exported and re-imported across the C data interface;
3. arrow's Parquet writer rebuilds a dictionary from those flat strings and hashes every value again.

For a low-cardinality column this is entirely wasted work — the file already had the dictionary, and the output file wants the same dictionary. Two things block the obvious shortcut today:

- paimon-cpp never sets `ArrowReaderProperties::set_read_dictionary` (`cpp/src/parquet/properties.h:888`), so dictionary columns are always flattened at read time;
- `ParquetFormatWriter::AddBatch` imports with the fixed logical `schema_` (`src/paimon/format/parquet/parquet_format_writer.cpp:76`), so a `DictionaryArray` cannot enter the writer even if the reader produced one.

Arrow 17 already supports both ends. `set_read_dictionary` makes the reader emit `DictionaryArray`, and `TypedColumnWriterImpl<DType>::WriteArrow` dispatches `::arrow::Type::DICTIONARY` to `WriteArrowDictionary` (`cpp/src/parquet/column_writer.cc:1324`), which writes dictionary-encoded pages directly without materializing the values. No arrow patch is required for the mechanism itself.

Velox made the same change on its Parquet writer, which — like ours — wraps `parquet::arrow::FileWriter`, and measured 2.4–3.2x for dictionary-encoded VARCHAR columns and up to 4.3x for all-dictionary rows ([velox#17988](https://github.com/facebookincubator/velox/issues/17988), [velox#17986](https://github.com/facebookincubator/velox/pull/17986)).

### Solution

1. **Read side**: opt-in `set_read_dictionary` for STRING/BINARY leaf columns in `CreateArrowReaderProperties` (`src/paimon/format/parquet/parquet_file_batch_reader.cpp`). Enabling it blindly is wrong — for a high-cardinality column it produces a useless dictionary and costs extra. Gate it on the column chunk actually being dictionary-encoded (`ColumnChunkMetaData::dictionary_page_offset()` / `encodings()`), and keep it off by default outside the rewrite path until measured.

2. **Writer side**: let `ParquetFormatWriter` accept dictionary-typed children. `Create` computes `write_schema` once from the logical type; a dictionary-encoded batch no longer matches it, so the schema needs reconciling per batch — this is what Velox calls the "reverse fixup". Simplest correct approach is to derive the import schema from the incoming batch's encoding and keep the Parquet write schema fixed, since the Parquet-level type is identical either way.

3. **Selective passthrough**, following Velox's `flattenIfNeeded()`: pass through dictionary-of-primitive only; flatten dictionary-of-complex and dictionary-of-dictionary. One awkward column must not force the whole batch to be flattened.

4. **Other consumers of the same batch must tolerate it.** `DataFileWriterBase::AddFileIndexBatch` (`src/paimon/core/io/data_file_writer_base.h:126`) hands the same array to `DataFileIndexWriter`, which fans out to `BitmapFileIndexWriter` / `RangeBitmapFileIndexWriter`; those read values directly. `src/paimon/common/data/columnar/columnar_utils.h:57` already unwraps `DictionaryArray` for string access and is a reasonable starting point. If a file index is configured on a passthrough column, either teach the index writer about dictionaries or fall back to flattening that column.

5. **Measure it** with the format-level benchmarks from #228, on low / medium / high cardinality VARCHAR, before enabling anything by default.

### Anything else?

Risks worth flagging up front:

- **Mixed encodings across batches.** A column can be dictionary-encoded in one input file and plain in the next, so consecutive batches in the same output row group may disagree. Arrow handles the fallback internally (dictionary → plain when the dictionary grows past the page limit), but the schema reconciliation in step 2 has to survive the transition mid-row-group.
- **Buffered row group interaction.** We write via `NewBufferedRowGroup` driven by `pool_->bytes_allocated()` (`src/paimon/format/parquet/parquet_format_writer.cpp:87`) and estimate size via the patched `GetBufferedSize()`. Dictionary passthrough changes what "buffered bytes" means, so `ReachTargetSize` behaviour should be re-checked — a rolling writer that mis-estimates size produces badly sized files.
- **Scope.** The clearest win is unaware-bucket append compaction, where the loop is a true passthrough. The primary-key merge path is not a passthrough (records flow through merge functions and sorting), so it is out of scope here.

Discovered while evaluating [velox#17988](https://github.com/facebookincubator/velox/issues/17988) against paimon-cpp. The other writer item from that issue, schema caching ([velox#17985](https://github.com/facebookincubator/velox/pull/17985)), is **already implemented** here — `ParquetFormatWriter::Create` computes the write schema once and `AddBatch` uses the `ImportRecordBatch(array, schema)` overload, which is exactly the shape Velox arrived at. Their parallel-column-writing experiment ([velox#18128](https://github.com/facebookincubator/velox/pull/18128)) was closed unmerged at 1.3–1.6x with ~58% of column-write time irreducibly serial, so it is not proposed here.

SteNicholas:
### Purpose

Linked issue: close #230

An append-only compaction rewrite copies rows into the new file without inspecting any value, so expanding a dictionary-encoded Parquet column on read and hashing it again on write is work neither side needs. This forwards the encoding instead.

**Reader.** `parquet.read.enable-dictionary-passthrough` makes `ParquetFileBatchReader` request `set_read_dictionary` for non-nested `STRING` columns whose every data page, in every row group of the file, is dictionary-encoded. A dictionary page alone cannot be the signal — a column that outgrew its page limit still carries the page it had already emitted — so the gate reads the encoding statistics. The file schema keeps reporting the logical type; only the emitted batches carry the dictionary.

`BINARY` is deliberately **not** forwarded, although Parquet stores it in the same `BYTE_ARRAY` leaf and dictionary-encodes it the same way: no value accessor here can read a `dictionary(int32, binary)`. `ColumnarUtils::GetView()` asserts on a dictionary whose values are neither `STRING` nor `LARGE_STRING` and returns an empty `string_view` in a release build, and `LiteralConverter` rejects it outright. Every consumer understands a `STRING` dictionary because the ORC reader has always produced one under lazy decoding; none was ever handed a `BINARY` one. Widening this needs those consumers first.

**Writer.** `ParquetFormatWriter` recovers each batch's encoding from its layout, because exporting through the Arrow C data interface drops the type. A layout pins down neither the index nor the offset width, so only `dictionary(int32, utf8|binary)` is recoverable — `ArrowUtils::IsDictionaryLayoutRecoverableValueType()`. That is the writer's capability and it matches Velox's `VARCHAR`/`VARBINARY`; the `STRING`-only restriction above belongs to the reader, not to the format, and is applied at the gate so the two do not get conflated. `CompactRewrite` decodes anything else **per column, while the type is still known**: the ORC reader's `dictionary(int64, large_utf8)` under lazy decoding, dictionaries below the top level, non-binary value types. The rest stay encoded, so one awkward column does not cost the others their encoding — unless a veto is in force, in which case every dictionary is decoded whatever its shape (see **Gating**). A dictionary holding nulls in its values is flattened at the writer — the one shape `parquet::arrow` rejects outright.

**Gating.** The option is **off by default and the rewrite never turns it on**; enabling it is the table's decision, taken after measuring. `GetDictionaryPassthroughVetoReason()` only *vetoes* it — when the output is not Parquet, when `parquet.enable-dictionary` is false, or when a shredding plan is active — and logs the reason at DEBUG for a table that did ask for the passthrough. Setting the option therefore never makes a rewrite fail; at worst it has no effect.

A veto is enforced on the batches as well as on the option, because turning the option off is not sufficient. A reader hands over a dictionary for reasons this option does not govern — a format's own lazy-decoding setting — and those stay on. `ParquetFormatWriter` is the only writer that recovers an encoding from a layout, so any other would import such a batch against its logical type and fail on the buffer count. `CompactRewrite` therefore derives `preserve_layout_recoverable_dictionaries` from the veto and decodes everything under one. The built-in ORC reader widens to `dictionary(int64, large_utf8)`, which is decoded for its shape anyway, so reaching this needs a reader outside this repository; `ArrowUtilsTest.TestFlattenUnresolvableDictionaries` covers it directly.

A file index configured on a forwarded column materializes that column alone.

**Known trade-off.** A Parquet column chunk carries one dictionary, so when the input files supply different ones the output keeps the first and falls back to plain for the rest of that row group. Output row groups are cut by `parquet.block.size` and by the writer's memory limit, which are not aligned to input file boundaries. The rewritten data is unchanged, but the output file can be larger than one written from materialized values, and that column then loses eligibility for the next compaction round. This is why the option is opt-in rather than automatic, and it is documented in `compaction.rst`.

Design follows Velox's `perf(parquet): Dictionary passthrough and selective flattening in Parquet writer` (facebookincubator/velox#17986): per-column selective flatten, only VARCHAR/VARBINARY passed through, dictionaries with null values flattened, import schema reconciled per batch. It differs in two places. Velox still holds a typed `Vector` inside its writer, whereas `FormatWriter::AddBatch(ArrowArray*)` here receives an untyped array, so the flatten has to run one layer up, before the type is dropped. And Velox's reader can hand out a dictionary for `VARBINARY` because its consumers are type-erased over `StringView`, while the accessors here switch on the Arrow type id — hence the reader-side `STRING` restriction.

### Tests

Unit:

- `ArrowUtilsTest.TestResolveDictionaryStructTypeFromLayout` — layout-derived resolution; rejects non-binary-like value types, `large_utf8`, and dictionaries below the top level; preserves a caller-declared dictionary type.
- `ArrowUtilsTest.TestFlattenUnresolvableDictionaries` — selective flatten: `dictionary(int64, large_utf8)` and `dictionary(int64, utf8)` decoded while the `int32` neighbour stays encoded; nested dictionary decoded; unchanged batch returned by identity; sliced batch keeps its offset. And with `preserve_layout_recoverable_dictionaries` false, the layout-recoverable `dictionary(int32, utf8)` is decoded too, and a batch of only such columns is no longer returned by identity.
- `ReaderUtilsTest.TestApplyBitmapToReadBatchKeepsDictionaryEncoding` — a deletion vector filters batches by slice + `arrow::Concatenate`; pins that the encoding survives it.
- `DataFileIndexWriterTest.TestDictionaryEncodedIndexedColumnRoundTrip` — bitmap index built from forwarded columns across two batches with different dictionaries, which also covers the writer-held Arrow pool.
- `ParquetFileBatchReaderTest.TestDictionaryPassthrough` — on / explicitly off / option absent / file without dictionary pages.
- `ParquetFileBatchReaderTest.TestDictionaryPassthroughSkipsBinaryColumn` — `f8` (`STRING`) is forwarded and `f9` (`BINARY`) is not, from the same file with the same physical type and pages.
- `ParquetFileBatchReaderTest.TestDictionaryPassthroughSkipsFallbackToPlain` — a column that falls back to plain inside one chunk is declined.
- `ParquetFileBatchReaderTest.TestDictionaryPassthroughRequiresEveryRowGroup` — first row group fully dictionary-encoded, second falls back; the whole column is declined.
- `ParquetFormatWriterTest.TestWriteDictionaryEncodedColumn` / `ChangingAcrossBatches` / `WithNullsInDictionary` / `WithNullRows` / `WithDuplicateValues` / `EmptyBatch` / `OfBinaryColumn` / `OfUnsupportedTypeIsRejected`.
- `ParquetFormatWriterTest.TestGetEstimateLengthWithDictionaryBatches` — `GetEstimateLength()` and `ReachTargetSize()` still drive file rolling when batches arrive encoded.

Integration:

- `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthrough` (Parquet + ORC) — the table opts in; asserts the input read types (`s` as `dictionary(int32, utf8)` on Parquet, `id` and `b` materialized, `s` as `dictionary(int64, large_utf8)` on ORC), a full read of those still-encoded files through the whole read path, the rewrite, that the output column has fallen back to plain, and a predicate read through the bitmap index.
- `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthroughDefaultOff` — a table that says nothing about the option lands on the same rows and on a *better* encoded output, which is why it is not on by default.
- `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthroughVetoedByShredding` — the option is set and vetoed; the input arrives encoded and the compacted column comes back dictionary-encoded end to end, which only a rewrite from materialized values produces.

Benchmarks (`benchmark/parquet_format_benchmark.cpp`): `BM_ParquetWrite_String`, `BM_ParquetWrite_DictionaryStringIntoStringSchema` and `BM_ParquetWrite_ChangingDictionaryStringIntoStringSchema` write the same logical column at the same cardinality — the third rotates the alphabet and shifts the indices back by the same amount, so every batch brings a different dictionary over identical data. The second against the first is what forwarding buys; the third against the second is what the fallback costs in time; the third against the first is the output size a rewrite that materialized and rebuilt would have produced. `parquet_format_benchmark_test.cpp` pins that premise. **These are format-writer microbenchmarks and do not stand in for compaction time, CPU and peak memory on a real table** — that measurement is what a user should make before enabling the option, which is why it ships off by default.

### API and Format

- New `ArrowUtils::IsDictionaryLayoutRecoverableValueType()`, `ArrowUtils::ResolveDictionaryStructTypeFromLayout()` and `ArrowUtils::FlattenUnresolvableDictionaries()`. Neither of the first two is Parquet-specific — they read `ArrowArray::dictionary` and the caller's declared type, and `DataFileWriterBase::AddFileIndexBatch` runs the resolution for every format. The flatten takes `preserve_layout_recoverable_dictionaries` with no default, so a new caller has to state whether its destination recovers an encoding from a layout.
- New table option `parquet.read.enable-dictionary-passthrough`, default `false`. It is a *read* option, so it applies to every read of the table and not only to the compaction rewrite; an eligible column then reaches the consumer as an Arrow `DictionaryArray`.
- `AppendOnlyFileStoreWrite` reads the two Parquet option names as local constants instead of including the format layer's headers, so core does not depend on symbols a replacement Parquet plugin need not export.

### Documentation

- `docs/source/user_guide/compaction.rst` — a "Dictionary Passthrough" section: what it does, that it is opt-in and why, the read-option caveat, per-file eligibility and the vetoes, that a veto is enforced on the batches and not only on the option, and a "Trade-off" subsection on the single dictionary per column chunk.
- `docs/source/examples/benchmark.rst` — the flat / shared-dictionary / changing-dictionary trio and what each pair measures.

### DIFF
--- a/docs/source/user_guide/compaction.rst
+++ b/docs/source/user_guide/compaction.rst
@@ -88,6 +88,40 @@ After compaction, if the last output file is still smaller than
 ``compaction.file-size``, it is placed back into the compaction queue for future
 merging.
 
+Dictionary Passthrough
+~~~~~~~~~~~~~~~~~~~~~~
+An append-only compaction rewrite copies rows into the new file without
+inspecting any value, so a Parquet column that an input file already stores
+dictionary-encoded is forwarded to the writer still encoded instead of being
+expanded to one copy of the value per row and re-encoded. This saves the reader
+materializing the values and the writer hashing them again; how much that is
+worth depends on the column, and low-cardinality ``STRING``/``BINARY`` columns
+benefit most. Primary-key compaction merges rows and is not covered.
+
+This applies automatically. Eligibility is decided per input file: a non-nested
+``STRING``/``BINARY`` column is forwarded when its data pages are
+dictionary-encoded throughout every row group of *that* file, so one input file
+can be read encoded while the next one is read as ordinary values, and the
+writer takes both. A high-cardinality column that started dictionary-encoded and
+fell back to plain encoding therefore does not qualify, even though it still
+carries a dictionary page. Passthrough is also skipped when the table writes a
+format other than Parquet, when ``parquet.enable-dictionary`` is ``false``
+because the writer would only expand the values again, or when variant/map
+shredding is configured because those writers reshape each batch against a fixed
+physical schema.
+
+If a file index is configured on a forwarded column, that column alone is
+materialized so the index still sees its values; the other columns stay encoded.
+
+Passthrough changes what the rewrite costs, not what it produces, with one
+exception worth knowing: a Parquet column chunk can only carry one dictionary,
+so when the input files supply different dictionaries the output column keeps
+the first and falls back to plain encoding for the rest of the row group. The
+rewritten data is unchanged either way, but the output file may be larger than a
+rewrite that rebuilt a single dictionary from materialized values. Set
+``parquet.read.enable-dictionary-passthrough`` to ``false`` on the table to turn
+the optimization off and always rebuild.
+
 Append-Only Table Compaction Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
--- a/docs/source/user_guide/compaction.rst
+++ b/docs/source/user_guide/compaction.rst
@@ -95,32 +95,66 @@ inspecting any value, so a Parquet column that an input file already stores
 dictionary-encoded is forwarded to the writer still encoded instead of being
 expanded to one copy of the value per row and re-encoded. This saves the reader
 materializing the values and the writer hashing them again; how much that is
-worth depends on the column, and low-cardinality ``STRING``/``BINARY`` columns
-benefit most. Primary-key compaction merges rows and is not covered.
+worth depends on the column, and low-cardinality ``STRING`` columns benefit most.
+Primary-key compaction merges rows and is not covered.
 
-This applies automatically. Eligibility is decided per input file: a non-nested
-``STRING``/``BINARY`` column is forwarded when its data pages are
-dictionary-encoded throughout every row group of *that* file, so one input file
-can be read encoded while the next one is read as ordinary values, and the
-writer takes both. A high-cardinality column that started dictionary-encoded and
-fell back to plain encoding therefore does not qualify, even though it still
-carries a dictionary page. Passthrough is also skipped when the table writes a
-format other than Parquet, when ``parquet.enable-dictionary`` is ``false``
-because the writer would only expand the values again, or when variant/map
-shredding is configured because those writers reshape each batch against a fixed
-physical schema.
+This is **off by default** and is enabled per table by setting
+``parquet.read.enable-dictionary-passthrough`` to ``true``. It is opt-in rather
+than automatic because of the output file size trade-off described below, which
+depends on the data and is not a win on every table. Measure before turning it
+on, and compare output file size as well as compaction time.
+
+Note that this is a *read* option, so it applies to every read of the table and
+not only to the compaction rewrite. The values are unchanged, but an eligible
+column reaches the consumer as an Arrow ``DictionaryArray`` rather than one value
+per row, and a consumer that reads columns through its own accessors has to
+unwrap it. Only a consumer that forwards batches without inspecting values -
+which is what the rewrite does - gains anything from the encoding.
+
+Once enabled, eligibility is decided per input file: a non-nested ``STRING``
+column is forwarded when its data pages are dictionary-encoded throughout every
+row group of *that* file, so one input file can be read encoded while the next
+one is read as ordinary values, and the writer takes both. A high-cardinality
+column that started dictionary-encoded and fell back to plain encoding therefore
+does not qualify, even though it still carries a dictionary page. ``BINARY`` is
+not forwarded although Parquet stores it in the same physical type and
+dictionary-encodes it the same way, because the value accessors cannot read a
+``BINARY`` dictionary. The rewrite also overrides the option back to ``false``
+when the table writes a format other than Parquet, when
+``parquet.enable-dictionary`` is ``false`` because the writer would only expand
+the values again, or when variant/map shredding is configured because those
+writers reshape each batch against a fixed physical schema. Setting the option
+therefore never makes a rewrite fail; at worst it has no effect.
 
 If a file index is configured on a forwarded column, that column alone is
 materialized so the index still sees its values; the other columns stay encoded.
 
-Passthrough changes what the rewrite costs, not what it produces, with one
-exception worth knowing: a Parquet column chunk can only carry one dictionary,
-so when the input files supply different dictionaries the output column keeps
-the first and falls back to plain encoding for the rest of the row group. The
-rewritten data is unchanged either way, but the output file may be larger than a
-rewrite that rebuilt a single dictionary from materialized values. Set
-``parquet.read.enable-dictionary-passthrough`` to ``false`` on the table to turn
-the optimization off and always rebuild.
+Trade-off
+^^^^^^^^^
+Passthrough changes what the rewrite costs, not what it produces. The rewritten
+data is identical either way, but the output file can be **larger**.
+
+A Parquet column chunk can carry only one dictionary. The writer keeps the first
+dictionary a column presents in a row group and falls back to plain encoding for
+the rest of that row group as soon as a different one arrives. Compaction merges
+several input files, each with its own dictionary, and output row groups are cut
+by ``parquet.block.size`` and by the writer's memory limit, which are not aligned
+to input file boundaries - a boundary may coincide, but nothing arranges for it.
+So a rewrite can keep the first input file's dictionary and write the rest of the
+row group plain, where materializing and rebuilding would have hashed the values
+into a single dictionary for that row group (up to
+``parquet.dictionary.page.size``, past which the writer falls back to plain in
+either case). Where that happens the output column also becomes ineligible for
+passthrough on the next compaction round, since its data pages are then only
+partly dictionary-encoded.
+
+How often it happens, and what it costs on a given table, is what the
+measurement above is for.
+
+Passthrough is therefore worth enabling when the reduction in compaction CPU
+matters more than output size - for example when the same dictionary recurs
+across input files, when the columns are wide enough that materializing them
+dominates, or when the output is short-lived. Leave it off otherwise.
 
 Append-Only Table Compaction Options
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/user_guide/compaction.rst
+++ b/docs/source/user_guide/compaction.rst
@@ -126,6 +126,11 @@ the values again, or when variant/map shredding is configured because those
 writers reshape each batch against a fixed physical schema. Setting the option
 therefore never makes a rewrite fail; at worst it has no effect.
 
+When the option is vetoed, the rewrite also enforces the veto on every input
+batch. A format reader can independently hand over dictionary-encoded columns
+because of its own lazy-decoding setting, so the rewrite decodes those columns
+before handing them to a writer that cannot accept dictionary arrays.
+
 If a file index is configured on a forwarded column, that column alone is
 materialized so the index still sees its values; the other columns stay encoded.
 
--- a/src/paimon/common/utils/arrow/arrow_utils.cpp
+++ b/src/paimon/common/utils/arrow/arrow_utils.cpp
@@ -44,7 +44,7 @@ namespace {
 // Whether `type` is a dictionary this can carry across the C data interface unchanged. The index
 // width is part of the test because nothing in a layout reveals it; see
 // ArrowUtils::IsDictionaryLayoutRecoverableValueType().
-bool IsResolvableDictionary(const arrow::DataType& type) {
+bool IsDictionaryLayoutRecoverable(const arrow::DataType& type) {
     if (type.id() != arrow::Type::DICTIONARY) {
         return false;
     }
@@ -558,7 +558,7 @@ Result<arrow::Compression::type> ArrowUtils::GetCompressionType(const std::strin
 // `int32` indices over `int32` offsets would silently reinterpret both buffers instead of failing.
 //
 // This narrows what may be carried; it cannot verify what was. See
-// ResolveParquetDictionaryStructType() for where the index width becomes a caller contract.
+// ResolveDictionaryStructTypeFromLayout() for where the index width becomes a caller contract.
 bool ArrowUtils::IsDictionaryLayoutRecoverableValueType(const arrow::DataType& type) {
     return arrow::is_binary_like(type.id());
 }
@@ -574,7 +574,7 @@ bool ArrowUtils::IsDictionaryLayoutRecoverableValueType(const arrow::DataType& t
 // batch whose dictionaries the schema does not declare, and it honours the contract by running
 // FlattenUnresolvableDictionaries() first. Closing the hole instead of narrowing it needs the real
 // `ArrowSchema` to reach the writer, which `FormatWriter::AddBatch(ArrowArray*)` drops.
-Result<std::shared_ptr<arrow::DataType>> ArrowUtils::ResolveParquetDictionaryStructType(
+Result<std::shared_ptr<arrow::DataType>> ArrowUtils::ResolveDictionaryStructTypeFromLayout(
     const std::shared_ptr<arrow::DataType>& logical_type, const ::ArrowArray* batch) {
     if (batch == nullptr || logical_type->id() != arrow::Type::STRUCT ||
         batch->n_children != logical_type->num_fields()) {
@@ -619,7 +619,8 @@ Result<std::shared_ptr<arrow::DataType>> ArrowUtils::ResolveParquetDictionaryStr
 
 Result<std::shared_ptr<arrow::StructArray>> ArrowUtils::FlattenUnresolvableDictionaries(
     const std::shared_ptr<arrow::StructArray>& batch,
-    const std::shared_ptr<arrow::DataType>& logical_type, arrow::MemoryPool* pool) {
+    const std::shared_ptr<arrow::DataType>& logical_type, arrow::MemoryPool* pool,
+    bool preserve_layout_recoverable_dictionaries) {
     const std::shared_ptr<arrow::DataType>& batch_type = batch->type();
     if (logical_type->id() != arrow::Type::STRUCT || !HasDictionary(*batch_type)) {
         return batch;
@@ -629,7 +630,13 @@ Result<std::shared_ptr<arrow::StructArray>> ArrowUtils::FlattenUnresolvableDicti
     arrow::FieldVector fields = batch_type->fields();
     for (int32_t i = 0; i < batch_type->num_fields(); ++i) {
         std::shared_ptr<arrow::Field> field = fields[i];
-        if (IsResolvableDictionary(*field->type()) || !HasDictionary(*field->type())) {
+        if (!HasDictionary(*field->type())) {
+            continue;
+        }
+        // Surviving the export is not enough when the destination imports against the logical
+        // type: an undeclared dictionary child of any shape then fails on the buffer count.
+        if (preserve_layout_recoverable_dictionaries &&
+            IsDictionaryLayoutRecoverable(*field->type())) {
             continue;
         }
         std::shared_ptr<arrow::Field> logical_field =
--- a/src/paimon/common/utils/arrow/arrow_utils.cpp
+++ b/src/paimon/common/utils/arrow/arrow_utils.cpp
@@ -24,6 +24,9 @@
 #include "arrow/array/concatenate.h"
 #include "arrow/array/util.h"
 #include "arrow/buffer.h"
+#include "arrow/c/abi.h"
+#include "arrow/compute/cast.h"
+#include "arrow/compute/exec.h"
 #include "arrow/type_traits.h"
 #include "arrow/util/bit_util.h"
 #include "arrow/util/bitmap_ops.h"
@@ -37,6 +40,55 @@
 namespace paimon {
 
 namespace {
+
+// Whether `type` is a dictionary this can carry across the C data interface unchanged. The index
+// width is part of the test because nothing in a layout reveals it; see
+// ArrowUtils::IsParquetDictionaryValueType().
+bool IsResolvableDictionary(const arrow::DataType& type) {
+    if (type.id() != arrow::Type::DICTIONARY) {
+        return false;
+    }
+    const auto& dictionary_type = checked_cast<const arrow::DictionaryType&>(type);
+    return dictionary_type.index_type()->id() == arrow::Type::INT32 &&
+           ArrowUtils::IsParquetDictionaryValueType(*dictionary_type.value_type());
+}
+
+// Whether `type` is or contains a dictionary at any depth.
+bool HasDictionary(const arrow::DataType& type) {
+    if (type.id() == arrow::Type::DICTIONARY) {
+        return true;
+    }
+    for (const std::shared_ptr<arrow::Field>& field : type.fields()) {
+        if (HasDictionary(*field->type())) {
+            return true;
+        }
+    }
+    return false;
+}
+
+// Whether any descendant of `array` carries a dictionary that `type` does not declare. `array`
+// itself is not examined; its caller has already handled the top level.
+bool HasUndeclaredDictionaryChild(const std::shared_ptr<arrow::DataType>& type,
+                                  const ::ArrowArray* array) {
+    if (array == nullptr || array->n_children != type->num_fields()) {
+        return false;
+    }
+    for (int64_t i = 0; i < array->n_children; ++i) {
+        const ::ArrowArray* child = array->children[i];
+        if (child == nullptr) {
+            continue;
+        }
+        const std::shared_ptr<arrow::DataType>& child_type =
+            type->field(static_cast<int>(i))->type();
+        if (child->dictionary != nullptr && child_type->id() != arrow::Type::DICTIONARY) {
+            return true;
+        }
+        if (HasUndeclaredDictionaryChild(child_type, child)) {
+            return true;
+        }
+    }
+    return false;
+}
 
 bool NeedsNormalization(const std::shared_ptr<arrow::ArrayData>& data) {
     if (data->offset != 0) {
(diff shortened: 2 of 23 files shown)


lucasfang:
LGTM

docs/source/user_guide/compaction.rst
```diff
@@ -88,6 +88,40 @@ After compaction, if the last output file is still smaller than
...
+fell back to plain encoding therefore does not qualify, even though it still
+carries a dictionary page. Passthrough is also skipped when the table writes a
+format other than Parquet, when ``parquet.enable-dictionary`` is ``false``
+because the writer would only expand the values again, or when variant/map
+shredding is configured because those writers reshape each batch against a fixed
+physical schema.
+
+If a file index is configured on a forwarded column, that column alone is
+materialized so the index still sees its values; the other columns stay encoded.
+
```
lxy-9602, line 136: I suggest keeping automatic enablement of `parquet.read.enable-dictionary-passthrough` disabled for compaction initially. Users could enable it after confirming a benefit with benchmarks representative of their workloads.

Velox has its own Parquet implementation, whose behavior is not necessarily consistent with Arrow Parquet. Therefore, before enabling this optimization by default, I would like to see before-and-after results from Paimon C++ using representative production data, including compaction time, CPU usage, peak memory, and output file size.

I am also concerned about a possible side effect related to output row-group boundaries. In the current Parquet write path, row groups are split not only according to the configured row-group size and row count, but also according to the writer’s memory limit.

For example, suppose `file0` and `file1` each contain three row groups. If compaction produces an output file with six aligned and independent row groups, each output column chunk could reuse this PR’s dictionary passthrough optimization as expected. However, the output row-group boundaries may not align with the input boundaries—for example, compaction might produce four row groups instead. In that case, one output row group may contain batches carrying different dictionaries from multiple input row groups or files. Arrow may retain the first dictionary and fall back to PLAIN encoding for the remaining data in that column chunk. Potentially, multiple output row groups could therefore become only partially dictionary-encoded, making the column ineligible for subsequent dictionary passthrough and possibly increasing the output file size.

src/paimon/common/utils/arrow/arrow_utils.h
```diff
@@ -69,6 +71,84 @@ class PAIMON_EXPORT ArrowUtils {
...
+    /// to be decoded while the type is still known.
+    ///
+    /// @param batch The batch to decode, matched to `logical_type` by field name; a column with no
+    ///              matching field is left alone.
+    /// @param logical_type The struct type the decoded columns are cast to. `batch` is returned
+    ///                     unchanged when it is not a struct.
+    /// @param pool Allocates the decoded columns. Only used when a column is actually decoded.
+    /// @return `batch` itself when nothing had to be decoded, otherwise a copy of it with the
+    ///         offset, length and validity of the original and the decoded columns swapped in.
+    static Result<std::shared_ptr<arrow::StructArray>> FlattenUnresolvableDictionaries(
```
lxy-9602, line 132: Could we simplify the comments a bit? They’re quite dense right now and getting hard to follow.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -484,4 +536,95 @@ Result<arrow::Compression::type> ArrowUtils::GetCompressionType(const std::strin
     return compression_type;
 }
 
+bool ArrowUtils::IsParquetDictionaryValueType(const arrow::DataType& type) {
+    return type.id() == arrow::Type::STRING || type.id() == arrow::Type::BINARY;
+}
```
lxy-9602, line 564: Could we use a more general function name here and avoid mentioning Parquet — something like `IsBinaryType`? Judging from the implementation alone, it doesn’t really seem Parquet-specific.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -484,4 +536,95 @@ Result<arrow::Compression::type> ArrowUtils::GetCompressionType(const std::strin
...
+            // Copy once, on the first column that has to be decoded: the parent keeps its offset,
+            // length and validity, and only the child data is swapped underneath it.
+            data = batch->data()->Copy();
+        }
+        // Decode the whole child rather than the slice the parent exposes, so the replacement
+        // lines up with the offset and length the parent still carries.
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            arrow::Datum decoded,
+            arrow::compute::Cast(arrow::MakeArray(data->child_data[i]), logical_field->type(),
+                                 arrow::compute::CastOptions::Safe(), &exec_context));
```
lxy-9602: May prefer `CastingUtils::Cast`.

src/paimon/core/io/data_file_index_writer.cpp
```diff
@@ -100,10 +104,40 @@ Status DataFileIndexWriter::AddBatch(const std::shared_ptr<arrow::StructArray>&
     if (finished_) {
         return Status::Invalid("Data file index writer has already finished");
     }
+    // Buffers allocated through the adaptor keep a raw pointer to it, so it has to outlive every
+    // array decoded below. Built on first use, since most batches decode nothing.
+    std::unique_ptr<arrow::MemoryPool> arrow_pool;
+    // One entry per indexed column, not per index: a column carrying both a bitmap and a bloom
```
lxy-9602, line 107: I’m a bit concerned that using a temporary `arrow_pool` here is somewhat risky — future changes could easily break its lifetime assumptions. Would it make sense to turn it into a member variable instead?

src/paimon/core/operation/append_only_file_store_write.cpp
```diff
@@ -278,9 +299,25 @@ Result<AppendOnlyFileStoreWrite::WriterFactory> AppendOnlyFileStoreWrite::GetDat
...
+    const std::shared_ptr<ShreddingWritePlanFactory>& plan_factory) const {
+    std::shared_ptr<FileFormat> file_format = options_.GetFileFormat();
+    if (!file_format || file_format->Identifier() != "parquet") {
+        return false;
+    }
+    PAIMON_ASSIGN_OR_RAISE(
+        bool enable_dictionary,
+        OptionsUtils::GetValueFromMap<bool>(options_.ToMap(), parquet::PARQUET_ENABLE_DICTIONARY,
+                                            ::parquet::DEFAULT_IS_DICTIONARY_ENABLED));
+    if (!enable_dictionary) {
```
lxy-9602, line 330: Could we use string literals directly here? Some engines have reimplemented the Parquet format layer, and I’m not sure their plugins still expose symbols like `PARQUET_ENABLE_DICTIONARY` and `DEFAULT_IS_DICTIONARY_ENABLED`.

src/paimon/format/parquet/parquet_format_writer.cpp
```diff
@@ -74,6 +81,51 @@ Status ParquetFormatWriter::AddBatch(ArrowArray* batch) {
...
+        std::shared_ptr<arrow::DataType> batch_type,
+        ArrowUtils::ResolveParquetDictionaryStructType(logical_struct_type_, batch));
+    if (batch_type == logical_struct_type_) {
+        return schema_;
+    }
+    if (dictionary_batch_type_ == nullptr || !dictionary_batch_type_->Equals(*batch_type)) {
+        dictionary_batch_type_ = batch_type;
+        dictionary_batch_schema_ = arrow::schema(batch_type->fields(), schema_->metadata());
+    }
+    return dictionary_batch_schema_;
```
lxy-9602: It looks like `dictionary_batch_schema_` and `dictionary_batch_type_` were made member variables mainly to avoid repeated `arrow::schema` conversion overhead. Have we already identified that as a clear hotspot? I feel `ResolveBatchSchema` could just return directly. Also, if the dictionary mode can switch, the benefit of caching these as member variables seems fairly limited.

src/paimon/format/parquet/parquet_format_writer.cpp
```diff
@@ -74,6 +81,51 @@ Status ParquetFormatWriter::AddBatch(ArrowArray* batch) {
...
+        }
+        if (columns.empty()) {
+            columns = record_batch->columns();
+            fields = record_batch->schema()->fields();
+        }
+        const auto& dictionary_type = checked_cast<const arrow::DictionaryType&>(*column->type());
+        PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+            arrow::Datum flattened,
+            arrow::compute::Cast(column, dictionary_type.value_type(),
+                                 arrow::compute::CastOptions::Safe(), &exec_context));
```
lxy-9602: `CastingUtils` maybe better

test/inte/append_compaction_inte_test.cpp
```diff
@@ -822,4 +824,239 @@ TEST_F(AppendCompactionInteTest, TestAppendTableCompactionWithIOException) {
...
+        arrow::field("b", arrow::binary()), arrow::field("u", arrow::utf8())};
+    auto schema = arrow::schema(fields);
+
+    std::map<std::string, std::string> options = {
+        {Options::FILE_FORMAT, file_format},
+        {Options::BUCKET, "1"},
+        {Options::BUCKET_KEY, "id"},
+        {Options::FILE_SYSTEM, "local"},
+        {"orc.read.enable-lazy-decoding", "true"},
+        // Above the distinct/total ratio of every column here, so ORC dictionary-encodes rather
```
lxy-9602, line 899: Thanks a lot for also fixing the orthogonal compaction bug in ORC when `orc.read.enable-lazy-decoding` is enabled. We’ll follow up with a corresponding fix for PK tables as well.

docs/source/user_guide/compaction.rst
SteNicholas: Agreed on both counts, and both are now in the branch.

The rewrite no longer enables anything. `parquet.read.enable-dictionary-passthrough` is off by default and the compaction path only ever *vetoes* it (`GetDictionaryPassthroughVetoReason()`); turning it on is the table's decision, taken after measuring. A veto is logged at DEBUG with its reason, but only for a table that did ask for the passthrough, so an ordinary rewrite says nothing.

Your row-group point is exactly right and I have written it into the docs rather than leaving it implicit. Output row groups are cut by `parquet.block.size` and by the writer's memory limit, which are not aligned to input file boundaries — a boundary may coincide, but nothing arranges for it. So a rewrite can keep the first input file's dictionary and write the rest of the row group plain, and that column then also loses eligibility for the next compaction round. `compaction.rst` now has a "Trade-off" subsection saying so, and telling users to compare output file size as well as compaction time before enabling it. `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthrough` asserts the compacted column really has fallen back to plain, and `...DefaultOff` asserts the same rewrite with the option off produces a column that is dictionary-encoded end to end — the two together pin the cost rather than describing it.

On evidence: the benchmark pair now writes the *same logical column* (the changing case rotates one alphabet and shifts the indices back, so only the dictionary object differs), which makes its delta attributable to the fallback. But those are format-writer microbenchmarks and I have said so in `benchmark.rst` — they are not the production compaction time / CPU / peak memory / output size you asked for. That measurement is still outstanding and the option ships off by default until it exists.

src/paimon/common/utils/arrow/arrow_utils.h
SteNicholas: Done — the header now carries the contract and nothing else; the Arrow/Parquet behaviour that explains it moved to the definitions, and the user-facing half to `compaction.rst`. This block went from 57 comment lines to 42, and the two long paragraphs on the index-width contract are no longer repeated in both the header and the `.cpp`.

I also removed the duplication that had built up elsewhere: the reason `BINARY` is excluded now lives in one place (the reader gate) plus one sentence in the user docs, and `parquet_format_defs.h` points at the gate instead of restating it. If any specific paragraph still reads as redundant, point at it and I will cut that one rather than doing another broad pass.

src/paimon/common/utils/arrow/arrow_utils.cpp
SteNicholas: You were right that the name over-promised, and chasing it turned up something worse, so this ended up as two predicates rather than one rename.

`ArrowUtils` is a generic utility, so it should not know about Parquet — that part is now `IsDictionaryLayoutRecoverableValueType()`, and it answers only "does `dictionary(int32, T)` survive the C data interface". It accepts `utf8` and `binary`, which is what the writer can recover and matches Velox's VARCHAR/VARBINARY. `large_utf8` stays out because a layout reports neither the index nor the offset width.

I did not go all the way to `IsBinaryType`, and the reason is the second predicate. While looking at this I found that no value accessor in the repository can read a `dictionary(int32, binary)`: `ColumnarUtils::GetView()` asserts on a dictionary whose values are neither `STRING` nor `LARGE_STRING` and returns an empty `string_view` in a release build, `LiteralConverter` rejects it, and `ColumnarBatchContext`'s fast path routes dictionaries back to `GetView()`. Every consumer here has only ever seen `STRING` dictionaries, because that is all the ORC reader produces under lazy decoding. Since the option is a *read* option and applies to every read of the table, that was a silent-empty-value path.

So the reader now forwards `STRING` alone, and it enforces that at the request site (`ResolveFullyDictionaryEncodedColumns()` requires `logical_type()->is_string()`, not just `BYTE_ARRAY`) — narrowing only the reporting side would have left Arrow emitting a `DictionaryArray` for a column the reader describes as `BINARY`. `ParquetFileBatchReaderTest.TestDictionaryPassthroughSkipsBinaryColumn` pins it with `f8`/`STRING` as the control, and `ParquetFormatWriterTest.TestWriteDictionaryOfBinaryColumn` pins that the writer keeps the `BINARY` capability so the reader's restriction does not leak into the format layer.

A name like `IsBinaryType` would have invited someone to add `LARGE_STRING` later — reasonable for a general type predicate, silently wrong here.

src/paimon/common/utils/arrow/arrow_utils.cpp
SteNicholas: Done — `CastingUtils::Cast()` here and in `ParquetFormatWriter::FlattenUnwritableDictionaries()`. It also drops the hand-built `arrow::compute::ExecContext` and the `arrow/compute/exec.h` include, since it takes the pool directly.

src/paimon/core/io/data_file_index_writer.cpp
SteNicholas: Good catch, and yes. `arrow_pool_` is now a member of `DataFileIndexWriter`, built on first use since most batches decode nothing.

Two details worth naming, since the hazard is easy to reintroduce. It is declared *before* `writers_`, so it is destroyed last: buffers allocated through the adaptor keep a raw pointer to it, and an index writer may still hold a decoded column. And the destructor is now out of line, so the header only needs a forward declaration of `arrow::MemoryPool`.

`DataFileIndexWriterTest.TestDictionaryEncodedIndexedColumnRoundTrip` now feeds two batches with different dictionaries and asserts the bitmap over both, so the pool has to survive across calls and each batch has to be decoded against its own alphabet.

src/paimon/core/operation/append_only_file_store_write.cpp
SteNicholas: Fair point, and it was a real layering break: `append_only_file_store_write.cpp` was the only file outside `src/paimon/format/parquet/` including `parquet_format_defs.h`, plus it pulled in `parquet/properties.h` for the default.

Both includes are gone. The option names and their defaults are now local `constexpr` in an anonymous namespace in that `.cpp`, so core links no Parquet symbols at all and a replacement Parquet plugin cannot break the build here. The file already compared the format identifier against a plain `"parquet"` literal, so this is consistent with what was there.

One small deviation from the literal suggestion: I used named constants rather than inlining the strings, because `parquet.read.enable-dictionary-passthrough` is needed in two functions and writing it twice invites a typo. They are string literals in the same translation unit, so the dependency is gone either way. Happy to inline them if you would rather.

src/paimon/format/parquet/parquet_format_writer.cpp
SteNicholas: No, it was not a measured hotspot — I had reasoned about it rather than profiled it, which is not a good enough reason for mutable state on the writer. Both members are gone and `ResolveBatchSchema()` builds and returns the schema directly; it is `const` now as a result.

Your second point is the stronger one: the cache was keyed on the batch type, so an alternating encoding would have rebuilt it on every batch anyway, and it added a way for the cached schema to drift from the batch it describes. Velox has to reconcile exactly that (it caches its Arrow schema for field-id fixups and then rewrites cached fields when a column's encoding changes across batches); rebuilding per batch means there is no such window here.

src/paimon/format/parquet/parquet_format_writer.cpp
SteNicholas: Done — `FlattenUnwritableDictionaries()` uses `CastingUtils::Cast()` with the writer's pool, so the local `arrow::compute::ExecContext` is gone here too.

test/inte/append_compaction_inte_test.cpp
SteNicholas: Thanks — and please do. The ORC path here only exercises it from the append side: with `orc.read.enable-lazy-decoding` on, the reader hands over `dictionary(int64, large_utf8)`, whose index and offset widths no `ArrowArray` layout can report, so `CompactRewrite` decodes it at the source through `FlattenUnresolvableDictionaries()` while the type is still known. `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthrough` is parameterised over ORC precisely to keep that path covered. If the PK fix needs the same decode-before-export step, that helper is reusable as is.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -33,11 +35,61 @@
...
+#include "paimon/core/casting/casting_utils.h"
 
 namespace paimon {
 
 namespace {
 
+// Whether `type` is a dictionary this can carry across the C data interface unchanged. The index
+// width is part of the test because nothing in a layout reveals it; see
+// ArrowUtils::IsDictionaryLayoutRecoverableValueType().
+bool IsResolvableDictionary(const arrow::DataType& type) {
```
lszskye: Some format(AliORC), will lazy decoding exposes dictionary-encoded STRING columns as dictionary<int32, utf8>. This type is considered resolvable by IsResolvableDictionary(), so FlattenUnresolvableDictionaries() keeps it encoded during append-only compaction.
However, the non-Parquet veto only disables parquet.read.enable-dictionary-passthrough; it does not disable aliorc.read.enable-lazy-decoding. As a result, an AliORC dictionary batch can still reach OrcFormatWriter::AddBatch(), where it is imported using the logical utf8 schema. The actual ArrowArray child is dictionary-encoded, so the import will fail. Even if the dictionary type were recovered during import, OrcAdapter::WriteBatch() does not support arrow::Type::DICTIONARY.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -500,4 +552,112 @@ Result<arrow::Compression::type> ArrowUtils::GetCompressionType(const std::strin
...
+// nothing here can tell `dictionary(int32(), utf8())` apart from `dictionary(int64(), utf8())`,
+// and the second would be read as the first.
+//
+// So the contract binds the producer, not the callers: `ParquetFormatWriter::ResolveBatchSchema`
+// and `DataFileWriterBase::AddFileIndexBatch` see only the layout.
+// `AppendOnlyFileStoreWrite::CompactRewrite` is today's only production path that can hand over a
+// batch whose dictionaries the schema does not declare, and it honours the contract by running
+// FlattenUnresolvableDictionaries() first. Closing the hole instead of narrowing it needs the real
+// `ArrowSchema` to reach the writer, which `FormatWriter::AddBatch(ArrowArray*)` drops.
+Result<std::shared_ptr<arrow::DataType>> ArrowUtils::ResolveParquetDictionaryStructType(
```
lszskye: See one function about parquet in framework is a little bit weird

src/paimon/common/utils/arrow/arrow_utils.cpp
SteNicholas: You are right, and the root of it is that the flatten was answering the wrong question. `FlattenUnresolvableDictionaries()` asked whether a dictionary survives the Arrow C data interface; what actually decides it is whether the writer receiving the batch recovers an encoding from a layout at all, and only `ParquetFormatWriter` does. Those are not the same question, and on `dictionary(int32, utf8)` they give opposite answers.

Fixed by passing that in rather than inferring it. `CompactRewrite` already computes the veto once, so it now derives `preserve_layout_recoverable_dictionaries` from it: under a veto every dictionary is decoded before the export, whatever its shape. That also closes the same hole on a second path — a table writing Parquet with `parquet.enable-dictionary` false is vetoed too, and a lazy-decoded `dictionary(int32, utf8)` would otherwise have been forwarded into the writer and written encoded, contradicting the option.

Two things worth being precise about. The lazy-decoding option is untouched, as you say: the veto still overrides only `parquet.read.enable-dictionary-passthrough`, and reaching across to disable another format's read option from the compaction path would be worse than decoding the batch. And this was never reachable from this repository — the built-in ORC reader widens to `dictionary(int64, large_utf8)`, which was already decoded for its shape, and the only producer of `dictionary(int32, utf8)` here is the Parquet reader, which the veto has already turned off. So it is a narrowed pre-existing gap rather than a regression, and it takes a reader like AliORC to hit.

Which is also the limit of the coverage I can give it. `ArrowUtilsTest.TestFlattenUnresolvableDictionaries` now pins the behaviour directly: the layout-recoverable `dictionary(int32, utf8)` is decoded when the destination resolves nothing, and a batch of only such columns is no longer returned by identity. But no end-to-end test here can tell the fix apart from the old behaviour, because no in-repo reader emits that shape into a non-Parquet writer. `AppendCompactionInteTest.TestAppendTableCompactionDictionaryPassthrough` on the ORC parameter does run the vetoed path end to end, it just cannot distinguish the two; I have written that into its comment rather than letting it look like coverage it is not. If you can run it against AliORC, that is the confirmation I am missing.

src/paimon/common/utils/arrow/arrow_utils.cpp
SteNicholas: Agreed, and it was a leftover from the previous round: `IsParquetDictionary...` became `IsDictionaryLayoutRecoverableValueType()` then, but `ResolveParquetDictionaryStructType()` right next to it did not, which left one feature carrying two vocabularies in one file.

It is now `ArrowUtils::ResolveDictionaryStructTypeFromLayout()`, and the file-local `IsResolvableDictionary()` became `IsDictionaryLayoutRecoverable()` so the whole path says "layout recoverable". Nothing in the body was ever Parquet-specific — it reads `ArrowArray::dictionary` and the caller's declared type. Arrow's Parquet reader under `set_read_dictionary` is just the producer that makes it necessary.

I did not move it into `src/paimon/format/parquet/`, which would be the other way to read your comment. `DataFileWriterBase::AddFileIndexBatch` calls it and is format-agnostic, so that would make core link Parquet symbols — the same layering break the earlier review had me remove from `append_only_file_store_write.cpp`.

lxy-9602 (approved):
+1

lszskye (approved):
+1

PR #258, opened 2026-08-29
fix(scan): preserve snapshot id for empty data evolution index plans

wangyong9999:
### Purpose

Linked issue: none.

A global-index miss returned an empty plan without the snapshot ID used by the scan.

**Internally evaluated index results:** retain the snapshot used by the Global Index and attach it to
an empty plan. For non-empty results, fail if latest advances before data planning instead of
combining index ranges and data splits from different snapshots.

**Externally supplied empty results:** return the shared empty plan with no snapshot ID. These
results carry no snapshot provenance, so the scan neither guesses one nor runs batch planning only
to obtain one.

**Time-travel boundary:** built-in Global Index evaluation explicitly rejects tag- and
timestamp-based startup modes. Explicit snapshot IDs and latest mode remain supported, including
the normal missing/invalid selector errors. Externally supplied results remain caller-controlled and
do not inherit this restriction.

Expected benefit: empty plans produced by internal index evaluation report the snapshot actually
used, while externally supplied results keep their previous lightweight behavior. There are no
shared-state, public API, ABI, storage-format, or protocol changes.

### Tests

- `paimon-global-index-test`: 138/138 passed.
- `paimon-core-test`: 1875/1875 passed.
- Regression coverage includes internal and externally supplied empty results, startup-mode
  precedence, missing/invalid snapshot selection, internal tag/timestamp rejection, external-result
  compatibility, and snapshot drift between internal index and data planning.
- clang-format 20.1.8, cpplint 2.0.2, codespell, and `git diff --check` passed.

### API and Format

None.

### Documentation

None.

### Generative AI tooling


src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -134,27 +144,41 @@ Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::WrapToIndexedSplits(
...
+    PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<GlobalIndexResult> result,
+                           index_scan_impl->Scan(predicate));
+    return std::optional<EvaluatedGlobalIndex>(EvaluatedGlobalIndex{result, snapshot_id.value()});
+}
+
+Result<std::optional<int64_t>> DataEvolutionBatchScan::ResolveGlobalIndexSnapshotId() const {
+    std::optional<int64_t> snapshot_id = core_options_.GetScanSnapshotId();
+    if (snapshot_id) {
+        return snapshot_id;
+    }
```
lxy-9602: Please validate the snapshot before returning the empty plan. With a pre-supplied empty `GlobalIndexResult`, a nonexistent `scan.snapshot-id` (for example, 999) is currently copied directly into a successful plan. Please resolve/load the snapshot here and add a regression test for nonexistent or expired snapshot IDs.

lxy-9602: This behavior seems to predate the PR: the global-index scan already appears to ignore tag- and timestamp-based time travel and fall back to the latest snapshot. What changes here is that the new resolver makes this more visible by attaching that latest snapshot ID to an empty plan. Since global-index time travel does not seem to be supported at the moment, would it make sense to reject these modes explicitly instead of silently scanning and reporting the latest snapshot?

wangyong9999: Addressed in ec761c57. A pre-supplied empty result now loads the selected snapshot before building the plan, so a nonexistent snapshot ID returns an error. Tag/timestamp options return `NotImplemented` when a Global Index result is used; predicates without a usable index still fall back to the normal batch scan. Regression tests cover both cases.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -42,25 +43,36 @@ DataEvolutionBatchScan::DataEvolutionBatchScan(
...
+                global_index_snapshot_id.value()));
+        }
+        return std::make_shared<PlanImpl>(global_index_snapshot_id,
+                                          std::vector<std::shared_ptr<Split>>());
     }
-    PAIMON_ASSIGN_OR_RAISE(RowRangeIndex row_range_index,
-                           RowRangeIndex::Create(row_ranges.value()));
+    PAIMON_ASSIGN_OR_RAISE(RowRangeIndex row_range_index, RowRangeIndex::Create(row_ranges));
     batch_scan_->WithRowRangeIndex(row_range_index);
     PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<Plan> data_plan, batch_scan_->CreatePlan());
```
wangyong9999, line 85: The index side now resolves its snapshot explicitly, but the data scan here still resolves latest on its own. A commit landing between the two applies row ranges computed at snapshot N to a data plan at N+1: matches added in N+1 are missed and the plan reports N+1 while the selection came from N. The race predates this PR, but with the resolved id now in hand it is cheap to close: when global_index_snapshot_id is set, compare data_plan->SnapshotId() against it and fail on mismatch (or pin the data scan to that id if the API allows).

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -44,25 +45,44 @@ DataEvolutionBatchScan::DataEvolutionBatchScan(
...
-    } else {
-        PAIMON_ASSIGN_OR_RAISE(row_ranges, final_global_index_result->ToRanges());
     }
-    if (!row_ranges) {
+    if (!final_global_index_result) {
         return batch_scan_->CreatePlan();
     }
-    if (row_ranges.value().empty()) {
-        return PlanImpl::EmptyPlan();
+    if (core_options_.GetScanTagName() || core_options_.GetScanTimestampMillis()) {
```
wangyong9999: Please resolve this from the effective `scan.mode`, not raw option presence. With `scan.mode=latest` plus a stale `scan.snapshot-id`, the normal data scan uses latest but the empty path below reports the stale ID. `from-snapshot` without a selector also falls back to latest instead of failing, and an ignored tag/timestamp can trigger this rejection. Reuse the same resolution and validation as `CreateStartingScanner`.

wangyong9999: Fixed in f6b9cb0a. Snapshot selection now follows the effective StartupMode, including selector precedence and missing-selector errors. Explicit IDs reuse the normal batch range check, and ignored selectors no longer trigger NotImplemented.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
wangyong9999: Fixed in f6b9cb0a. The data plan now checks its snapshot against the one used by the internal index scan and fails if latest moved between the two.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -44,27 +90,43 @@ DataEvolutionBatchScan::DataEvolutionBatchScan(
...
-    PAIMON_ASSIGN_OR_RAISE(RowRangeIndex row_range_index,
-                           RowRangeIndex::Create(row_ranges.value()));
+    PAIMON_ASSIGN_OR_RAISE(std::vector<Range> row_ranges, final_global_index_result->ToRanges());
+    if (row_ranges.empty()) {
+        if (!global_index_snapshot_id) {
+            const std::shared_ptr<SnapshotManager>& snapshot_manager =
+                snapshot_reader_->GetSnapshotManager();
+            PAIMON_ASSIGN_OR_RAISE(std::optional<Snapshot> snapshot,
+                                   ResolveGlobalIndexScanSnapshot(core_options_, snapshot_manager));
+            if (!snapshot) {
```
lxy-9602: If `global_index_result_` is set externally, how do we infer its snapshot ID? In that case, using `ResolveGlobalIndexScanSnapshot` does not seem very appropriate.

I’d suggest being stricter here: if predicates are present with global index, we should either reject time travel options such as timestamp/tag, or properly implement the corresponding support like java, rather than adding defensive checks in multiple places.

wangyong9999: Fixed in 74003307. Supplied empty results now take the selected snapshot from normal batch planning; we no longer infer it through a separate resolver. Tag/timestamp are rejected before evaluating a Global Index candidate, and the shared scanner change is removed.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
wangyong9999: Updated in 74003307. Supplied empty results now use normal batch planning, so snapshot validation stays on the regular path. Tag/timestamp are rejected as soon as a Global Index path is requested, including predicates without an applicable index.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -44,27 +59,43 @@ DataEvolutionBatchScan::DataEvolutionBatchScan(
...
+    PAIMON_ASSIGN_OR_RAISE(std::vector<Range> row_ranges, final_global_index_result->ToRanges());
+    if (row_ranges.empty()) {
+        if (global_index_snapshot_id) {
+            return std::make_shared<PlanImpl>(global_index_snapshot_id,
+                                              std::vector<std::shared_ptr<Split>>());
+        }
+
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<Plan> data_plan, batch_scan_->CreatePlan());
+        return std::make_shared<PlanImpl>(data_plan->SnapshotId(),
+                                          std::vector<std::shared_ptr<Split>>());
```
lxy-9602, line 81: Are we sure about this? Do we still need to run a batch scan on an empty result just to get the snapshot ID? It seems the snapshot should be `null` here, since we cannot determine the snapshot of an externally provided `global_index_result`.

wangyong9999: Agreed. Fixed in 9d8746c9. An externally supplied empty result now returns the empty plan directly with no associated snapshot. Only results evaluated inside this scan retain the snapshot used for index evaluation.

test/inte/global_index_test.cpp
```diff
@@ -1504,6 +1507,98 @@ TEST_P(GlobalIndexTest, TestDataEvolutionBatchScan) {
...
+                                                {{Options::SCAN_SNAPSHOT_ID, "1"},
+                                                 {Options::SCAN_TAG_NAME, "ignored"},
+                                                 {Options::SCAN_TIMESTAMP_MILLIS, "0"}},
+                                                empty_index_result));
+    ASSERT_TRUE(supplied_explicit_plan->Splits().empty());
+    ASSERT_FALSE(supplied_explicit_plan->SnapshotId());
+
+    Result<std::shared_ptr<Plan>> missing_selector_result =
+        ScanGlobalIndexAndData(table_path, predicate, {{Options::SCAN_MODE, "from-snapshot"}});
+    ASSERT_TRUE(missing_selector_result.status().IsInvalid())
```
lxy-9602: One minor suggestion: could you please use `ASSERT_NOK_WITH_MSG` here?

wangyong9999: Updated in 008e5679. The test now uses `ASSERT_NOK_WITH_MSG` and checks the missing-selector message.

lxy-9602 (dismissed):
+1

test/inte/global_index_test.cpp
```diff
@@ -1504,6 +1507,97 @@ TEST_P(GlobalIndexTest, TestDataEvolutionBatchScan) {
...
+
+    ASSERT_OK_AND_ASSIGN(
+        std::shared_ptr<Plan> supplied_nonexistent_snapshot_plan,
+        ScanGlobalIndexAndData(table_path, /*predicate=*/nullptr,
+                               {{Options::SCAN_SNAPSHOT_ID, "999"}}, empty_index_result));
+    ASSERT_TRUE(supplied_nonexistent_snapshot_plan->Splits().empty());
+    ASSERT_FALSE(supplied_nonexistent_snapshot_plan->SnapshotId());
+
+    ASSERT_NOK(ScanGlobalIndexAndData(table_path, predicate, {{Options::SCAN_SNAPSHOT_ID, "999"}}));
+}
```
lxy-9602, line 1601: It seems there may be a missed update here.

wangyong9999: Fixed in 59a5e8bd. This assertion now uses `ASSERT_NOK_WITH_MSG` and checks the nonexistent snapshot error.

lxy-9602 (approved):
+1 Appreciate your patience and the adjustments!

PR #263, opened 2026-08-31
fix: align format options and listagg semantics with Java

lxy-9602:
### Purpose

This change aligns several format-related behaviors with the latest Java Paimon implementation:

- Keep legacy manifest formats readable, but allow only Avro for manifest writes.
  - `manifest.format` is retained as a deprecated compatibility option.
  - Writing with a non-Avro manifest format now returns an explicit error.
  - Read paths continue to use the configured legacy format.
- Convert the `parquet/append_09` test fixture manifests from ORC to Avro while preserving the existing manifest filenames and snapshots.
- Rename the Parquet dictionary option from `parquet.enable-dictionary` to `parquet.enable.dictionary`, matching Java Paimon.
- Align `listagg` blank-string handling with Java:
  - Add `StringUtils::IsBlank` with Java `Character.isWhitespace` semantics.
  - Ignore empty, whitespace-only, and supported Unicode blank inputs.
  - Prevent blank accumulators from producing a leading delimiter.
  - Ignore blank tokens when `distinct` is enabled.

### Tests

- read and write access for Avro and legacy manifest formats
- rejection of non-Avro manifest writes
- disabling Parquet dictionary encoding with `parquet.enable.dictionary`
- ASCII and Unicode behavior of `StringUtils::IsBlank`
- `listagg` handling of blank inputs, accumulators, and distinct tokens
- commit tests using the Avro-converted `parquet/test_data` fixture

### API and Format
Behavioral and format-related changes:

- Remove `manifest.format` option.
- Avro is now the only writable manifest format.
- Existing non-Avro manifests remain readable.
- The Parquet dictionary option is renamed to `parquet.enable.dictionary`.

### Generative AI tooling
Generated-by: Codex (GPT-5)


src/paimon/common/utils/string_utils_test.cpp
```diff
@@ -209,6 +210,21 @@ TEST_F(StringUtilsTest, TestIsNullOrWhitespaceOnly) {
...
+        "",     " ",        "   ",      "\t",       "\n",       "\r",
+        "\r\n", " \t\n\r ", u8"\u1680", u8"\u2000", u8"\u3000", u8" \t\u3000\u2000\n"};
+    for (const std::string& blank : blank_strings) {
+        ASSERT_TRUE(StringUtils::IsBlank(blank)) << blank;
+    }
+
+    ASSERT_FALSE(StringUtils::IsBlank("user1"));
+    ASSERT_FALSE(StringUtils::IsBlank(" user1 "));
+    ASSERT_FALSE(StringUtils::IsBlank(u8"\u00a0"));
+    ASSERT_FALSE(StringUtils::IsBlank(std::string("\xc0\x80", 2)));
```
dalingmeng, line 227: \u00a0 covers the non-breaking case nicely. Could we add the ones the implementation is structurally most exposed on? IsJavaWhitespace skips U+2007 by splitting 0x2000–0x200a into 0x2000–0x2006 + 0x2008–0x200a, but the test only exercises \u2000 — merging those two ranges back into one would keep every test green while silently diverging from Character.isWhitespace, and IsNullOrWhitespaceOnly would carry that into partition path computation.
> ```suggestion
>     ASSERT_FALSE(StringUtils::IsBlank(std::string("\xc0\x80", 2)));
>     // Non-breaking or otherwise excluded by Character.isWhitespace.
>     ASSERT_FALSE(StringUtils::IsBlank(u8"\u2007"));  // FIGURE SPACE
>     ASSERT_FALSE(StringUtils::IsBlank(u8"\u202f"));  // NARROW NO-BREAK SPACE
>     ASSERT_FALSE(StringUtils::IsBlank(u8"\u0085"));  // NEL
>     ASSERT_FALSE(StringUtils::IsBlank(u8"\u180e"));  // not whitespace since Java 8
> ```
>

src/paimon/common/utils/string_utils.cpp
```diff
@@ -85,18 +93,54 @@ bool StringUtils::EndsWith(const std::string& str, const std::string& suffix) {
...
+        if (!IsJavaWhitespace(code_point)) {
             return false;
         }
+        offset += length;
     }
     return true;
 }
 
+bool StringUtils::IsNullOrWhitespaceOnly(const std::string& str) {
+    return IsBlank(str);
```
dalingmeng, line 141: This function goes from std::isspace (ASCII only) to delegating to IsBlank (Unicode). The direction is right — Java's InternalRowPartitionComputer:86 uses the Unicode-aware isNullOrWhitespaceOnly, so C++ was genuinely out of line here.But the function has 12 callers under src/, and two of them change actual output rather than just tightening validation: binary_row_partition_computer.cpp:158 (a blank partition value now falls into the default partition, i.e. the partition directory changes) and jieba_analyzer.cpp:111 (blank terms get filtered out, so index content changes). That reaches beyond what the PR title covers, and the description doesn't mention it.Partition paths determine where data is written, so a change like that needs to be independently revertable — sitting in the same commit as the listagg fix means a partition-layout problem can only be rolled back by reverting listagg along with it. Splitting costs almost nothing: IsBlank is a new function, listagg can use it directly, and this PR just needs to not change that one delegation line.If you consider the risk negligible, keeping it here is fine too — but please state the fix in the PR description and add a Unicode-whitespace partition value case for BinaryRowPartitionComputer.

lucasfang:
LGTM

SteNicholas (approved):
LGTM.

PR #264, opened 2026-08-31
fix(catalog): validate identifier names used to build catalog paths

lucasfang:
### Purpose

`FileSystemCatalog` builds every path it touches from the database name and from the components parsed out of the table name, and it did so without checking that those names are usable as a single path component. The same holds for the branch name, which selects a directory under the table root. This PR adds that validation.

The rule itself lives in `PathUtil::CheckSinglePathComponent(kind, name)`, next to the other path helpers, so that both the catalog and the branch layer share one implementation. A name is rejected when it is empty or whitespace-only, is exactly `.` or `..`, contains `/` or `\`, or contains control characters. Validation is purely lexical, so it behaves identically for `local`, `oss://`, `hdfs://` and any other `FileSystem`, needs no extra IO and has no symlink TOCTOU semantics. Names that merely contain a dot stay valid (`my.db`, `a..b`), as do non-ascii names. A rejected name is escaped in the error message (`\n`, `\r`, `\t`, `\\`, `\x{xx}` for any other control byte), so it can neither add a line to the log the error is written to nor truncate the C string it is copied into.

`CatalogUtils::CheckValidDatabaseName` and `CheckValidTableName` delegate to that helper and are applied in `FileSystemCatalog::NewDatabasePath` and `FileSystemCatalog::NewDataTablePath`. Those two functions are the only place where catalog paths are built, so all entry points (create, drop, rename, get, `List*`, `*Exists`) are covered at once and future entry points inherit the check. `FileSystemCatalog::TableExists` and `LoadTableSchema` validate the identifier as a whole in addition, because for a system table name they resolve a rebuilt data table identifier that no longer carries the branch component. `RestCatalog` is out of scope: it resolves a location on the server and builds url-encoded request paths rather than file system paths, and its own name handling is tracked separately. One C++-specific difference from apache/paimon-rust (`validate_identifier_name`, PR #334): `NewDataTablePath` uses the parsed table name rather than the raw object string, so `CheckValidTableName` validates each parsed component (data table name, branch name, system table name).

A branch name is validated through `BranchManager::CheckValidBranch`, the single place that owns the branch semantics: a branch that `BranchManager::NormalizeBranch` maps to `main` (empty or whitespace-only) names no directory of its own and is accepted, anything else must stay a single path component. That check is applied everywhere a caller-supplied branch reaches `BranchManager::BranchPath`: the branch component of a table identifier, `FileSystemCatalog::ListSnapshots`, the `branch` and `scan.fallback-branch` options in `CoreOptions`, and `ReadContextBuilder::Finish` / `WriteContextBuilder::Finish` for the branch set through `WithBranch`. Without the last three, a branch such as `rt/../../../../../outside` still escaped the table root even though the catalog path was validated.

`BranchManager::BranchPath` now normalizes its argument as well, because it decided with the exact string comparison of `IsMainBranch` and therefore only behaved correctly for an already normalized branch. Every manager that builds a branch path (`SchemaManager`, `SnapshotManager`, `TagManager`, `ConsumerManager`) satisfied that implicit contract by normalizing in its own constructor, but `RealtimeCommitProperties::OffsetsDirectory` does not: for `branch = "   "` a commit wrote the real-time offsets under the main branch, since `FileStoreCommitImpl` passes the normalized `SnapshotManager::Branch()`, while `OrphanFilesCleanerImpl` passed the raw `CoreOptions::GetBranch()` and looked for them under `branch/branch-   /metadata`. Normalizing in the one function that turns a branch into a path keeps every producer and consumer on the same directory instead of adding a fifth copy of the same normalization.

Behavior change to be aware of: such names used to be accepted silently and now return `Status::Invalid` before any file system access happens.

### Tests

- `FileSystemCatalogTest.TestIdentifierNameValidationRules`: table-driven coverage of the rejected forms (`""`, `"   "`, `"."`, `".."`, a name with a slash, a name with a backslash, a name with `\n`, a name with `\0`), checked through both `CreateDatabase` and `CreateTable` against the expected error message. The same test asserts that `my.db`, `a..b` and `数据` remain creatable as databases and that `orders` and `订单` remain creatable as tables, so the rules are not over-tightened.
- `FileSystemCatalogTest.TestRejectInvalidNames`: asserts that a rejected name is refused by every catalog entry point (`CreateDatabase`, `DatabaseExists`, `ListTables`, `DropDatabase`, `CreateTable`, `TableExists`, `GetTableLocation`, `GetTable`, `DropTable`, `RenameTable`, `GetDatabaseLocation`, `ListSnapshots`, plus the branch component of a table name and a system table name that carries such a branch), that nothing is created or deleted on those paths, and that a legitimate table in the same warehouse is untouched.
- `PathUtilsTest.TestCheckSinglePathComponent` and `TestCheckSinglePathComponentEscapesRejectedName`: the rule set itself, and the escaping of a rejected name in the error message.
- `BranchManagerTest.TestCheckValidBranch`: a branch normalized to `main` is accepted, `..`, `rt/../../../../../outside` and a branch with a newline are rejected.
- `BranchManagerTest.TestBranchPath` and `RealtimeCommitPropertiesTest` offsets-directory assertions: an empty and a whitespace-only branch resolve to the table root and to the main offsets directory rather than to a `branch/branch-   ` directory.
- `CoreOptionsTest.TestRejectBranchLeavingTableRoot`, `ReadContextTest.TestRejectBranchLeavingTablePath` and `WriteContextTest.TestRejectBranchLeavingRootPath`: the same branch is rejected through the `branch` and `scan.fallback-branch` options and through `WithBranch` on both context builders.
- `RestCatalogTest.DatabaseOperations`: adjusted to the new `GetDatabaseLocation` signature, and still asserts that a database the server cannot resolve yields an empty location rather than an error.
- Local run: the whole `unittest` target passes, 28/28 test binaries, since the check now sits on the catalog, options and read/write context paths.

### API and Format

No storage format change. One public API signature change: `Catalog::GetDatabaseLocation` now returns `Result<std::string>` instead of `std::string`, so that a name that cannot form a location is reported as an error rather than as an empty string. This is source-incompatible for a caller that uses the returned string directly; the repository has no such caller outside the two implementations and their tests. `RestCatalog::GetDatabaseLocation` keeps returning an empty string for a database the server cannot resolve, which is now documented in `include/paimon/catalog/catalog.h`. Everything else is internal: `PathUtil::CheckSinglePathComponent` and `BranchManager::CheckValidBranch` are new internal helpers, `CatalogUtils::CheckValidDatabaseName` and `CheckValidTableName` are internal, and the private static `FileSystemCatalog::NewDatabasePath` changes from `std::string` to `Result<std::string>`.

### Documentation

No documentation change needed: this is a robustness fix rather than a new feature, and the only user-visible contract updates are the `GetDatabaseLocation` signature and doc comment described above.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/src/paimon/core/catalog/catalog_utils.cpp
+++ b/src/paimon/core/catalog/catalog_utils.cpp
@@ -18,13 +18,12 @@
 
 #include "paimon/core/catalog/catalog_utils.h"
 
-#include <algorithm>
-#include <cctype>
 #include <optional>
 
 #include "fmt/format.h"
 #include "paimon/catalog/catalog.h"
-#include "paimon/common/utils/string_utils.h"
+#include "paimon/common/utils/path_util.h"
+#include "paimon/core/utils/branch_manager.h"
 #include "paimon/result.h"
 
 namespace paimon {
@@ -34,27 +33,6 @@ namespace {
 Status SystemTableError(const Identifier& identifier, const std::string& action) {
     return Status::Invalid(fmt::format("Cannot '{}' for system table '{}', please use data table.",
                                        action, identifier.ToString()));
-}
-
-/// Rejects names that cannot be used as a single path component: such a name would make the
-/// path built from it escape the directory it is joined to.
-Status CheckValidIdentifierName(const std::string& kind, const std::string& name) {
-    const char* reason = nullptr;
-    if (StringUtils::IsNullOrWhitespaceOnly(name)) {
-        reason = "cannot be empty or whitespace";
-    } else if (name == "." || name == "..") {
-        reason = "cannot be '.' or '..'";
-    } else if (name.find('/') != std::string::npos || name.find('\\') != std::string::npos) {
-        reason = "cannot contain path separators";
-    } else if (std::any_of(name.begin(), name.end(), [](char c) {
-                   return std::iscntrl(static_cast<unsigned char>(c)) != 0;
-               })) {
-        reason = "cannot contain control characters";
-    }
-    if (reason != nullptr) {
-        return Status::Invalid(fmt::format("{} name {}: '{}'", kind, reason, name));
-    }
-    return Status::OK();
 }
 
 }  // namespace
@@ -95,30 +73,25 @@ Status CatalogUtils::CheckNotBranch(const Identifier& identifier, const std::str
 }
 
 Status CatalogUtils::CheckValidDatabaseName(const std::string& db_name) {
-    return CheckValidIdentifierName("database", db_name);
+    return PathUtil::CheckSinglePathComponent("database", db_name);
 }
 
 Status CatalogUtils::CheckValidTableName(const Identifier& identifier) {
     PAIMON_ASSIGN_OR_RAISE(std::string data_table_name, identifier.GetDataTableName());
-    PAIMON_RETURN_NOT_OK(CheckValidIdentifierName("table", data_table_name));
+    PAIMON_RETURN_NOT_OK(PathUtil::CheckSinglePathComponent("table", data_table_name));
     PAIMON_ASSIGN_OR_RAISE(std::optional<std::string> branch, identifier.GetBranchName());
     if (branch) {
-        PAIMON_RETURN_NOT_OK(CheckValidIdentifierName("branch", branch.value()));
+        // The branch of an identifier selects the same directory as the `branch` option, so both
+        // go through the same check.
+        PAIMON_RETURN_NOT_OK(BranchManager::CheckValidBranch(branch.value()));
     }
     PAIMON_ASSIGN_OR_RAISE(std::optional<std::string> system_table,
                            identifier.GetSystemTableName());
     if (system_table) {
-        PAIMON_RETURN_NOT_OK(CheckValidIdentifierName("system table", system_table.value()));
+        PAIMON_RETURN_NOT_OK(
+            PathUtil::CheckSinglePathComponent("system table", system_table.value()));
     }
     return Status::OK();
-}
-
-Status CatalogUtils::CheckValidBranchName(const std::string& branch) {
-    // An empty branch selects the main branch, see BranchManager::NormalizeBranch.
-    if (StringUtils::IsNullOrWhitespaceOnly(branch)) {
-        return Status::OK();
-    }
-    return CheckValidIdentifierName("branch", branch);
 }
 
 }  // namespace paimon
--- a/src/paimon/core/catalog/catalog_utils.cpp
+++ b/src/paimon/core/catalog/catalog_utils.cpp
@@ -18,10 +18,13 @@
 
 #include "paimon/core/catalog/catalog_utils.h"
 
+#include <algorithm>
+#include <cctype>
 #include <optional>
 
 #include "fmt/format.h"
 #include "paimon/catalog/catalog.h"
+#include "paimon/common/utils/string_utils.h"
 #include "paimon/result.h"
 
 namespace paimon {
@@ -31,6 +34,27 @@ namespace {
 Status SystemTableError(const Identifier& identifier, const std::string& action) {
     return Status::Invalid(fmt::format("Cannot '{}' for system table '{}', please use data table.",
                                        action, identifier.ToString()));
+}
+
+/// Rejects names that cannot be used as a single path component: such a name would make the
+/// path built from it escape the directory it is joined to.
+Status CheckValidIdentifierName(const std::string& kind, const std::string& name) {
+    const char* reason = nullptr;
+    if (StringUtils::IsNullOrWhitespaceOnly(name)) {
+        reason = "cannot be empty or whitespace";
+    } else if (name == "." || name == "..") {
+        reason = "cannot be '.' or '..'";
+    } else if (name.find('/') != std::string::npos || name.find('\\') != std::string::npos) {
+        reason = "cannot contain path separators";
+    } else if (std::any_of(name.begin(), name.end(), [](char c) {
+                   return std::iscntrl(static_cast<unsigned char>(c)) != 0;
+               })) {
+        reason = "cannot contain control characters";
+    }
+    if (reason != nullptr) {
+        return Status::Invalid(fmt::format("{} name {}: '{}'", kind, reason, name));
+    }
+    return Status::OK();
 }
 
 }  // namespace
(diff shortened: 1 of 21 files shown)


src/paimon/core/catalog/file_system_catalog.cpp
```diff
@@ -511,6 +522,7 @@ Status FileSystemCatalog::RenameTable(const Identifier& from_table, const Identi
 
 Result<std::vector<SnapshotInfo>> FileSystemCatalog::ListSnapshots(
     const Identifier& identifier, const std::string& branch) const {
+    PAIMON_RETURN_NOT_OK(CatalogUtils::CheckValidBranchName(branch));
     PAIMON_ASSIGN_OR_RAISE(bool exists, TableExists(identifier));
```
lxy-9602, line 525: **Branch validation does not cover all filesystem entry points**

This validation only protects `FileSystemCatalog::ListSnapshots` and branch names encoded in catalog identifiers. Other public branch inputs, including `ReadContextBuilder::WithBranch`, `WriteContextBuilder::WithBranch`, `Options::BRANCH`, and `SCAN_FALLBACK_BRANCH`, are still passed to `BranchManager::BranchPath` without path-safety validation.

For example, if a valid `rt` branch already exists, a branch value such as `rt/../../../../../outside` produces an invalid path.

src/paimon/core/catalog/file_system_catalog.cpp
```diff
@@ -123,7 +127,8 @@ Result<bool> FileSystemCatalog::TableExists(const Identifier& identifier) const
 }
 
 std::string FileSystemCatalog::GetDatabaseLocation(const std::string& db_name) const {
-    return NewDatabasePath(warehouse_, db_name);
+    // An invalid name has no valid location, keep the same convention as RestCatalog.
+    return NewDatabasePath(warehouse_, db_name).value_or("");
```
zjw1111: why not return `Result`?

zjw1111 (commented):
Three inline findings from the identifier validation review.

src/paimon/core/catalog/catalog_utils.cpp
```diff
@@ -33,6 +36,27 @@ Status SystemTableError(const Identifier& identifier, const std::string& action)
...
+Status CheckValidIdentifierName(const std::string& kind, const std::string& name) {
+    const char* reason = nullptr;
+    if (StringUtils::IsNullOrWhitespaceOnly(name)) {
+        reason = "cannot be empty or whitespace";
+    } else if (name == "." || name == "..") {
+        reason = "cannot be '.' or '..'";
+    } else if (name.find('/') != std::string::npos || name.find('\\') != std::string::npos) {
+        reason = "cannot contain path separators";
+    } else if (std::any_of(name.begin(), name.end(), [](char c) {
+                   return std::iscntrl(static_cast<unsigned char>(c)) != 0;
```
zjw1111: `StringUtils::IsNullOrWhitespaceOnly` and `std::iscntrl` classify individual UTF-8 bytes under the current C locale, while the referenced Rust implementation uses Unicode-aware `trim` and `char::is_control`. Under the typical C locale, a name consisting of U+2003 EM SPACE or containing U+0085 is accepted here but rejected by Rust. C++ can therefore create catalog objects that Rust cannot operate on, and the result also depends on process locale.

src/paimon/core/catalog/catalog_utils.cpp
```diff
@@ -33,6 +36,27 @@ Status SystemTableError(const Identifier& identifier, const std::string& action)
...
+               })) {
+        reason = "cannot contain control characters";
+    }
+    if (reason != nullptr) {
+        return Status::Invalid(fmt::format("{} name {}: '{}'", kind, reason, name));
```
zjw1111: The rejected `name` is formatted verbatim even when it contains control characters. A newline remains in the returned `Status` and can inject additional log lines when the status is propagated, while an embedded NUL can truncate downstream C-string output. The referenced Rust implementation escapes these characters through debug formatting.

src/paimon/core/catalog/catalog_utils.cpp
```diff
@@ -70,4 +94,31 @@ Status CatalogUtils::CheckNotBranch(const Identifier& identifier, const std::str
...
+Status CatalogUtils::CheckValidDatabaseName(const std::string& db_name) {
+    return CheckValidIdentifierName("database", db_name);
+}
+
+Status CatalogUtils::CheckValidTableName(const Identifier& identifier) {
+    PAIMON_ASSIGN_OR_RAISE(std::string data_table_name, identifier.GetDataTableName());
+    PAIMON_RETURN_NOT_OK(CheckValidIdentifierName("table", data_table_name));
+    PAIMON_ASSIGN_OR_RAISE(std::optional<std::string> branch, identifier.GetBranchName());
+    if (branch) {
+        PAIMON_RETURN_NOT_OK(CheckValidIdentifierName("branch", branch.value()));
```
zjw1111: This path sends an identifier branch through the generic validator, so a whitespace-only branch is rejected. `CheckValidBranchName` below and the existing `BranchManager::NormalizeBranch` both treat a whitespace-only branch as the main branch. Consequently, `ListSnapshots(id, "   ")` is accepted while the equivalent `t$branch_   ` identifier is rejected, producing inconsistent branch semantics between entry points.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -159,16 +159,16 @@ Status RestCatalog::DropDatabase(const std::string& name, bool ignore_if_not_exi
...
     if (CatalogUtils::IsSystemDatabase(db_name)) {
-        return "";
+        return std::string();
     }
     Result<GetDatabaseResponse> response = api_->GetDatabase(db_name);
     if (!response.ok()) {
         PAIMON_LOG_WARN(logger_, "failed to get location of database %s: %s", db_name.c_str(),
                         response.status().ToString().c_str());
-        return "";
+        return std::string();
```
zjw1111: already has result, still need empty string here?

lucasfang: This does not check whether the database actually exists.
An implementation that resolves the location on a server, such as the REST catalog, returns an empty string for a database it cannot resolve.

src/paimon/core/core_options.cpp
```diff
@@ -799,8 +799,14 @@ struct CoreOptions::Impl {
...
         // Parse scan.fallback-branch - fallback branch when partition not found
         PAIMON_RETURN_NOT_OK(parser.Parse(Options::SCAN_FALLBACK_BRANCH, &scan_fallback_branch));
+        if (scan_fallback_branch) {
+            PAIMON_RETURN_NOT_OK(BranchManager::CheckValidBranch(scan_fallback_branch.value()));
+        }
         // Parse branch - branch name, default "main"
         PAIMON_RETURN_NOT_OK(parser.Parse(Options::BRANCH, &branch));
+        // Both branches name a directory under the table root, so they must stay a single path
+        // component.
+        PAIMON_RETURN_NOT_OK(BranchManager::CheckValidBranch(branch));
```
zjw1111, line 809: [P2] A whitespace-only branch passes this validation but remains unnormalized in `CoreOptions`. `OrphanFilesCleanerImpl` later passes the raw value from `GetBranch()` to `RealtimeCommitProperties::OffsetsDirectory()`, which resolves it to `branch/branch-   /metadata` instead of the main branch's `metadata` directory. Consequently, orphan cleanup for `branch = "   "` skips the main branch's real-time offset files even though `CheckValidBranch` treats that value as main.

zjw1111 (commented):
Found one remaining issue in the whitespace-branch orphan-cleanup path.

zjw1111 (approved):
+1

lxy-9602 (approved):
+1

PR #265, opened 2026-08-31
perf(predicate): evaluate IN and NOT IN with arrow::compute::is_in

### LINKED ISSUE #262: [Feature] Optimize IN and NOT IN predicate evaluation with a dedicated literal lookup set
### Search before asking

### Motivation

Evaluating an `IN` / `NOT IN` predicate on a batch currently goes through `MultiLiteralsLeafFunction`, which materializes the whole column into `Literal` objects (one heap allocation per row, and non-owning string literals still require building the `Literal` wrapper) and then linearly scans all literals for every row. The cost is `O(rows × literals)` plus `O(rows)` heap allocations per batch. With large `IN` lists (e.g. thousands of partition keys pushed down from SQL), predicate evaluation becomes a measurable part of scan time even though membership testing only needs set lookups.

### Solution

Build an immutable, type-specialized lookup structure once when the predicate is constructed, and probe arrow arrays in `O(rows)` without any per-row allocation:

- Introduce `LiteralSet` in `src/paimon/common/predicate/`, created from the predicate literals at construction time and owned by `LeafPredicateImpl`. Unsupported types (FLOAT/DOUBLE with NaN semantics, TIMESTAMP with unit conversion, DECIMAL with cross-scale comparison) or heterogeneous literals make the construction return null, and every path falls back to the existing `In` / `NotIn` implementation, so observable behavior stays identical.
- Integer family (TINYINT/SMALLINT/INT/BIGINT/DATE, widened to int64): a dense bitmap when the value span is small and close to the literal count, otherwise a hash set, with a min/max range check rejecting out-of-range values before any lookup.
- BOOLEAN: two flags.
- STRING/BINARY: a hash set of `string_view`s backed by owned storage, with length-range and first-byte bitmap filters applied before hashing; dictionary-encoded arrays are probed once per dictionary plus `O(rows)` index follows.
- NULL semantics match the existing functions exactly: null literals are ignored by `IN`, a null literal makes `NOT IN` false for every row, and null column values never match.
- The structure is immutable after construction and shared through `std::shared_ptr<const LiteralSet>` when predicates are rebound to other schemas, so concurrent evaluation stays lock-free and the structure is not rebuilt per reader.

### Anything else?

No public API, storage format, or protocol change; everything is internal to `src/paimon/common/predicate/`. The statistics/min-max pruning path keeps using the existing `LeafFunction` implementation.

lucasfang:
### Purpose

Linked issue: close #262

`MultiLiteralsLeafFunction::Test` evaluates an `IN` / `NOT IN` predicate by materializing the whole column into `Literal` objects, one heap allocation per row, and then comparing every row against every literal. That is `O(rows × literals)` comparisons plus `O(rows)` allocations per batch, and with the large `IN` lists that get pushed down from SQL it becomes a measurable part of scan time even though membership testing only needs a set lookup.

This PR builds the literals into an arrow array once per batch and probes the whole column with `arrow::compute::is_in`, which turns the evaluation into `O(literals)` of setup plus an `O(rows)` hash probe with no per-row allocation. Every `LeafFunction` is a shared stateless singleton, so the value set cannot be cached on the function; building it per batch is what buys the vectorized probe.

Changes:

- `LiteralConverter::ConvertLiteralsToArray` is added as the reverse of the existing `ConvertLiteralsFromArray`. It writes `BOOLEAN`, `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `FLOAT`, `DOUBLE`, `DATE`, `STRING`, `BINARY`, `DECIMAL` and `TIMESTAMP`, keeps a null literal as an arrow null so the result has exactly one entry per literal, and returns `Status::Invalid` for any other field type. For `DECIMAL` the arrow type carries a precision and a scale that a `FieldType` alone does not settle, so the array is written with the pair the literals carry themselves: at least one literal has to be non null to settle it, and every non null one has to carry the same pair, because rescaling a value to another scale loses digits or overflows. The type is built with `arrow::Decimal128Type::Make`, since `arrow::decimal128` checks the precision fatally. For `TIMESTAMP` the array is written with the finest time unit that keeps every value (nanosecond if any nanos-of-millisecond is not a whole microsecond, microsecond if any is non zero, millisecond otherwise), which again needs one non null literal to settle.
- `MultiLiteralsLeafFunction::Test` moves out of the header into a new `multi_literals_leaf_function.cpp`. It takes the value set path for the field types listed above; `FLOAT` and `DOUBLE` join it except for a NaN literal, because arrow hashes the raw bits of a float while `FieldsComparator::CompareFloatingPoint` makes every NaN equal. `DECIMAL` probes only a column carrying the scale of the literals, because `is_in` compares a value set against a column of another scale by casting one side to the other, and that cast fails on a value that does not fit the other scale where `Literal::CompareTo` rescales one value at a time and merely finds no match. `TIMESTAMP` probes only a column of the very unit the values need and without a time zone, because casting to a coarser unit fails on a value the unit does not keep, casting to a finer one can overflow int64, and `is_in` refuses to compare a zoned timestamp against an unzoned one while a literal has no zone to speak of.
- `SetLookupOptions::EMIT_NULL` gives the exact null semantics the predicate already has: the nulls of the value set are ignored, which is what `IN` does with a null literal, and a null input row becomes a null output row, which both `IN` and `NOT IN` report as false.
- The row-by-row path is kept for everything the probe does not cover: the field types above when the column parameter does not match, literals of mixed types, which must keep reporting the `Literal::CompareTo` error rather than being built into one typed value set, and `NOT IN` holding a null literal, which is false for every row anyway. The fallback compares by value, so a decimal literal of another scale or a timestamp literal of another unit still matches the same number or instant.

Beyond the performance change this also fixes an evaluation failure: `ConvertLiteralsFromArray` rejects several dictionary layouts, for example int32 indices over an int64 dictionary or int8 indices over a utf8 dictionary, and an `IN` predicate over such a column used to fail the whole evaluation. `is_in` decodes the dictionary itself, so those columns now evaluate normally. It also promotes both sides to their common type when the column is read as a wider or narrower arrow type than the value set, and only reports a mismatch when the field type genuinely disagrees with the column.

### Tests

New unit tests in `src/paimon/common/predicate/multi_literals_leaf_function_test.cpp`, each asserting both `IN` and `NOT IN` results:

- Value set path per field type: `TestInt`, `TestBigInt`, `TestInt64Boundaries`, `TestTinyIntAndSmallInt`, `TestDate`, `TestBoolean`, `TestString`, `TestStringWithoutEmptyLiteral`, `TestBinary`, `TestFloatAndDouble`, `TestNegativeZeroDoesNotMatchPositiveZero`
- Decimal and timestamp value sets: `TestDecimal` including a column of the same scale but another precision and values beyond 64 bits, `TestTimestamp` for the millisecond and the nanosecond unit, `TestDecimalNullLiterals` and `TestTimestampNullLiterals` for null literals in the set and for nothing but null literals
- Fallback parity with the row-by-row path: `TestDecimalOffTheValueSetPath` for another scale, mixed precision and scale, and the scale pair where `is_in` would have failed the cast, `TestTimestampOffTheValueSetPath` for a coarser unit, a finer unit and a column with a time zone, `TestNanLiteralStaysOffTheValueSetPath`
- Dictionary columns: `TestDictionaryString`, `TestLargeStringDictionary`, `TestDictionaryWithNullValue`
- Null and empty handling: `TestNullLiteralIgnoredForIn`, `TestOnlyNullLiterals`, `TestEmptyLiterals`
- Error parity: `TestMixedLiteralTypesReportTheError`
- Type resolution: `TestProbeResolvesArrowTypeOnItsOwn`, which also pins the dictionary layouts that `ConvertLiteralsFromArray` rejects, and `TestProbeFailsOnUnrelatedArrowType`
- `TestSlicedArray` for sliced primitive and string arrays

New unit tests in `src/paimon/common/predicate/literal_converter_test.cpp`:

- `TestLiteralsToArray`: every writable field type, `BOOLEAN`, `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `FLOAT`, `DOUBLE`, `DATE`, `STRING`, `BINARY`, `DECIMAL` and `TIMESTAMP` with one case per millisecond, microsecond and nanosecond unit, each with a null literal in the middle, asserting the arrow array and that it reads back through `ConvertLiteralsFromArray` as the literals it was built from
- `TestLiteralsToArrayWithoutValue`: empty input, and input that is all nulls
- `TestLiteralsToArrayUnsupportedType`: `TIMESTAMP` and `DECIMAL` without a non null literal, decimal literals that do not share one precision and scale, a decimal precision arrow rejects instead of aborting, and `BLOB`, `ARRAY`, `MAP`, `STRUCT` and `UNKNOWN` all report the field type in the error

New end-to-end tests in `src/paimon/common/predicate/predicate_test.cpp`:

- `TestLargeStringIn`: a 1000-literal STRING `IN` / `NOT IN` over an array that includes the empty string and a null
- `TestInAfterRebind`: rebinding by field name and by field index keeps results identical
- `TestInt64BoundaryIn`: `IN` / `NOT IN` with `INT64_MIN` and `INT64_MAX` through `PredicateBuilder`

Verified with the full `paimon-common-test` suite: 1537 tests from 178 test suites, all passed, plus the wider regression `paimon-parquet-format-test` (202 tests), `paimon-orc-format-test` (119 tests) and `paimon-core-test` (1863 tests), whose existing decimal and timestamp `IN` / `NOT IN` cases now run through the new value set path.

### API and Format

No. No header under `include/` is touched, and neither the storage format nor the protocol changes. `LiteralConverter` gains one static method, which is purely additive.

### Documentation

No. This is a performance optimization, plus the dictionary-column fix described above, with no user-visible API or configuration change.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/src/paimon/common/predicate/literal_converter.cpp
+++ b/src/paimon/common/predicate/literal_converter.cpp
@@ -355,7 +355,7 @@ Result<std::shared_ptr<arrow::Array>> BuildDecimalArray(const std::vector<Litera
             PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.AppendNull());
             continue;
         }
-        Decimal value = literal.GetValue<Decimal>();
+        auto value = literal.GetValue<Decimal>();
         if (value.Precision() != precision || value.Scale() != scale) {
             return Status::Invalid(fmt::format(
                 "Not support converting literals of DECIMAL type to an arrow array, {} and {} do "
--- a/src/paimon/common/predicate/literal_converter.cpp
+++ b/src/paimon/common/predicate/literal_converter.cpp
@@ -29,6 +29,7 @@
 #include "arrow/array/builder_decimal.h"
 #include "arrow/array/builder_primitive.h"
 #include "arrow/array/builder_time.h"
+#include "arrow/memory_pool.h"
 #include "arrow/type.h"
 #include "arrow/type_traits.h"
 #include "arrow/util/decimal.h"
@@ -309,8 +310,8 @@ namespace {
 // Appends every literal, read by `extract`, to a `BuilderType`, keeping a null literal a null.
 template <typename BuilderType, typename Extract>
 Result<std::shared_ptr<arrow::Array>> BuildArray(const std::vector<Literal>& literals,
-                                                 Extract extract) {
-    BuilderType builder;
+                                                 Extract extract, arrow::MemoryPool* pool) {
+    BuilderType builder(pool);
     PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(static_cast<int64_t>(literals.size())));
     for (const auto& literal : literals) {
         if (literal.IsNull()) {
@@ -328,7 +329,8 @@ Result<std::shared_ptr<arrow::Array>> BuildArray(const std::vector<Literal>& lit
 // array holds one arrow type, so a non null literal is needed to settle it, and every non null
 // literal has to carry the same pair: rescaling a value to another scale loses digits or overflows,
 // which is not something a conversion decides on its own.
-Result<std::shared_ptr<arrow::Array>> BuildDecimalArray(const std::vector<Literal>& literals) {
+Result<std::shared_ptr<arrow::Array>> BuildDecimalArray(const std::vector<Literal>& literals,
+                                                        arrow::MemoryPool* pool) {
     std::optional<Decimal> typed_value;
     for (const auto& literal : literals) {
         if (!literal.IsNull()) {
@@ -346,7 +348,7 @@ Result<std::shared_ptr<arrow::Array>> BuildDecimalArray(const std::vector<Litera
     // `arrow::decimal128` checks the precision fatally, `Make` reports it instead.
     PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::DataType> type,
                                       arrow::Decimal128Type::Make(precision, scale));
-    arrow::Decimal128Builder builder(type);
+    arrow::Decimal128Builder builder(type, pool);
     PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(static_cast<int64_t>(literals.size())));
     for (const auto& literal : literals) {
         if (literal.IsNull()) {
@@ -371,7 +373,8 @@ Result<std::shared_ptr<arrow::Array>> BuildDecimalArray(const std::vector<Litera
 // Writes the literals to a timestamp array of the finest time unit that keeps every value, one
 // `is_in` can only compare against a column of that very unit. One array holds one arrow type, so
 // a non null literal is needed to settle the unit.
-Result<std::shared_ptr<arrow::Array>> BuildTimestampArray(const std::vector<Literal>& literals) {
+Result<std::shared_ptr<arrow::Array>> BuildTimestampArray(const std::vector<Literal>& literals,
+                                                          arrow::MemoryPool* pool) {
     bool has_value = false;
     for (const auto& literal : literals) {
         if (!literal.IsNull()) {
@@ -384,11 +387,8 @@ Result<std::shared_ptr<arrow::Array>> BuildTimestampArray(const std::vector<Lite
             "Not support converting literals of TIMESTAMP type to an arrow array without a non "
             "null literal to take the time unit from");
     }
-    // The pool is the default one every other builder here takes implicitly, a builder of a
-    // parameterized type just has no default to fall back on.
     arrow::TimestampBuilder builder(
-        arrow::timestamp(LiteralConverter::MinRequiredTimeUnit(literals)),
-        arrow::default_memory_pool());
+        arrow::timestamp(LiteralConverter::MinRequiredTimeUnit(literals)), pool);
     const DateTimeUtils::TimeType time_type = DateTimeUtils::GetTimeTypeFromArrowType(
         checked_pointer_cast<arrow::TimestampType>(builder.type()));
     PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(static_cast<int64_t>(literals.size())));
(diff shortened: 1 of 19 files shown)


src/paimon/common/predicate/literal_converter.cpp
```diff
@@ -284,4 +289,163 @@ std::vector<Literal> LiteralConverter::GetLiteralFromTimestampArray(const arrow:
...
     return literals;
 }
+
+namespace {
+// Appends every literal, read by `extract`, to a `BuilderType`, keeping a null literal a null.
+template <typename BuilderType, typename Extract>
+Result<std::shared_ptr<arrow::Array>> BuildArray(const std::vector<Literal>& literals,
+                                                 Extract extract) {
+    BuilderType builder;
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(builder.Reserve(static_cast<int64_t>(literals.size())));
```
lxy-9602, line 315: Does this builder need to take a memory pool?

src/paimon/common/predicate/literal_converter.cpp
```diff
@@ -284,4 +289,163 @@ std::vector<Literal> LiteralConverter::GetLiteralFromTimestampArray(const arrow:
...
+            break;
+        }
+    }
+    if (!has_value) {
+        return Status::Invalid(
+            "Not support converting literals of TIMESTAMP type to an arrow array without a non "
+            "null literal to take the time unit from");
+    }
+    arrow::TimestampBuilder builder(arrow::timestamp(MinRequiredTimeUnit(literals)),
+                                    arrow::default_memory_pool());
```
lxy-9602: Why are we using `arrow::default_memory_pool()` here? Other builders don’t seem to pass a memory pool. Also, should this use the default pool, or should it use the caller-provided pool instead?

src/paimon/common/predicate/multi_literals_leaf_function.cpp
```diff
@@ -0,0 +1,254 @@
...
+            continue;
+        }
+        const int32_t nano = literal.GetValue<Timestamp>().GetNanoOfMillisecond();
+        if (nano % 1000 != 0) {
+            return arrow::TimeUnit::NANO;
+        }
+        needs_micro = needs_micro || nano != 0;
+    }
+    return needs_micro ? arrow::TimeUnit::MICRO : arrow::TimeUnit::MILLI;
+}
```
lxy-9602: Could we extract `MinRequiredTimeUnit` into a shared helper? It seems to appear in multiple places.

src/paimon/common/predicate/multi_literals_leaf_function_test.cpp
```diff
@@ -0,0 +1,539 @@
...
+    ASSERT_EQ(EvalIn({TimestampLiteral(1000)}, nano_array), std::vector<char>({1, 0}));
+    ASSERT_EQ(EvalNotIn({TimestampLiteral(1000)}, nano_array), std::vector<char>({0, 0}));
+
+    // A column with a time zone as well: `is_in` refuses to compare a zoned timestamp against an
+    // unzoned value set, so the row by row path compares by value instead.
+    arrow::TimestampBuilder builder(arrow::timestamp(arrow::TimeUnit::MILLI, "UTC"),
+                                    arrow::default_memory_pool());
+    ASSERT_TRUE(builder.Append(1000).ok());
+    ASSERT_TRUE(builder.Append(2000).ok());
+    ASSERT_TRUE(builder.AppendNull().ok());
```
lxy-9602: Why is the code above still using JSON, while here it switches to `builder.append`?

src/paimon/common/predicate/multi_literals_leaf_function_test.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+TEST_F(MultiLiteralsLeafFunctionTest, TestMixedLiteralTypesReportTheError) {
+    // Literals of mixed types make `Literal::CompareTo` fail, so they must keep reporting the error
+    // through the row by row path instead of being built into one typed value set. The column value
+    // matches none of them, otherwise the comparison would stop before reaching the odd literal.
+    auto array = arrow::ipc::internal::json::ArrayFromJSON(arrow::int32(), R"([0])").ValueOrDie();
+    ASSERT_NOK(In::Instance().Test(*array, {Literal(1), Literal(int64_t{2})}).status());
+    ASSERT_NOK(In::Instance().Test(*array, {Literal(1), StringLiteral("a")}).status());
+    ASSERT_NOK(NotIn::Instance().Test(*array, {Literal(1), Literal(int64_t{2})}).status());
+}
```
lxy-9602, line 471: ASSERT_NOK(Func());

src/paimon/common/predicate/multi_literals_leaf_function_test.cpp
```diff
@@ -0,0 +1,539 @@
...
+
+TEST_F(MultiLiteralsLeafFunctionTest, TestProbeResolvesArrowTypeOnItsOwn) {
+    // A column read as a narrower arrow type than the field type is promoted to the common type, so
+    // the probe stays correct instead of having to fall back.
```
lxy-9602: Is this kind of test necessary? I thought predicate validation already checks that the literal type matches the schema.

src/paimon/common/predicate/multi_literals_leaf_function_test.cpp
```diff
@@ -0,0 +1,529 @@
...
+        return Literal(Timestamp::FromEpochMillis(millis, nanos));
+    }
+
+    // Evaluates the whole batch and returns the per row result, asserting the call succeeded.
+    static std::vector<char> Eval(const LeafFunction& function,
+                                  const std::vector<Literal>& literals,
+                                  const std::shared_ptr<arrow::Array>& array) {
+        Result<std::vector<char>> result =
+            function.Test(*array, literals, arrow::default_memory_pool());
+        EXPECT_OK(result.status());
```
lxy-9602: EXPECT_OK_AND_ASSIGN

lxy-9602 (approved):
+1

PR #267, opened 2026-08-31
perf(scan): reuse loaded global index planning context

wangyong9999:
### Purpose

Linked issue: N/A

`DataEvolutionBatchScan` used `GlobalIndexScan::Create()` while planning an index predicate. That public entry point reloaded the latest table schema and snapshot and constructed its own planning context, even though `TableScan` had already resolved the same state.

**The DataEvolution scan now builds its internal `GlobalIndexScanImpl` from the schema, snapshot manager, `CoreOptions`, filesystem, executor, memory pool, and caller-provided cache already owned by the current `TableScan`.** The index and data planning stages therefore use the same scan context, and repeated plans can hit the cache supplied through `ScanContextBuilder::WithCache()`.

This PR does not retain `GlobalIndexReader` objects or BTree pages across queries and adds no process-global registry. Cross-query reader ownership, eviction, memory accounting, and `IsThreadSafe()` handling remain query-engine responsibilities through the existing public `GlobalIndexScan`, `CreateReader`, and `SetGlobalIndexResult` APIs.

### Tests

- `paimon-global-index-test`: 132/132 passed.
- `FileFormat/GlobalIndexTest.TestDataEvolutionBatchScan/*`: 6/6 passed.
- The DataEvolution integration case verifies that a caller-provided manifest cache has no new supplier calls on the second scan.
- `git diff --check` passes.

The local environment does not have `pre-commit` installed; GitHub Actions cover the repository lint hooks.

### API and Format

No public API, storage-format, or protocol change. The modified constructor is internal to the core implementation.

### Documentation

No user-facing option or behavior change.

### Generative AI tooling

Generated-by: Codex (GPT-5)


wangyong9999:
The failure came from process-wide BTree pages warmed by an earlier fixture. The cache key now includes the backend namespace, so cross-fixture pages cannot be reused. I restored the exact file-open assertion; all 12 primary-key sorted-index integration cases pass for Parquet and ORC. The latest fork workflows are waiting for maintainer approval.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -42,9 +45,40 @@
...
+    // Keep the allocator alive until after cache_manager releases all cached pages.
+    std::shared_ptr<MemoryPool> cache_pool;
+    std::shared_ptr<CacheManager> cache_manager;
+};
+
+std::shared_ptr<CacheManager> GetSharedCacheManager(int64_t cache_size,
+                                                    double high_priority_pool_ratio) {
+    using CacheConfig = std::pair<int64_t, double>;
+    static std::mutex mutex;
+    static std::map<CacheConfig, SharedCacheManager> cache_managers;
```
wangyong9999: `cache_managers` owns a full LRU for every distinct table-level `(cache_size, ratio)` forever. Each cache is bounded, but the process total is not; reading tables with different settings keeps all previous budgets resident. Please use one process-wide budget, or bound and evict this registry.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
```diff
@@ -57,7 +91,8 @@ Result<std::unique_ptr<BTreeGlobalIndexer>> BTreeGlobalIndexer::Create(
         double high_priority_pool_ratio,
         OptionsUtils::GetValueFromMap<double>(options, BtreeDefs::kBtreeIndexHighPriorityPoolRatio,
                                               BtreeDefs::kDefaultBtreeIndexHighPriorityPoolRatio));
-    auto cache_manager = std::make_shared<CacheManager>(cache_size, high_priority_pool_ratio);
+    std::shared_ptr<CacheManager> cache_manager =
+        GetSharedCacheManager(cache_size, high_priority_pool_ratio);
```
wangyong9999: This manager is shared across all indexers with the same size settings, but page keys contain only `file_path/offset/length`. Two custom `FileSystem` or reader backends can expose the same logical path with different bytes, so the second scan can reuse the first backend's pages without opening its own file and return wrong matches. Include a stable backend namespace in the key, or scope the manager by backend.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -150,12 +150,12 @@ Result<std::shared_ptr<FileStoreScan::RawPlan>> FileStoreScan::CreatePlan() cons
         snapshot.has_value() && scan_mode_ == ScanMode::ALL &&
         core_options_.GetScanManifestEntryCacheMaxSnapshots() > 0 &&
         core_options_.GetCache() != nullptr && !table_path_.empty() &&
-        !row_range_index_.has_value() && bucket_filter_.has_value();
+        !row_range_index_.has_value();
```
wangyong9999: The whole-table key is only `(table_path, branch)`, and a hit is accepted by numeric snapshot ID. After a table is dropped and recreated at the same path, snapshot IDs restart while the caller-owned cache can still hold the old entry, so an ordinary scan can return stale splits. Include a table or snapshot generation identity (for example the manifest-list name) in the cached entry/key, or invalidate it on recreation.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -347,16 +347,21 @@ Status FileStoreScan::ReadManifestEntriesWithCache(
...
+            if (MayContainBucket(meta, bucket.value())) {
+                bucket_manifest_metas.push_back(meta);
+            }
         }
+        PAIMON_RETURN_NOT_OK(
+            ReadAndMergeBucketFileEntries(bucket_manifest_metas, bucket.value(), manifest_entries));
+    } else {
+        std::vector<ManifestEntry> unmerged_entries;
+        PAIMON_RETURN_NOT_OK(
+            ReadFileEntries(all_manifest_metas, &unmerged_entries, /*apply_scan_filter=*/false));
```
wangyong9999: On this no-bucket miss we materialize every manifest entry, then serialize up to `max_snapshots` full table states. If that value exceeds the cache capacity, `Put` fails silently and every later scan of the same snapshot repeats the full read, merge, copy, and serialization. Only use this path when the value can be retained; otherwise fall back to `filtered_manifest_file_metas`, or store snapshots separately so older states can be evicted.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
wangyong9999: Fixed in b35e9c70. Shared page keys now include an internal backend namespace. Filesystem-backed readers share by FileSystem instance; unknown readers are isolated by reader identity, and cached values retain the namespace owner until eviction.

src/paimon/common/global_index/btree/btree_global_indexer.cpp
wangyong9999: Fixed in 9cd2147b. The process now retains at most four cache configurations and evicts the least recently used registry entry. An evicted manager survives only while an active reader still references it.

src/paimon/core/operation/file_store_scan.cpp
wangyong9999: Fixed in 9cd2147b. Bucket cache entries now store the snapshot generation derived from the base and delta manifest-list names. A hit requires both snapshot ID and generation to match.

src/paimon/core/operation/file_store_scan.cpp
wangyong9999: Fixed in 9cd2147b by removing the no-bucket whole-table manifest cache path. Only bounded bucket-specific entries remain, so an oversized whole-table value cannot trigger repeated materialization and failed insertion.

lxy-9602:
Thanks for working on the repeated Global Index lookup performance.

Some downstream query engines, such as StarRocks, have a similar requirement. Their approach is to create a `GlobalIndexReader` through the public API in `global_index_scan.h` and retain/cache the reader for multiple queries. This already preserves the BTree reader's internal block cache without introducing process-wide cache sharing inside paimon-cpp.

The process-wide design in this PR introduces additional concepts and hidden global state, including the shared `CacheManager` registry, cache configuration LRU, `cache_namespace`, backend lifetime retention, and cross-reader allocator management. It also makes memory accounting and cache isolation harder for downstream engines that already have their own resource management.

Could we instead encourage the query engine to split the query into two explicit stages?

1. **Global Index stage:** create or reuse a `GlobalIndexReader` through `GlobalIndexScan`, evaluate the index predicate, and produce a `GlobalIndexResult`. This stage can be distributed by shard.
2. **Data Scan stage:** pass the returned `GlobalIndexResult` through `ScanContextBuilder::SetGlobalIndexResult()`, then use `TableScan` to generate the data scan plan and execute it.

The engine can cache readers using a key containing the table identity, snapshot/index generation, field, index type, shard/partition, row range, and relevant options. It should also ensure that the index lookup and subsequent data scan use the same snapshot, and respect `GlobalIndexReader::IsThreadSafe()` when sharing readers.

This keeps cache ownership, eviction, memory accounting, and isolation under the query engine's control while reusing the existing public API. Could we evaluate this approach before adding process-wide cache sharing to paimon-cpp?

wangyong9999:
Agreed. I removed the process-wide reader and BTree cache registry, namespace, and allocator-lifetime changes. The query engine now owns cross-query `GlobalIndexReader` reuse through the existing public API, pins the data scan to the same snapshot, and honors `IsThreadSafe()` around shared-reader evaluation.

This PR is now limited to reusing the schema, snapshot context, executor, memory pool, and caller-provided manifest cache already available inside one `TableScan`. The resulting diff is 4 files; `paimon-global-index-test` passes 132/132.

lxy-9602 (approved):
+1

PR #269, opened 2026-08-31
feat(realtime): support external offsets and shredding-aware reads

lxy-9602:
### Purpose

Linked issue: #158 

This change improves realtime writes and reads.

For realtime-enabled writes, external engines must provide a leading non-nullable `INT64 _REALTIME_OFFSET`:

```text
[_REALTIME_OFFSET, table write fields...]
```

Offsets must be strictly increasing for each partition-bucket, but gaps are allowed. Paimon persists the exclusive committed offset in the snapshot offset file for WAL recovery. `_REALTIME_OFFSET` is not written into data files.

Because offsets may be sparse, `OffsetRange` remains left-closed and right-open but no longer represents the row count.

This PR also moves offset filtering and physical-to-logical conversion into a shared Paimon read pipeline for append-only and primary-key tables. It supports consistent disk and realtime-memory reads for:
- selected-key and shared-shredding MAP access;
- nested projections;
- shredded VARIANT access and binary fallback.

### Tests
- external and sparse realtime offsets;
- invalid, duplicate, backward, and overflowing offsets;
- commit progress, snapshot offsets, and recovery;
- exact row counting with sparse offsets;
- append-only and primary-key realtime reads;
- MAP and VARIANT projections across disk and memory.

### API and Format
Realtime write input and RealtimeStore APIs are updated to support external offsets and shared read processing.
Data-file and snapshot offset-file formats are unchanged.

### Generative AI tooling


include/paimon/realtime/offset_range.h
```diff
@@ -29,7 +29,8 @@ namespace paimon {
 struct PAIMON_EXPORT OffsetRange {
     OffsetRange(int64_t begin, int64_t end) : begin(begin), end(end) {}
 
-    /// Returns the number of offsets covered by this range.
+    /// Returns the width of this range. Real-time offsets may have gaps, so this is not a row
+    /// count.
     int64_t Count() const {
```
zjw1111: any use of this func now?

src/paimon/core/operation/commit/realtime_commit_properties.cpp
```diff
@@ -285,9 +280,9 @@ Result<std::map<std::string, std::string>> RealtimeCommitProperties::Build(
         }
         auto offset_iter = merged_offsets.find(partition_bucket);
         int64_t previous_end_offset = offset_iter == merged_offsets.end() ? 0 : offset_iter->second;
-        if (offset_range.begin != previous_end_offset) {
+        if (offset_range.begin < previous_end_offset) {
```
HaHaJeff, line 283: Could this skip an earlier prepared segment?\n\nFor example, if P0 contains offset 10 and P1 contains offset 30, committing only P1 advances the watermark to 31. P0 may then be reclaimed and later treated as already committed, although its files were never published.\n\nI think row offsets can remain sparse, but commit progress should still be contiguous.

src/paimon/core/realtime/realtime_primary_key_writer.cpp
```diff
@@ -245,15 +235,14 @@ Result<CommitIncrement> RealtimePrimaryKeyWriter::PrepareCommit(bool wait_compac
     return increment;
 }
 
-Status RealtimePrimaryKeyWriter::FlushSegment(const std::shared_ptr<RealtimeSegmentHandle>& segment,
-                                              const OffsetRange& sealed_offsets) {
+Status RealtimePrimaryKeyWriter::FlushSegment(
+    const std::shared_ptr<RealtimeSegmentHandle>& segment) {
     PAIMON_ASSIGN_OR_RAISE(std::vector<std::unique_ptr<BatchReader>> readers,
                            realtime_store_->CreateCommitReaders(segment));
```
HaHaJeff, line 242: Could we validate the raw PK commit readers before MOR?

This path no longer checks segment->GetRowCount(). If a custom store misses or duplicates a mutation, deduplication may hide it and incomplete data could be committed.

We should at least check the raw row count and validate that _REALTIME_OFFSET is non-null, unique, and within the sealed range. The same offset validation may also be useful for append.

src/paimon/core/io/key_value_data_file_record_reader.cpp
```diff
@@ -107,26 +110,34 @@ Result<std::unique_ptr<KeyValueRecordReader::Iterator>> KeyValueDataFileRecordRe
...
+    row_kind_array_ = checked_pointer_cast<arrow::NumericArray<arrow::Int8Type>>(row_kind);
+    if (row_kind_array_->null_count() != 0) {
+        return Status::Invalid("VALUE_KIND column contains null");
+    }
     arrow::ArrayVector key_fields;
     key_fields.reserve(key_schema_->num_fields());
     for (const auto& key_field : key_schema_->fields()) {
-        // skip special fields
-        key_fields.emplace_back(data_batch->GetFieldByName(key_field->name()));
+        std::shared_ptr<arrow::Array> field_array = data_batch->GetFieldByName(key_field->name());
```
HaHaJeff, line 135: Do we also need to validate the key and value field types here?

`GetFieldByName` only verifies that the field exists. Since `ColumnarRowRef` later selects typed getters from the expected schemas, a custom store returning the same field name with an incompatible Arrow type may produce incorrect PK ordering/MOR results, or even an out-of-bounds read for a width mismatch.

Would comparing the actual types against `key_schema_` and `value_schema_` on the first batch be sufficient? The metadata does not need to match exactly.

src/paimon/core/realtime/realtime_primary_key_reader.cpp
```diff
@@ -18,493 +18,74 @@
...
-                "realtime primary-key transport field id {} type {} does not match row type {}",
-                field_id, transport_field->type()->ToString(), row_field->type()->ToString()));
+Result<std::vector<std::unique_ptr<KeyValueRecordReader>>> CreateKeyValueReaders(
+    std::vector<std::unique_ptr<BatchReader>>&& readers,
+    const std::shared_ptr<arrow::Schema>& key_schema,
+    const std::shared_ptr<arrow::Schema>& value_schema,
+    const std::shared_ptr<MemoryPool>& memory_pool) {
+    std::vector<std::unique_ptr<KeyValueRecordReader>> result;
+    result.reserve(readers.size());
+    for (std::unique_ptr<BatchReader>& reader : readers) {
```
HaHaJeff, line 57: The raw readers need failure-path cleanup here.

If a later reader is null, readers already returned by the plugin are destroyed without an explicit `Close()`. Some store implementations may release spill handles or other external resources only from `Close()`.

A cleanup guard together with a tracking-reader test would cover this. The append `pipeline->Wrap()` failure path appears to need the same treatment.

src/paimon/core/realtime/realtime_primary_key_writer.cpp
lxy-9602: Agreed that we can compare the total number of rows emitted by the raw commit readers with `segment->GetRowCount()`, consistent with the append path.

While I don’t think core should revalidate `_REALTIME_OFFSET` uniqueness or range here. `RealtimeStore::CreateCommitReaders` already requires implementations to expose every sealed row exactly once, similar to how core trusts other plugin (e.g., file-format readers) to return correct field values. Offset uniqueness would also require additional state because PK readers are sorted by key and sequence rather than offset.

Even comprehensive offset validation cannot prove that the plugin returned the correct row contents or associated each offset with the correct mutation. Extending validation in this direction would effectively make core revalidate the plugin’s entire output, which is inconsistent with the trust boundary used for other plugin. A custom store violating this contract should be treated as a plugin bug.

include/paimon/realtime/offset_range.h
BinChenn: Range() may be more accurate

src/paimon/core/realtime/realtime_primary_key_writer.cpp
HaHaJeff: This boundary makes sense. The raw row-count check before MOR should be sufficient here.

lxy-9602: I agree that validating the raw row count would provide useful defense in depth. However, there is no simple reliable check at this layer with the current streaming reader contract.

Unlike append, PK commit readers are multiple one-shot streams consumed through MOR. The row count after MOR is not comparable because deduplication may reduce it. Checking before MOR would require wrapping every raw reader with shared counting state, detecting completion across all reader EOFs, and propagating a mismatch before the rolling file writer completes. Pre-reading would instead require buffering or reopening the streams.

Since `CreateCommitReaders` already requires plugins to expose every sealed row exactly once, I will add a TODO/follow-up for cardinality validation.

src/paimon/core/operation/commit/realtime_commit_properties.cpp
lxy-9602: Good point. This can happen if the caller passes only P1, but we consider that invalid input to `CommitWithProgress`.

The range here describes the envelope of actual row offsets, which may be sparse. Therefore, the committer cannot distinguish a legitimate offset gap from an omitted prepared segment, and requiring numeric continuity would reject valid sparse offsets.

Streaming engines generally enforce this at the checkpoint coordinator layer. For example, Flink’s checkpoint-subsuming contract requires a later completed checkpoint to include all artifacts from earlier uncommitted checkpoints. The sink coordinator retains prepared-but-uncommitted committables and submits the complete pending prefix when committing a later checkpoint. Low-level offset APIs such as Kafka also trust the caller to commit the next valid consumption position rather than validating that every preceding record was processed.

We will keep sparse ranges supported and document that the upstream coordinator must provide all earlier prepared-but-uncommitted progress for each partition-bucket. Omitting an earlier prepared segment violates the API contract and may cause its unpublished data to be treated as committed and reclaimed.

src/paimon/core/io/key_value_data_file_record_reader.cpp
lxy-9602: The physical Arrow type returned by a file reader may differ from the logical schema type. For example, a logical string field may be returned as a dictionary-encoded array to avoid unnecessary decoding and copying (e.g., when set `orc.read.enable-lazy-decoding` or `parquet.read.enable-dictionary-passthrough`). The read path already handles such representations explicitly.

Therefore, comparing the actual type directly with `key_schema_` or `value_schema_` would reject valid reader output. For the same reason as the discussion above, we will only verify that the required fields exist and will not validate exact type equality here for now.

src/paimon/core/realtime/realtime_primary_key_reader.cpp
lxy-9602: Done

include/paimon/realtime/offset_range.h
lxy-9602: Given that `OffsetRange` may be non-contiguous now, I removed this `Count` API to avoid ambiguity. Thanks!

include/paimon/realtime/realtime_store.h
```diff
@@ -134,7 +135,7 @@ class PAIMON_EXPORT RealtimeStore {
 
     /// Adds a batch to the current building segment.
     ///
-    /// The row count matches the framework-assigned `offset_range`.
+    /// The offset envelope may contain gaps and does not imply the batch row count.
```
zjw1111: is `and does not imply the batch row count` needed to say?

src/paimon/core/realtime/realtime_primary_key_reader.cpp
```diff
@@ -58,453 +43,66 @@ void CloseReaders(const std::vector<std::unique_ptr<Reader>>& readers) {
...
-    std::shared_ptr<ColumnarBatchContext> key_ctx_;
-    std::shared_ptr<ColumnarBatchContext> value_ctx_;
-    std::shared_ptr<arrow::NumericArray<arrow::Int8Type>> row_kind_array_;
-    std::shared_ptr<arrow::NumericArray<arrow::Int64Type>> sequence_number_array_;
-    std::vector<int64_t> selected_rows_;
-};
-
 }  // namespace
 
 std::shared_ptr<arrow::Schema> RealtimePrimaryKeyLayout::CreateSchema(
```
zjw1111, line 71: direct use SpecialFields::CompleteSequenceAndValueKindField?

src/paimon/core/realtime/realtime_offset_utils.h
```diff
@@ -0,0 +1,102 @@
...
+        std::shared_ptr<arrow::StructArray> data;
+        OffsetRange offset_range;
+    };
+
+    static std::shared_ptr<arrow::Schema> CreateInputSchema(
+        const std::shared_ptr<arrow::Schema>& write_schema) {
+        arrow::FieldVector fields = {
+            DataField::ConvertDataFieldToArrowField(SpecialFields::RealtimeOffset())};
+        fields.insert(fields.end(), write_schema->fields().begin(), write_schema->fields().end());
+        return arrow::schema(std::move(fields), write_schema->metadata());
```
zjw1111, line 54: need set `write_schema->metadata()` or not? now different with SpecialFields::CompleteSequenceAndValueKindField, one set, the other not set

test/inte/realtime_write_inte_test.cpp
```diff
@@ -840,6 +897,10 @@ class RealtimeWriteInteTest : public ::testing::Test {
         return Status::OK();
     }
 
+    void RunUnionReadWithSelectedMapKeys(bool primary_key);
+
+    void RunUnionReadWithVariantAccess(bool primary_key);
+
     void RunConcurrencyTest(bool primary_key);
```
zjw1111: change these Run* func to lambda?

include/paimon/realtime/realtime_store.h
zjw1111: the comments like "offsets may have gaps, so its count is not the batch row count." appear many times

src/paimon/core/realtime/realtime_primary_key_reader.cpp
lxy-9602: The ordering is currently inconsistent: `CompleteSequenceAndValueKindField` adds `SequenceNumber` before `ValueKind`, while `RealtimePrimaryKeyLayout::CreateSchema` does it the other way around. I’ll clean this up and unify the schema-related logic in a follow-up PR.

test/inte/realtime_write_inte_test.cpp
```diff
@@ -897,10 +897,6 @@ class RealtimeWriteInteTest : public ::testing::Test {
         return Status::OK();
     }
 
-    void RunUnionReadWithSelectedMapKeys(bool primary_key);
-
-    void RunUnionReadWithVariantAccess(bool primary_key);
-
     void RunConcurrencyTest(bool primary_key);
```
zjw1111: also this func

zjw1111 (approved):
+1

src/paimon/core/realtime/realtime_primary_key_writer.cpp
```diff
@@ -245,15 +235,15 @@ Result<CommitIncrement> RealtimePrimaryKeyWriter::PrepareCommit(bool wait_compac
     return increment;
 }
 
-Status RealtimePrimaryKeyWriter::FlushSegment(const std::shared_ptr<RealtimeSegmentHandle>& segment,
-                                              const OffsetRange& sealed_offsets) {
+Status RealtimePrimaryKeyWriter::FlushSegment(
+    const std::shared_ptr<RealtimeSegmentHandle>& segment) {
+    // TODO(xinyu.lxy): Validate row count from commit readers.
```
SteNicholas, line 240: `GetRowCount()` was added because sparse offset envelopes no longer encode cardinality, but this path never compares the raw rows returned by `CreateCommitReaders()` with the segment count. A store that drops or duplicates a mutation can therefore pass through MOR, publish files, and advance the committed offset past data that was never persisted. Please add shared counting around the raw reader streams and fail at EOF unless the total equals `segment->GetRowCount()`, as the append path already does.

src/paimon/core/realtime/realtime_offset_batch_reader.cpp
```diff
@@ -0,0 +1,106 @@
...
+        if (!offset_field || offset_field->type_id() != arrow::Type::INT64) {
+            return Status::Invalid("realtime query batch must contain int64 _REALTIME_OFFSET");
+        }
+        std::shared_ptr<arrow::Int64Array> offsets =
+            checked_pointer_cast<arrow::Int64Array>(offset_field);
+        if (offsets->null_count() != 0) {
+            return Status::Invalid("realtime query offset column contains null");
+        }
+
+        RoaringBitmap32 output_bitmap;
```
SteNicholas, line 74: This adapter keeps only rows selected by the store bitmap. PK queries pass `predicate=nullptr`, so every raw mutation must participate in MOR; a partial bitmap can silently omit the newest update or delete and let an older disk row resurface. The previous PK adapter explicitly required the bitmap cardinality to equal the batch length. Please preserve that check for PK query readers, for example through a strict/all-rows pipeline mode.

src/paimon/core/realtime/realtime_primary_key_writer.cpp
lxy-9602: This is similar to one comments above. We should clearly define the plugin contract and trust the results returned by the plugin. For example, when reading through ORC/Parquet format readers, we do not re-validate the returned row count against the value recorded in metadata.

Also, for MOR, the row count during the process does not necessarily match the final file result, since further merging happens later. Verifying the row count in that case would require more extensive changes, and I’m not sure such changes are worthwhile if they are only for row-count validation. So I left a TODO here as a follow-up item for now.

src/paimon/core/realtime/realtime_offset_batch_reader.cpp
lxy-9602: Done.

src/paimon/core/realtime/realtime_offset_batch_reader.cpp
```diff
@@ -0,0 +1,110 @@
...
+                return Status::Invalid(fmt::format(
+                    "selected row id {} is out of bounds for realtime query batch length {}", row,
+                    offsets->length()));
+            }
+            const int64_t offset = offsets->Value(row);
+            if (offset >= visible_offsets_.begin && offset < visible_offsets_.end) {
+                output_bitmap.Add(row);
+            }
+        }
+        if (static_cast<int64_t>(input_bitmap.Cardinality()) != offsets->length()) {
```
SteNicholas, line 87: This shared adapter is also used by append reads, whose `RealtimeQueryContext` may contain a predicate and explicitly allows the plugin to prune candidate rows. `BatchReader::NextBatchWithBitmap()` represents those index-filtered rows with a partial bitmap, so this unconditional check rejects valid plugin output before the framework exact predicate filter runs. Please add a strict/all-rows mode enabled only by the PK query path, while preserving partial bitmaps for append queries.

lxy-9602: Thanks for the comments. This is not an issue in the current implementation, because predicate pushdown is currently applied at the batch level: an entire batch is either kept or skipped, and we do not go through a bitmap path. So this validation is fine for now.

Once we support DV and finer-grained pushdown (like pk predicate pushdown), the `Cardinality` check will likely be removed for both PK and append mode. That is planned for the next PR or the one after that.

SteNicholas (approved):
LGTM.

PR #271, opened 2026-09-01
feat(file-index): support additional types and validate decimal scale

zjw1111:
### Purpose

Linked issue: #173

Extend file-index read/write support while keeping serialized data compatible with Java Paimon, and prevent silent DECIMAL index misfiltering at table entry points:

- Support FLOAT and DOUBLE in bitmap indexes.
- Support DECIMAL in BSI indexes.
- Support STRING, DECIMAL, and TIMESTAMP in range-bitmap indexes.
- Add a range-bitmap type adapter for logical-to-storage type and literal conversion.
- Add variable-length dictionary chunks for string keys and avoid eagerly reading chunk bodies.
- Cover canonical and payload NaNs, signed zero, infinities, and Java-generated bitmap bytes.
- Reject DECIMAL predicate literals whose scale differs from the schema in TableScan and TableRead creation.
- Allow different DECIMAL precision metadata when the value fits the field precision, and reject values that do not fit.
- Document that direct FileIndexReader and GlobalIndexReader Visit* calls must use the indexed field scale.
- Add regression coverage showing that mathematically equivalent DECIMAL literals with a different scale produce incorrect results in BSI, range-bitmap, and BTree readers.

### Tests

- pre-commit run --files <all changed files>
- git diff --check upstream/main...HEAD
- Compiled the original file-index translation units with the repository Debug flags, including -Werror and ASan.
- BitmapIndexTest.* (14 tests)
- BitSliceIndexBitmapIndexReaderTest.*:ChunkedDictionaryTest.*:RangeBitmapFileIndexTest.*:RangeBitmapTypeAdapterTest.* (55 tests)
- SizeLimitAndCardinality/* (70 tests)
- PredicateValidatorTest.* (3 tests)
- BitSliceIndexBitmapIndexReaderTest.TestDecimalType and RangeBitmapFileIndexTest.TestWriteAndReadStringDecimalAndTimestamp (2 focused DECIMAL scale tests)
- Compression/BTreeGlobalIndexIntegrationTest.WriteAndReadDecimalCompactData/* (3 compression variants)

### API and Format

No public API signatures are changed. TableScan and TableRead now reject DECIMAL predicate literals with a scale different from the schema, or a value that does not fit the field precision. Direct index-reader APIs retain their existing behavior and now document the same-scale requirement.

The bitmap, BSI, and range-bitmap formats are extended to the additional logical types without changing their format versions. Floating-point bitmap keys are verified against Java-generated V1/V2 bytes.

### Documentation

Add public FileIndexReader and GlobalIndexReader notes requiring DECIMAL literals passed to Visit* to use the indexed field scale.

### Generative AI tooling

Generated-by: OpenAI Codex (GPT-5).


src/paimon/common/file_index/bitmap/bitmap_file_index_test.cpp
```diff
@@ -620,6 +645,229 @@ TEST_F(BitmapIndexTest, TestTimestampType) {
...
+
+    for (int32_t version : {1, 2}) {
+        check_float(version);
+        check_double(version);
+    }
+}
+
+TEST_F(BitmapIndexTest, TestFloatingPointJavaCompatibility) {
+    // Generated by BitmapFloatingPointCompatibilityTest with Apache Paimon Java at
+    // 0043a70fd88ac75dcb83a8f2da5e72ce91e22b1f. The Java writer receives canonical,
```
lxy-9602: Is `BitmapFloatingPointCompatibilityTest` a test for the Java repo? I’m not sure I understand the intent behind this test name. If it is not necessary, maybe we should just remove it together with this commit.

src/paimon/core/operation/internal_read_context.cpp
```diff
@@ -276,10 +276,10 @@ Result<std::unique_ptr<InternalReadContext>> InternalReadContext::Create(
     }
     // validate predicate
     if (context->GetPredicate()) {
-        PAIMON_RETURN_NOT_OK(PredicateValidator::ValidatePredicateWithSchema(
-            *read_schema, context->GetPredicate(), /*validate_field_idx=*/true));
         PAIMON_RETURN_NOT_OK(
             PredicateValidator::ValidatePredicateWithLiterals(context->GetPredicate()));
+        PAIMON_RETURN_NOT_OK(PredicateValidator::ValidatePredicateWithSchema(
+            *read_schema, context->GetPredicate(), /*validate_field_idx=*/true));
     }
```
lxy-9602, line 281: Is this assumption a bit too tightly coupled? It depends on the call order always being `ValidatePredicateWithLiterals` first and then `ValidatePredicateWithSchema`. If these two are always used together, maybe it would make sense to wrap them into a single abstraction.

lxy-9602 (dismissed):
+1

lxy-9602 (approved):
+1

PR #272, opened 2026-09-01
feat(core): add a block cache for the reads the prefetched ranges do not cover

### LINKED ISSUE #273: [Feature] Add a block cache for the reads that the prefetched ranges do not cover
## Search before asking

## Motivation

`ReadAheadCache` caches exactly the byte ranges registered through `Init()` and nothing else. Any read outside those ranges is counted as a miss and handed back to the caller, which then reads it from the underlying stream — and that read is not cached either, so the next reader asking for the same bytes pays for it again.

That leaves a whole class of reads uncached, even though a single `ReadAheadCache` is shared by all the readers `PrefetchFileBatchReaderImpl` creates for one data file and outlives a round of registered ranges:

1. **Reads issued before the ranges are known.** A reader has to read a file's metadata in order to compute the ranges it wants to prefetch, so those reads are by construction never covered by a registered range. They are also identical for every reader of the file, so with `prefetch_max_parallel_num` readers the same bytes are fetched N times, each time a separate full-latency request on the critical path before any data can be read.
2. **Reads outside the current prefetch plan.** A reader registers the ranges of the columns and row groups it is about to read; anything it reads beyond that plan — because the plan was narrowed, because a range was skipped, or because a different reader on the same file registered a different plan — misses and goes to the stream uncached.
3. **Reads repeated across rounds.** A reader that is `Reset()` and re-initialized with a new range set (late materialization reads the probe columns first and the payload columns afterwards) re-reads everything the new plan does not cover, including the bytes the previous round already read.
4. **No amortization and no visibility.** An uncovered read goes to the stream at exactly its own size, so a small read costs a full request, and it is only counted as a cache miss — the metrics do not show that a part of the remote IO of a scan is the same bytes fetched over and over.

The bytes in all these cases share the same properties: they are read repeatedly, by readers that share a cache, and they are small relative to the data. That is what a cache is for, so the miss path should not end at the underlying stream.

## Solution

Add a `FileBlockCache`, owned by `ReadAheadCache`, as a second level below the registered-range lookup: when no registered range covers a read, the block cache serves it at a fixed block granularity instead of letting it fall through to the stream. It is a general mechanism over the whole file, not tied to any particular region or reader.

- **Fixed-size blocks over the file.** The file is divided into blocks of `block_size` bytes and a read is served out of the single block containing it, so a read fetches its whole block once and every later read landing in that block — from any reader sharing the cache, in any round — is served from memory. The block granularity is also what amortizes the request cost of many small reads in the same neighbourhood.
- **Single flight.** A block is published (with its promise) before its fetch is dispatched, so concurrent readers of the same block wait for that one fetch instead of issuing their own. This is what collapses the N identical metadata reads of the parallel readers of a file into one.
- **Block alignment.** Blocks are counted from the end of the file, so block 0 is `[file_size - block_size, file_size)`. Two reasons: no block can then reach past EOF, and the metadata read that starts the reading of a columnar file (arrow reads exactly the last 64 KiB of a Parquet file as the footer) covers exactly one block instead of straddling two.
- **No eviction.** Once the capacity is reached `Read()` declines instead of replacing a block, which keeps every dispatched fetch reachable through its block so that `Release()` and the destructor can wait for the fetches still writing into the block buffers.
- **Transparent, always.** The cache is a pure optimization and never changes what the caller observes: a read it cannot serve — larger than a block, straddling two blocks, an exhausted capacity, or a failed block fetch — is declined and the caller reads the bytes itself. Since a block fetch reads more than was asked for, its failure in particular must not become the caller's error; a block whose fetch failed is simply not fetched again.
- **Configuration.** `CacheConfig` gains `block_size` (default 64 KiB, matching arrow's footer read size) and `block_cache_limit` (default 1 MiB per file). Zero for either disables the block cache, and so does an unknown file size.
- **Metrics.** `read-ahead-cache.block.{hits,hit-bytes}` for the reads served by a block and `read-ahead-cache.block.{fetches,fetch-bytes}` for the fetches issued for the blocks themselves, the latter also included in `read-ahead-cache.io.{count,bytes}` since they do go to the stream. Comparing the two pairs shows how many bytes the block granularity read on top of what was asked for.

## Anything else?

The block cache follows the lifetime rules of the cache that owns it: it survives `Reset()`, because the blocks cache the *file* rather than a round of registered ranges and a reader reusing the cache reads the same file again, and it is released by `ReleaseBuffers()`, because the file is not read again afterwards.

Reads that a registered range does cover are unaffected — the block cache is consulted only after the lookup over the registered ranges finds nothing — so the prefetch path keeps its exact-range granularity and only the previously uncached reads change behavior.

lucasfang:
### Purpose

Linked issue: close #273

`ReadAheadCache` caches exactly the ranges registered through `Init()`. A read that no registered range covers is reported as a miss and handed back to the caller, which reads it from the underlying stream — uncached, so the next reader asking for the same bytes pays for it again. This leaves every read outside the current prefetch plan uncached: the metadata reads a reader must issue *before* it can compute the ranges it wants to prefetch, the reads beyond a narrowed plan, and the reads repeated after a `Reset()` with a new range set. Since one `ReadAheadCache` is shared by all the readers `PrefetchFileBatchReaderImpl` creates for a data file and outlives a round of registered ranges, those bytes are fetched once per reader and once per round, each time a separate full-latency request.

This PR adds a `FileBlockCache` as a second level below the registered-range lookup: when no registered range covers a read, the block cache serves it at a fixed block granularity instead of letting it fall through to the stream. It is a general mechanism over the whole file — any reader, any offset, any round — with the metadata reads that precede a prefetch plan as its most visible beneficiary.

Changes:

- `src/paimon/common/utils/file_block_cache.h` / `.cpp` (new): a cache of fixed-size blocks of one file. A read is served out of the single block containing it, so the first read of a region fetches its whole block and every later read landing in that block is served from memory. A block is published together with its promise *before* its fetch is dispatched, so concurrent readers of the same block wait for that single fetch (`shared_future`) instead of issuing their own; the wait and the copy happen outside the lock. Blocks are keyed by index counted from the end of the file (block 0 is `[file_size - block_size, file_size)`), which keeps them disjoint by construction, keeps every block inside the file, and makes the 64 KiB footer read of arrow cover exactly one block instead of straddling two. Blocks are never evicted — once the capacity is reached `Read()` declines instead of replacing a block — which keeps every dispatched fetch reachable through its block so `Release()` and the destructor can wait for the fetches still writing into the block buffers. `Read()` returns `bool` and is transparent by contract: it declines, leaving `dest` untouched so the caller reads the bytes itself, when one block cannot serve the range, when the capacity is exhausted, or when the block fetch failed. A block fetch reads more than the caller asked for, so its failure must not become the caller's error; the failed block is left in place with its failed future and is not fetched again.
- `src/paimon/common/utils/read_ahead_cache.h` / `.cpp`: the constructor takes the `file_size` the blocks are indexed against and creates the block cache when the size is known and both block-cache options are non-zero. `Read()` consults the block cache when no registered range covers the request, before counting a miss. `Reset()` now clears only the prefetch state and zeroes the block counters, keeping the blocks (they cache the file, not a round of registered ranges) via the new private `ReleasePrefetchBuffers()`; `ReleaseBuffers()` releases the blocks as well, since the file is not read again. `CollectMetrics()` publishes the block counters and folds the block fetches into the io counters.
- `include/paimon/utils/prefetch_cache_config.h`: `CacheConfig` gains `GetBlockSize`/`SetBlockSize` (default `64 * 1024`, matching arrow's `kDefaultFooterReadSize`) and `GetBlockCacheLimit`/`SetBlockCacheLimit` (default `1024 * 1024`, per file). Zero for either disables the block cache.
- `src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp`: passes the data file size to the cache; a non-positive size means unknown and leaves the block cache off.
- `src/paimon/testing/utils/gated_async_input_stream.h` (new): an `InputStream` wrapper that holds the `ReadAsync` callbacks until `ReleaseAll()`, so tests can observe a cache while its fetches are in flight.
- `src/paimon/CMakeLists.txt`: registers the new source and test file.

New metrics: `read-ahead-cache.block.{hits,hit-bytes}` for the reads served by a block and `read-ahead-cache.block.{fetches,fetch-bytes}` for the fetches issued for the blocks themselves. A read is counted as a hit, a block hit or a miss, so `read.count = read.hits + block.hits + read.misses`. The block fetches also go to the underlying stream and are therefore included in `read-ahead-cache.io.{count,bytes}`, so comparing them with the block hit bytes shows how many bytes the block granularity added on top of the requested ones.

### Tests

- UT `src/paimon/common/utils/file_block_cache_test.cpp` (new, 9 cases): `TestServesRepeatedTailReads`, `TestClampsLowestBlockAtFileStart`, `TestSingleFlightForConcurrentReads` (two readers of the same block, one fetch), `TestExhaustedCapacityDeclinesNewBlocks`, `TestDeclinesStraddlingAndOversizedReads`, `TestDisabledByZeroCapacityOrBlockSize`, `TestResetCountersKeepsBlocksAndReleaseDropsThem`, `TestFetchErrorDeclinesTheReads` (a failed block is declined without being fetched again, the other blocks keep working), `TestFileSizeLargerThanTheFileOnlyLosesTheLastBlock` (a file size larger than the physical file only stops one block from being cached and never fails a read).
- UT `src/paimon/common/utils/read_ahead_cache_test.cpp`: adds `TestBlockCacheServesUncoveredReads`, `TestBlockCacheDeclinedReadIsAMiss`, `TestBlockCacheSurvivesReset`, `TestBlockCacheDisabled`, and covers the new metrics.
- UT `src/paimon/format/parquet/parquet_file_batch_reader_test.cpp`: adds `TestBlockCacheSharesTailReadsAcrossReaders`, an end-to-end check that the parallel readers of one file fetch their common metadata once.
- UT `src/paimon/common/io/cache_input_stream_test.cpp`: adapted to the new constructor, with the file size left unknown so the block cache stays off and the miss fallback of `CacheInputStream` is still exercised.
- Actual results: `cmake --build build --target unittest` → 28/28 ctest suites passed (134 s). `paimon-common-test --gtest_filter='TestFileBlockCache.*:TestReadAheadCache.*:CacheInputStreamTest.*'` → 37/37 passed. `paimon-parquet-format-test` 202/202, `paimon-orc-format-test` 117/117, `paimon-read-inte-test` 294/294, `paimon-read-inte-with-index-test` 68/68, `paimon-scan-and-read-inte-test` 490/490 passed. `git diff --check` clean and `pre-commit run --files <changed files>` all passed.

### API and Format

Public API: `include/paimon/utils/prefetch_cache_config.h` adds four `CacheConfig` accessors (`Get/SetBlockSize`, `Get/SetBlockCacheLimit`) with defaults, so existing callers are source and behavior compatible. `ReadAheadCache` is `PAIMON_EXPORT`ed but internal to the read path; its constructor gains a `file_size` parameter and every in-tree caller is updated. `FileBlockCache` is not part of the public headers.

Storage format and protocol: unchanged. This only affects how the read path fetches bytes, not what is written or read.

### Documentation

No new table option or user-facing configuration key. The behavior change is internal to the read-ahead cache and the new metrics are documented in `ReadAheadCacheMetrics`, next to the existing ones.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/include/paimon/utils/prefetch_cache_config.h
+++ b/include/paimon/utils/prefetch_cache_config.h
@@ -67,10 +67,41 @@ class PAIMON_EXPORT CacheConfig {
         pre_buffer_limit_ = pre_buffer_limit;
     }
 
+    /// Returns the granularity (in bytes) of the block cache entries serving the
+    /// small reads that the prefetched ranges do not cover.
+    uint64_t GetBlockSize() const {
+        return block_size_;
+    }
+
+    /// Sets the granularity (in bytes) of the block cache entries.
+    void SetBlockSize(uint64_t block_size) {
+        block_size_ = block_size;
+    }
+
+    /// Returns the maximum total size (in bytes) of the block cache entries of
+    /// one file. Zero disables the block cache.
+    uint64_t GetBlockCacheLimit() const {
+        return block_cache_limit_;
+    }
+
+    /// Sets the maximum total size (in bytes) of the block cache entries of one
+    /// file. Zero disables the block cache.
+    void SetBlockCacheLimit(uint64_t block_cache_limit) {
+        block_cache_limit_ = block_cache_limit;
+    }
+
  private:
     uint64_t range_size_limit_;
     uint64_t hole_size_limit_;
     uint64_t pre_buffer_limit_;
+    // Blocks are aligned to the END of the file, so a block never reaches past
+    // EOF. 64 KiB matches the footer read of the parquet reader (arrow's
+    // kDefaultFooterReadSize), which then covers exactly one block instead of
+    // straddling two.
+    uint64_t block_size_ = 64 * 1024;
+    // One block is enough for the metadata tail of a file; the limit only
+    // bounds the pathological case, as blocks are never evicted.
+    uint64_t block_cache_limit_ = 1024 * 1024;
 };
 
 }  // namespace paimon
--- a/include/paimon/utils/prefetch_cache_config.h
+++ b/include/paimon/utils/prefetch_cache_config.h
@@ -34,10 +34,8 @@ namespace paimon {
 /// ReadAheadCache to balance memory usage, I/O efficiency, and latency hiding.
 class PAIMON_EXPORT CacheConfig {
  public:
-    CacheConfig();
-    CacheConfig(uint64_t range_size_limit, uint64_t hole_size_limit, uint64_t pre_buffer_limit);
-
     /// Returns the maximum allowed size (in bytes) for a single cached range.
+    /// Defaults to 32 MiB.
     uint64_t GetRangeSizeLimit() const {
         return range_size_limit_;
     }
@@ -47,7 +45,8 @@ class PAIMON_EXPORT CacheConfig {
         range_size_limit_ = range_size_limit;
     }
 
-    /// Returns the maximum gap size (in bytes) considered mergeable between adjacent ranges.
+    /// Returns the maximum gap size (in bytes) considered mergeable between
+    /// adjacent ranges. Defaults to 8 KiB.
     uint64_t GetHoleSizeLimit() const {
         return hole_size_limit_;
     }
@@ -57,7 +56,8 @@ class PAIMON_EXPORT CacheConfig {
         hole_size_limit_ = hole_size_limit;
     }
 
-    /// Returns the maximum size to pre-buffer ahead of the current read position.
+    /// Returns the maximum size to pre-buffer ahead of the current read
+    /// position. Defaults to 256 MiB.
     uint64_t GetPreBufferLimit() const {
         return pre_buffer_limit_;
     }
@@ -68,7 +68,7 @@ class PAIMON_EXPORT CacheConfig {
     }
 
     /// Returns the granularity (in bytes) of the block cache entries serving the
-    /// small reads that the prefetched ranges do not cover.
+    /// small reads that the prefetched ranges do not cover. Defaults to 64 KiB.
     uint64_t GetBlockSize() const {
         return block_size_;
     }
@@ -79,7 +79,7 @@ class PAIMON_EXPORT CacheConfig {
     }
 
     /// Returns the maximum total size (in bytes) of the block cache entries of
-    /// one file. Zero disables the block cache.
+    /// one file. Zero disables the block cache. Defaults to 1 MiB.
     uint64_t GetBlockCacheLimit() const {
         return block_cache_limit_;
     }
@@ -91,9 +91,18 @@ class PAIMON_EXPORT CacheConfig {
     }
 
  private:
-    uint64_t range_size_limit_;
-    uint64_t hole_size_limit_;
-    uint64_t pre_buffer_limit_;
+    // The defaults are aligned with the reader's request granularity and with
+    // realistic data file sizes:
+    // - range_size_limit matches the parquet reader's 32 MiB request blocks
+    //   (Arrow ReadRangeCache's own range limit); a smaller limit cuts entries
+    //   below the request size, so a request can never be served from one piece.
+    // - pre_buffer_limit must exceed the LARGEST single read a reader issues
+    //   (coalesced column-chunk reads of ~128 MiB were observed): fetches are
+    //   only dispatched up to this window, so a request reaching past it can
+    //   never be served and falls back to a second fetch of the same bytes.
+    uint64_t range_size_limit_ = 32 * 1024 * 1024;
+    uint64_t hole_size_limit_ = 8 * 1024;
+    uint64_t pre_buffer_limit_ = 256 * 1024 * 1024;
     // Blocks are aligned to the END of the file, so a block never reaches past
     // EOF. 64 KiB matches the footer read of the parquet reader (arrow's
     // kDefaultFooterReadSize), which then covers exactly one block instead of
--- a/include/paimon/utils/prefetch_cache_config.h
+++ b/include/paimon/utils/prefetch_cache_config.h
@@ -104,9 +104,10 @@ class PAIMON_EXPORT CacheConfig {
     uint64_t hole_size_limit_ = 8 * 1024;
     uint64_t pre_buffer_limit_ = 256 * 1024 * 1024;
     // Blocks are aligned to the END of the file, so a block never reaches past
-    // EOF. 64 KiB matches the footer read of the parquet reader (arrow's
-    // kDefaultFooterReadSize), which then covers exactly one block instead of
-    // straddling two.
+    // EOF. 64 KiB is the granularity the reads no prefetched range covers are
+    // shared at: small enough that a metadata read at the tail of a file is
+    // served by one block instead of straddling two, large enough that a block
+    // fetch does not pull in much more than the reads ask for.
     uint64_t block_size_ = 64 * 1024;
     // One block is enough for the metadata tail of a file; the limit only
     // bounds the pathological case, as blocks are never evicted.
--- a/src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
+++ b/src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
@@ -215,6 +215,9 @@ Result<std::unique_ptr<PrefetchFileBatchReaderImpl>> PrefetchFileBatchReaderImpl
     if (batch_size <= 0) {
         return Status::Invalid("batch size should be greater than 0.");
     }
+    if (data_file_size < 0) {
+        return Status::Invalid("data file size should not be negative.");
+    }
     if (reader_builder == nullptr) {
         return Status::Invalid("reader_builder should not be nullptr.");
     }
@@ -238,9 +241,9 @@ Result<std::unique_ptr<PrefetchFileBatchReaderImpl>> PrefetchFileBatchReaderImpl
         }
         // The file size lets the cache align its blocks to the end of the file,
         // where the metadata the readers read before any range is registered
-        // lives. A non-positive size means unknown and disables the block cache.
-        const uint64_t file_size = data_file_size > 0 ? static_cast<uint64_t>(data_file_size) : 0;
-        cache = std::make_shared<ReadAheadCache>(input_stream, cache_config, file_size, pool);
+        // lives. A zero size means unknown and disables the block cache.
+        cache = std::make_shared<ReadAheadCache>(input_stream, cache_config,
+                                                 static_cast<uint64_t>(data_file_size), pool);
     }
     std::vector<std::future<Result<std::unique_ptr<FileBatchReader>>>> futures;
     for (uint32_t i = 0; i < prefetch_max_parallel_num; i++) {
--- a/src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
+++ b/src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
@@ -236,7 +236,11 @@ Result<std::unique_ptr<PrefetchFileBatchReaderImpl>> PrefetchFileBatchReaderImpl
         if (io_metrics) {
             input_stream = std::make_shared<MetricsInputStream>(input_stream, io_metrics);
         }
-        cache = std::make_shared<ReadAheadCache>(input_stream, cache_config, pool);
+        // The file size lets the cache align its blocks to the end of the file,
+        // where the metadata the readers read before any range is registered
+        // lives. A non-positive size means unknown and disables the block cache.
+        const uint64_t file_size = data_file_size > 0 ? static_cast<uint64_t>(data_file_size) : 0;
+        cache = std::make_shared<ReadAheadCache>(input_stream, cache_config, file_size, pool);
     }
     std::vector<std::future<Result<std::unique_ptr<FileBatchReader>>>> futures;
     for (uint32_t i = 0; i < prefetch_max_parallel_num; i++) {
--- a/src/paimon/common/utils/file_block_cache.h
+++ b/src/paimon/common/utils/file_block_cache.h
@@ -1,0 +1,145 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <cstdint>
+#include <future>
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+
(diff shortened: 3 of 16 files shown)


include/paimon/utils/prefetch_cache_config.h
zjw1111: Where is this parameterized constructor used?

zjw1111, line 35: Consider moving the implementation of `CacheConfig` into a separate `.cpp` file or into the `.h` file. It is currently placed in `read_ahead_cache.cpp`, which looks a bit odd.

zjw1111, line 65: Add default value descriptions for these parameters.

src/paimon/common/utils/file_block_cache_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+    EXPECT_TRUE(file.is_open());
+    file.write(kContent, sizeof(kContent) - 1);
+    EXPECT_FALSE(file.fail());
+    file.close();
+
+    Result<std::unique_ptr<FileSystem>> fs = FileSystemFactory::Get("local", path, {});
+    EXPECT_TRUE(fs.ok());
+    Result<std::unique_ptr<InputStream>> in = fs.value()->Open(path);
+    EXPECT_TRUE(in.ok());
+    return std::move(in).value();
```
zjw1111: code style

src/paimon/common/utils/read_ahead_cache_test.cpp
```diff
@@ -62,7 +65,7 @@ TestCacheEnv CreateTestFileAndCache(const std::string& filename, const std::stri
     auto in = std::move(in_result).value();
```
zjw1111: code style

src/paimon/testing/utils/gated_async_input_stream.h
```diff
@@ -0,0 +1,114 @@
...
+        pending_.push_back({buffer, size, offset, std::move(callback)});
+    }
+    Result<std::string> GetUri() const override {
+        return inner_->GetUri();
+    }
+    Result<int64_t> Length() const override {
+        return inner_->Length();
+    }
+
+    int AsyncReadCount() {
```
zjw1111: code style

src/paimon/testing/utils/gated_async_input_stream.h
```diff
@@ -0,0 +1,114 @@
...
+        char* buffer;
+        int64_t size;
+        int64_t offset;
+        std::function<void(Status)> callback;
+    };
+
+    std::shared_ptr<InputStream> inner_;
+    std::mutex mutex_;
+    std::vector<PendingRead> pending_;
+    int async_read_count_ = 0;
```
zjw1111: code style

include/paimon/utils/prefetch_cache_config.h
zjw1111: maybe can remove it?

src/paimon/common/reader/prefetch_file_batch_reader_impl.cpp
```diff
@@ -236,7 +236,11 @@ Result<std::unique_ptr<PrefetchFileBatchReaderImpl>> PrefetchFileBatchReaderImpl
         if (io_metrics) {
             input_stream = std::make_shared<MetricsInputStream>(input_stream, io_metrics);
         }
-        cache = std::make_shared<ReadAheadCache>(input_stream, cache_config, pool);
+        // The file size lets the cache align its blocks to the end of the file,
+        // where the metadata the readers read before any range is registered
+        // lives. A non-positive size means unknown and disables the block cache.
+        const uint64_t file_size = data_file_size > 0 ? static_cast<uint64_t>(data_file_size) : 0;
```
zjw1111: Under what circumstances could a negative value appear here? If we're concerned about a negative `data_file_size`, we should validate it globally at the very beginning.

src/paimon/common/utils/file_block_cache_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// Assert that the range is served out of a block with the expected content.
+void AssertServed(const ByteRange& range, const std::string& expected, FileBlockCache* cache) {
+    std::string dest(range.length, 'X');
+    ASSERT_TRUE(cache->Read(range, dest.data())) << expected;
+    EXPECT_EQ(expected, std::string_view(dest.data(), range.length));
```
zjw1111: code style

src/paimon/common/utils/read_ahead_cache_test.cpp
zjw1111: where uses `path` and `pool` in TestCacheEnv?

src/paimon/common/utils/read_ahead_cache.cpp
```diff
@@ -147,6 +152,9 @@ class ReadAheadCache::Impl {
     std::vector<std::atomic<bool>> is_cached_;
     std::vector<ByteRange> pending_ranges_;
     bool is_initialized_ = false;
+    // Caches the reads that no registered range covers, or null when the block
+    // cache is disabled. Owns its own locking and counters.
+    std::unique_ptr<FileBlockCache> block_cache_;
     // Statistics of the Read() requests issued to the cache, aggregated over
     // all streams sharing this cache.
     std::atomic<uint64_t> read_count_{0};
```
zjw1111: TODO: metrics refactor to a struct

src/paimon/common/utils/file_block_cache.h
```diff
@@ -0,0 +1,145 @@
...
+    uint64_t capacity_;
+    std::shared_ptr<MemoryPool> memory_pool_;
+    // Blocks are aligned, so keying them by index keeps them disjoint by
+    // construction and needs no ordering. A plain mutex is enough: only the
+    // reads that no prefetched range covers touch the map, and they are few.
+    mutable std::mutex mutex_;
+    std::unordered_map<uint64_t, std::shared_ptr<Block>> blocks_;
+    // Bytes held by blocks_, guarded by mutex_ and bounded by capacity_.
+    uint64_t cached_bytes_ = 0;
+    std::atomic<uint64_t> hits_{0};
```
zjw1111: TODO: metrics refactor to a struct

src/paimon/common/utils/file_block_cache.h
```diff
@@ -0,0 +1,145 @@
...
+    std::atomic<uint64_t> hit_bytes_{0};
+    std::atomic<uint64_t> fetches_{0};
+    std::atomic<uint64_t> fetch_bytes_{0};
```
zjw1111: A broader question: at the moment there are many layers of metrics across different levels of the read path — `ParquetMetrics`, `PrefetchMetrics`, `ReadAheadCache` metrics, `FileBlockCache` metrics. Have you looked into whether the meanings of these metrics overlap or conflict with each other, and whether the metric collection is accurate?

src/paimon/common/utils/file_block_cache.h
```diff
@@ -0,0 +1,145 @@
...
+/// costs the caching of that block, see Read().
+///
+/// A block is published before its fetch is dispatched, so concurrent readers of
+/// the same block wait for that one fetch instead of issuing their own.
+///
+/// Blocks are never evicted: once the capacity is reached Read() declines
+/// instead of replacing a block. That keeps every dispatched fetch reachable
+/// through its block, so Release() and the destructor can wait for the fetches
+/// still writing into the block buffers.
+class PAIMON_EXPORT FileBlockCache {
```
zjw1111, line 67: Please do a self-test and analysis of whether there are any lifecycle issues between `bytes` and `pool` inside `ReadAheadCache` and `FileBlockCache`, especially in asynchronous S3/OSS scenarios. Please do not use `GetDefaultPool()`, since it is a global static pool and may hide lifecycle problems even if they exist.

src/paimon/common/utils/file_block_cache.h
```diff
@@ -0,0 +1,145 @@
...
+namespace paimon {
+
+/// A cache of fixed-size blocks of one file, serving the reads that no
+/// prefetched range covers: a parquet reader reads the footer and the page index
+/// before any range can be registered, and every reader of the file reads the
+/// same bytes.
+///
+/// Blocks are aligned to the END of the file: block 0 is
+/// [file_size - block_size, file_size). The metadata of a parquet/orc file lives
+/// in its tail and arrow reads exactly the last 64 KiB as the footer, so an
```
zjw1111: > arrow reads exactly the last 64 KiB as the footer

means parquet or just arrow? what is the behavior or any bad case in orc and other format?

src/paimon/common/utils/file_block_cache.h
lucasfang: I checked the four layers against each other. For one data file read through a prefetch reader with the cache on:

| layer | scope | what it counts |
| --- | --- | --- |
| `parquet.read.*` (ParquetMetrics) | per sub-reader, summed over the `prefetch.parallelism` sub-readers of the file | the rows, batches and row groups the parquet reader produced, plus `storage-read-bytes` |
| `prefetch.*` (PrefetchMetrics) | per file, one prefetch reader | read ranges, batches produced/consumed/discarded, queue depth, reader-read and consumer-wait latencies |
| `io.*` (PrefetchIoMetrics) | per file, aggregated over every instrumented stream: the cache's stream plus one per sub-reader | the requests that actually reach the file system |
| `read-ahead-cache.*` | per cache, one per file | the requests the sub-readers issue to the cache (`read.*`, `block.hits`) and the fetches the cache issues itself (`io.*`, including `block.fetch*`) |

No two layers publish the same thing under the same name

src/paimon/common/utils/file_block_cache.h
lucasfang: ORC does not go through arrow, but it does go through this cache: the prefetch reader is format-agnostic (`AbstractSplitRead::CreateFileBatchReader` wraps every format except `blob`, `avro` and `mosaic`), and the ORC file is read by the ORC C++ library through `OrcInputStreamImpl`. Its tail reads are:

- The postscript read: the last `DIRECTORY_SIZE_GUESS = 16 KiB` of the file, in `orc/c++/src/Reader.cc`. That falls inside the last 64 KiB block, so all the readers of the file share one fetch for it, which is the same benefit parquet gets.
- The footer read, issued only when the postscript plus the footer do not fit into those 16 KiB: `footerSize` bytes at `fileLength - tailSize`, an offset and a size of its own. It can straddle two blocks or exceed one, and `CanServe()` declines both, so that read goes to the stream uncached, like any read one block cannot serve.

So the bad case for ORC is a lost caching opportunity for one read, never a wrong result or a failed read: `Read()` returns false with `dest` untouched and ORC reads the bytes itself. A format that issues no metadata read at all simply never asks the block cache for anything.

To answer the question behind it directly: yes, a read larger than `block_size`, or straddling two blocks, is not cached at all — `CanServe()` declines it by design, because one block serving one read is what keeps a served read to a single fetch and a single copy, and keeps a declined read free of partial-hit assembly. The block cache is the fallback for the uncovered reads of a file, whose granularity is set by `block_size` and whose total is bounded by `block_cache_limit`; it is not a general-purpose file cache, and a caller whose read is bigger than a block keeps reading that read from the stream, exactly as before this PR. Serving large uncovered reads would mean splitting a request across blocks, fetching each with its own single flight and assembling the copies, which I would rather keep out of this PR; I can file a follow-up issue if you want it.

Since none of this is specific to one format, the header comment no longer quotes the parquet or the ORC read sizes. It now states the mechanism — the fallback below the prefetched ranges, blocks aligned to the end of the file, one block per read and what is declined — and `CacheConfig::block_size_`'s comment states the granularity trade-off instead of naming a reader. The format-specific numbers are answered here, in the thread that asked for them.

zjw1111 (approved):
+1

lxy-9602 (approved):
+1

PR #274, opened 2026-09-02
feat(rest): access table data with the temporary credentials issued by the REST catalog 

### LINKED ISSUE #369: [Feature] Access table data with the temporary credentials issued by the REST catalog
## Search before asking

## Motivation

The C++ REST catalog can only read and write table data with the static credentials configured in the catalog options (`fs.oss.accessKeyId` and friends). A REST catalog that issues per-table temporary credentials (data tokens, e.g. DLF) has no way to have those credentials used for that table's data IO, and no way to keep them fresh before they expire. Callers that bring their own `FileSystem` also have no supported abstraction for sourcing credentials that expire.

## Solution

Add data-token support to the REST catalog: when `data-token.enabled=true`, the catalog loads a table's temporary credentials and serves a refreshing `FileSystem` for that table through `Catalog::GetTableFileSystem(identifier)`. `ReadContextBuilder`, `ScanContextBuilder` and `WriteContextBuilder` gain `WithCatalog(catalog, identifier)` so the table's schema and file system are resolved from the catalog in one call. The credentials are cached and reloaded before expiry by rebuilding the delegate file system keyed by the issued token, so a rotation of one table's credentials does not disturb the others.

Also add a generic `CredentialProvider` + `CredentialProviderFactory` extension point: a caller that brings its own `FileSystem` builds a provider with `CredentialProviderFactory::Get` and consults `GetCredentials()` at each access, then passes that file system in through `Catalog::Create` or a builder's `WithFileSystem`. The built-in file systems (`oss`, `s3`, `local`, `jindo`) keep signing with the static credentials of their own options.

## Anything else?

`data-token.enabled` defaults to false, and the new API is additive, so the change is backward compatible.

lucasfang:
### Purpose

Linked issue: close #369 

Let the REST catalog drive table-data IO with the per-table temporary credentials it issues (data tokens, e.g. DLF) instead of the static credentials in the catalog options. When `data-token.enabled=true`, `Catalog::GetTableFileSystem(identifier)` returns a `FileSystem` that reloads the table's credentials before they expire by rebuilding its delegate keyed by the issued token. `ReadContextBuilder`, `ScanContextBuilder`, `WriteContextBuilder` and `CommitContextBuilder` gain `WithCatalog(catalog, identifier)` to resolve the table's schema and file system from the catalog in one call. A generic `CredentialProvider` + `CredentialProviderFactory` extension point is added for callers that bring their own `FileSystem`; built-in file systems keep signing with the static credentials of their own options. `Catalog::Create` takes an optional `fs_scheme_to_identifier_map` so a catalog that resolves several URI schemes keeps routing each scheme to its backend, including the delegate file systems it rebuilds from per-table data tokens.

### Tests

Unit tests: `rest_credential_provider_test`, `rest_token_file_system_test`, `rest_catalog_test`, `rest_messages_test`, `resource_paths_test`, `dlf_auth_test`, `credential_provider_factory_test`. Context and operation coverage: `read_context_test`, `scan_context_test`, `table_read_test`, `table_scan_test`, `file_store_commit_test`, `file_store_write_test`, `system_table_test`, `format_table_test`, `file_system_catalog_test`. Verified locally: `paimon-rest-test` (150 tests) and `paimon-fs-test` (69 tests) pass; `pre-commit` and `git diff --check` are clean.

### API and Format

Additive public API under `include/`: new headers `paimon/fs/credential_provider.h` and `paimon/fs/credential_provider_factory.h`; `CredentialProvider::MergeOptionsWithCredentials`; `Catalog::GetTableFileSystem`; an optional trailing `fs_scheme_to_identifier_map` parameter on `Catalog::Create`; `CatalogOptions::DATA_TOKEN_ENABLED` and `CatalogOptions::DLF_OSS_ENDPOINT`; `WithCatalog` on the read/scan/write/commit context builders. Every new parameter is optional and defaults preserve the current behavior. No storage format or protocol change. `data-token.enabled` defaults to false, so existing callers are unaffected.

### Documentation

`docs/source/user_guide/catalog.rst` updated: reading, scanning, writing and committing through the catalog with `WithCatalog`, and a new "Authenticating with credentials of your own" section covering the `CredentialProvider` / `CredentialProviderFactory` path for a caller-supplied file system.

### Generative AI tooling

Generated-by: Qoder


JingsongLi (commented):
Source-level review focused on behavioral parity with Java RESTTokenFileIO (apache/paimon@d77efe0d2e3d15e33df0eb8211f833035e0959b8). Local runtime validation could not be completed.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -414,6 +432,23 @@ std::shared_ptr<FileSystem> RestCatalog::GetFileSystem() const {
...
+        return fs_;
+    }
+    // The credentials are issued for the data table, so a system table shares those of
+    // the table it belongs to.
+    PAIMON_ASSIGN_OR_RAISE(Identifier load_identifier, ToLoadIdentifier(identifier));
+    return table_fs_cache_->Get(
+        load_identifier.GetFullName(),
+        [this, &load_identifier](const std::string&) -> Result<std::shared_ptr<FileSystem>> {
+            std::shared_ptr<FileSystem> fs = std::make_shared<RestTokenFileSystem>(
+                api_, api_->GetMergedOptions(), load_identifier);
```
JingsongLi, line 447: **[P2] Preserve Java's token-keyed FileIO cache semantics**

Java's [RESTTokenFileIO](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-common/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java#L71-L82) caches backend FileIO instances by credential map and expiration, with idle expiration. This instead caches entire table-bound wrappers by their printable name. For example, `Identifier("a.b", "c")` and `Identifier("a", "b.c")` both use `a.b.c` and receive the same wrapper, which remains bound to the first table's token endpoint. Also, after dropping and recreating a table at a different storage location, a new lookup still receives the old wrapper and its unexpired credentials because DDL never invalidates this cache. Please preserve token-keyed backend reuse; if wrappers are additionally cached, use the complete Identifier and invalidate them on table changes, as Java's [CachingCatalog](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-core/src/main/java/org/apache/paimon/catalog/CachingCatalog.java) does.

src/paimon/rest/rest_token_file_system.cpp
```diff
@@ -0,0 +1,184 @@
...
+
+    RestToken token{MergeTokenOptions(response.GetToken()), response.GetExpiresAtMillis()};
+    // The credentials are the only file system options that change, so the file system is
+    // rebuilt from the catalog options with the credentials merged over them.
+    std::map<std::string, std::string> fs_options = catalog_options_;
+    for (const auto& [key, value] : token.token) {
+        fs_options[key] = value;
+    }
+    PAIMON_ASSIGN_OR_RAISE(CoreOptions core_options,
+                           CoreOptions::FromMap(fs_options, /*specified_file_system=*/nullptr));
```
JingsongLi, line 81: **[P2] Keep ValidToken independent of delegate initialization**

Java's [validToken()](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-common/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java#L296-L303) only refreshes and returns credentials, specifically so native engines can use their own file system. Here `ValidToken()` calls `Refresh()`, which runs the full `CoreOptions::FromMap()` before saving the token. That also initializes data/manifest format factories, so a native integration that only needs credentials and omits Avro can fail on the default `manifest.format=avro` even when the REST response is valid. Please separate token refresh from delegate construction, and let `ValidToken()` perform only the former. A token-only test without registered format factories would cover this distinction.

src/paimon/rest/rest_messages.cpp
```diff
@@ -353,6 +355,22 @@ void GetTableResponse::FromJson(const rapidjson::Value& obj) noexcept(false) {
...
+    obj.AddMember(rapidjson::StringRef(kFieldToken),
+                  RapidJsonUtil::SerializeValue(token_, allocator).Move(), *allocator);
+    obj.AddMember(rapidjson::StringRef(kFieldExpiresAtMillis),
+                  RapidJsonUtil::SerializeValue(expires_at_millis_, allocator).Move(), *allocator);
+    return obj;
+}
+
+void GetTableTokenResponse::FromJson(const rapidjson::Value& obj) noexcept(false) {
+    token_ = RapidJsonUtil::DeserializeKeyValue<std::map<std::string, std::string>>(
+        obj, kFieldToken, {});
```
JingsongLi, line 370: **[P2] Reject a missing or null token instead of falling back to catalog credentials**

This default makes a missing or `null` `token` indistinguishable from an explicitly empty object. With existing catalog storage credentials, a successful HTTP response that omits the token can therefore continue into delegate construction using those credentials. Java retains a null response field and fails when [mergeTokenWithCatalogOptions()](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-common/src/main/java/org/apache/paimon/rest/RESTTokenFileIO.java#L281-L293) copies it, so it does not turn this malformed response into fallback data access. Please require a non-null token object and test missing/null separately from an explicitly empty `{}`.

src/paimon/core/table/source/table_scan.cpp
```diff
@@ -97,6 +99,15 @@ class TableScanImpl {
...
         std::string branch = BranchManager::NormalizeBranch(core_options.GetBranch());
         auto snapshot_manager = std::make_shared<SnapshotManager>(fs, context->GetPath(), branch);
+        if (VersionManagedCatalog* versioned = AsVersionManaged(context->GetCatalog())) {
+            std::shared_ptr<Catalog> catalog = context->GetCatalog();
+            Identifier identifier = context->GetIdentifier().value();
+            // Keep the catalog alive while the scan uses its snapshot loader.
+            snapshot_manager->SetSnapshotLoader(
+                [catalog, versioned, identifier]() -> Result<std::optional<Snapshot>> {
+                    return versioned->LoadSnapshot(identifier);
+                });
```
SteNicholas: **Keep the catalog snapshot body when planning the initial scan**. Installing this loader does not make the default scan work when the latest snapshot exists only in the catalog. `FullStartingScanner::Scan()` calls `LatestSnapshotId()` (which now uses this loader), then calls `SnapshotManager::LoadSnapshot(id)`, which unconditionally reads `<table>/snapshot/snapshot-<id>` from the filesystem. Consequently both a default batch scan and the initial full streaming scan fail with `NotExist` even though the catalog returned a valid snapshot. The new `TestCatalogScanReadsSnapshotFromCatalog` leaves the fixture's snapshot file in place and therefore masks this failure. Please carry the returned `Snapshot` into the initial plan, and cover a catalog-held snapshot with no corresponding snapshot file.

src/paimon/rest/rest_token_file_system.cpp
```diff
@@ -0,0 +1,125 @@
...
+      fs_cache_(fs_cache != nullptr ? std::move(fs_cache) : CreateFileSystemCache()),
+      provider_(std::make_shared<RestCredentialProvider>(api, catalog_options, identifier,
+                                                         std::move(clock))) {}
+
+Result<std::shared_ptr<FileSystem>> RestTokenFileSystem::BuildFileSystem(
+    const RestToken& token) const {
+    std::map<std::string, std::string> fs_options = catalog_options_;
+    for (const auto& [key, value] : token.token) {
+        fs_options[key] = value;
+    }
```
SteNicholas: **Prevent bucket credentials from overriding the issued token**. This merge only replaces identically named keys. For an OSS catalog configured with `fs.oss.bucket.b.accessKeyId`/`accessKeySecret`, a token containing `fs.oss.accessKeyId`/`accessKeySecret`/`securityToken` leaves the bucket keys intact. `OssFileSystemFactory::FindOption()` prefers those bucket keys, so access to `oss://b/...` is signed with the old catalog key pair and the new STS token, which is an invalid credential combination; if a bucket-scoped security token is configured too, the issued credentials are ignored entirely. Please ensure the issued credentials take precedence after bucket-option resolution, and add coverage with bucket-scoped catalog credentials.

src/paimon/rest/rest_catalog.cpp
```diff
@@ -438,6 +453,23 @@ std::shared_ptr<FileSystem> RestCatalog::GetFileSystem() const {
     return fs_;
 }
 
+Result<std::shared_ptr<FileSystem>> RestCatalog::GetTableFileSystem(
+    const Identifier& identifier) const {
+    if (!data_token_enabled_) {
+        return fs_;
+    }
```
SteNicholas: **Preserve an explicitly supplied catalog filesystem**. When `Catalog::Create(..., custom_fs)` is combined with `data-token.enabled=true` (including when enabled by `/v1/config`), this method ignores `fs_` and always returns a new `RestTokenFileSystem`. The read/write/commit paths using `WithCatalog` therefore discard the caller's filesystem and its `CredentialProvider`: `BuildFileSystem()` reconstructs a built-in delegate from options with `specified_file_system=nullptr`. This breaks the newly documented `Catalog::Create` injection path, which promises to use the supplied filesystem as-is, and can send custom-scheme accesses to the default local backend. Please retain whether a filesystem was explicitly supplied and preserve that override here, with a test that enables data tokens while injecting a custom filesystem.

docs/source/user_guide/catalog.rst
```diff
@@ -233,6 +252,51 @@ so its restored file references are visible to the expiration operation.
    A branch the catalog holds without that directory, or one created while
    expiration runs, is not found: do not expire a table with such a branch, and
    serialize branch creation and expiration through the upstream coordinator.
+=======
```
SteNicholas: **Remove the leftover merge-conflict separators**.There are standalone `=======` conflict markers here and at line 363, followed by a duplicated trailing sentence at line 364. `git diff --check` fails on both markers, so the repository's required diff check does not pass. Please finish resolving these documentation hunks and remove the duplicated text.

src/paimon/rest/rest_token_file_system.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // issued credentials then win whichever bucket the table's data lives in.
+        size_t prefix_length = FileSystemOptionPrefixLength(key);
+        if (prefix_length != 0) {
+            std::string bucket_prefix = key.substr(0, prefix_length) + "bucket.";
+            std::string bucket_suffix = "." + key.substr(prefix_length);
+            for (auto it = fs_options.begin(); it != fs_options.end();) {
+                if (it->first.size() > bucket_prefix.size() + bucket_suffix.size() &&
+                    StringUtils::StartsWith(it->first, bucket_prefix) &&
+                    StringUtils::EndsWith(it->first, bucket_suffix)) {
+                    it = fs_options.erase(it);
```
SteNicholas: **Preserve bucket-scoped credentials supplied by the token itself.** The erase loop runs against `fs_options` after earlier entries from the same token have already been inserted. If the response contains both global OSS credentials and a complete `fs.oss.bucket.b.*` credential set, `std::map` iteration inserts `fs.oss.bucket.b.securityToken` before processing `fs.oss.securityToken`; the latter then erases the token's own bucket STS value. The bucket access key and secret survive, so the OSS factory signs with the bucket key pair and the global STS token, and authentication fails even with an empty catalog-options map. I reproduced this with the current merge and OSS lookup bodies. Please remove conflicting catalog keys before overlaying the complete token map, or explicitly preserve keys present in `token.token`, and add a mixed global/bucket token test.

src/paimon/rest/rest_token_file_system.cpp
```diff
@@ -0,0 +1,167 @@
...
+        // A file system resolves a bucket-scoped option such as
+        // "fs.oss.bucket.<bucket>.accessKeyId" ahead of the flat "fs.oss.accessKeyId" the
+        // token carries. Keeping a catalog's bucket-scoped credential would sign an access
+        // with that stale key pair together with the token's security token, an invalid
+        // combination, so drop the bucket-scoped variant of every option the token sets: the
+        // issued credentials then win whichever bucket the table's data lives in.
+        size_t prefix_length = FileSystemOptionPrefixLength(key);
+        if (prefix_length != 0) {
+            std::string bucket_prefix = key.substr(0, prefix_length) + "bucket.";
+            std::string bucket_suffix = "." + key.substr(prefix_length);
```
SteNicholas: **Apply temporary-token precedence across the OSS token aliases.** A token containing fresh `fs.oss.accessKeyId`, `fs.oss.accessKeySecret`, and `fs.oss.sessionToken` leaves an existing catalog `fs.oss.securityToken` (including its bucket-scoped variant) untouched, because this cleanup only matches the exact option suffix. `OssFileSystemFactory::Create()` reads `securityToken` first and consults `sessionToken` only when it is empty, so it combines the refreshed key pair with the old STS token and fails authentication. Both aliases are supported by the OSS backend. Please normalize the aliases or clear the catalog values of both when the issued credentials supply either, and test flat as well as bucket-scoped stale `securityToken` values.

src/paimon/rest/rest_token_file_system.cpp
```diff
@@ -0,0 +1,127 @@
...
+    std::map<std::string, std::string> fs_options = catalog_options;
+    for (const auto& [key, value] : token.token) {
+        fs_options[key] = value;
+    }
+    return fs_options;
+}
+
+Result<std::shared_ptr<FileSystem>> RestTokenFileSystem::BuildFileSystem(
+    const RestToken& token) const {
+    PAIMON_ASSIGN_OR_RAISE(CoreOptions core_options,
```
lszskye: Need parse `fs_scheme_to_identifier_map` here.

src/paimon/core/operation/read_context.cpp
lszskye, line 370: if catalog!= nullptr && `fs_scheme_to_identifier_map`!= nullptr  will fail here

lszskye: `fs_scheme_to_identifier_map` is also supported in `WriteContext`, please check `WriteContext` at the same time

lucasfang:
Hi @JingsongLi @SteNicholas @lszskye, thank you all for the thorough reviews. I've addressed the review comments in the latest updates, and the branch is ready for another pass. Could you please take a look when you have time and let me know if anything else needs attention? Thanks!

src/paimon/rest/rest_catalog.cpp
```diff
@@ -438,6 +460,34 @@ std::shared_ptr<FileSystem> RestCatalog::GetFileSystem() const {
...
+    // would keep serving the credentials of a dropped table to a table recreated at another
+    // location.
+    //
+    // The shared cache is keyed by the credentials alone and holds the file systems built
+    // from the catalog options, which every table agrees on, so a rotation rebuilds them.
+    const std::map<std::string, std::string>& catalog_options = api_->GetMergedOptions();
+    std::shared_ptr<RestCredentialProvider> provider =
+        std::make_shared<RestCredentialProvider>(api_, catalog_options, load_identifier);
+    return std::make_shared<RestTokenFileSystem>(std::move(provider), catalog_options,
+                                                 token_fs_cache_, fs_scheme_to_identifier_map_);
```
JingsongLi, line 488: **[P2] Keep external tables on their catalog-configured filesystem**

Please preserve the external-table distinction before constructing this wrapper. Java's [CatalogUtils.loadTable()](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogUtils.java#L343-L347) selects `externalFileIO` for external tables, and [RESTCatalog.getTable()](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-core/src/main/java/org/apache/paimon/rest/RESTCatalog.java#L379-L389) supplies the catalog-configured filesystem for that path. Here `GetTableResponse::IsExternal()` is never consulted, so an external native table opened through `WithCatalog()` with `data-token.enabled=true` requests `/token` before accessing storage. If the server only issues tokens for internal tables, the read fails even though the catalog storage credentials can access the external location. Please retain the external flag when resolving table IO and return the configured filesystem for external tables; a test whose external-table token endpoint rejects requests would cover this difference.

src/paimon/core/operation/read_context.cpp
```diff
@@ -306,6 +318,11 @@ Result<std::unique_ptr<ReadContext>> ReadContextBuilder::Finish() {
     if (impl_->format_table_ != nullptr) {
         // The table already answers each of these, and from a source this cannot see behind, so a
         // second answer is refused rather than silently dropped.
+        if (impl_->catalog_ != nullptr) {
+            return Status::Invalid(
+                "a format table carries the file system it was loaded through, so WithCatalog() "
+                "cannot be used with one");
```
JingsongLi, line 324: **[P2] Attach the token filesystem when loading internal format tables**

Please also wire the table-scoped filesystem into `RestCatalog::LoadFormatTable()`. That method still calls `FormatTable::Create(fs_, ...)`, and format-table scans, reads and writes use that stored catalog-wide filesystem. With `data-token.enabled=true` and storage access available only through the table token, the documented `ReadContextBuilder(format_table)` / `ScanContextBuilder(format_table)` paths therefore fail without ever requesting a token; adding `WithCatalog()` is rejected here. Java [selects the internal/external FileIO before dispatching to the format-table loader](https://github.com/apache/paimon/blob/d77efe0d2e3d15e33df0eb8211f833035e0959b8/paimon-core/src/main/java/org/apache/paimon/catalog/CatalogUtils.java#L347-L364), so internal format tables receive the refreshing token filesystem as well. Please select and store the appropriate filesystem during format-table loading, respecting the external-table distinction, and cover an internal format table whose storage requires its token.

JingsongLi (commented):
Re-reviewed 19d2189, focusing on consistency with the Java REST implementation. The three findings from my previous review are addressed: token-keyed backend reuse, token-only refresh, and rejecting missing/null token responses. Two table-IO integration gaps remain below. Validation: all 28 changed C++ translation units passed syntax checks using the existing dependency headers, and git diff --check passed. Runtime tests were not run locally.

JingsongLi (dismissed):
Approving this change. The external-table credential handling and internal format-table token filesystem integration noted in my latest review can be addressed in follow-up work; they are non-blocking for this PR.

lszskye:
+1

JingsongLi (approved):
+1

PR #277, opened 2026-09-02
perf(scan): reuse immutable snapshot read views

wangyong9999:
### Purpose

Linked issue: N/A

Repeated latest-snapshot batch scans currently resolve and parse the table schema, `LATEST` hint,
and snapshot JSON again. This is especially expensive for point-lookup engines, while a
process-global cache in paimon-cpp would impose freshness and lifecycle policy on every caller.

This change adds a caller-owned, immutable `SnapshotReadView` handoff:

`Plan -> SnapshotReadView -> ScanContextBuilder -> SnapshotReader`

The view retains the parsed snapshot and table schema, is bound to one normalized base-table path
and normalized branch, and has no global state, TTL, stale-read policy, singleflight, or quota
policy. It is only published and accepted for non-streaming, non-real-time `latest` /
`latest-full` batch scans. A read-optimized (`$ro`) scan uses the same base-table identity, so a
view can be shared safely between base-table and `$ro` planning while table and branch validation
remain enforced. Empty-table views are reusable as well.

For data-evolution global-index scans, an internally evaluated index captures the view before the
index lookup and the data scan consumes the same view, so row ids and data are planned against one
snapshot. A caller-supplied `GlobalIndexResult` keeps its legacy behavior unless the caller also
supplies the matching view; paimon-cpp does not invent snapshot provenance for an external result.

The existing `ReadContext::SetTableSchema` path is also honored by system-table reads, allowing a
`$ro` reader to reuse its caller-provided parsed schema instead of loading it again.

### Tests

- `pre-commit` on all changed files and `git diff --check`
- CMake/Ninja build of `paimon-core-test` and `paimon-global-index-test`
- Full debug `paimon-core-test` suite: 1,897/1,897 passed
- Full debug `paimon-global-index-test` suite: 132/132 passed
- Final focused core coverage: 18/18 passed
- Data-evolution global-index variants: 6/6 passed
- Fork workflows for pre-commit, Build and Test, and GCC 8 Build and Test passed at `e13ba033`

### API and Format

This adds the public `SnapshotReadView`, `Plan::GetSnapshotReadView()`, and
`ScanContextBuilder::WithSnapshotReadView()` APIs. `SnapshotReadView::TablePath()` identifies the
normalized base-table root; `Branch()` carries the normalized branch separately. Existing source
implementations remain compatible because `Plan::GetSnapshotReadView()` has a default
implementation. Consumers using a prebuilt shared library must rebuild against the matching
headers because `Plan`'s vtable and `ScanContext`'s layout change.

There is no storage-format, file-format, or wire-protocol change.

### Documentation

The new public API and its supported scan boundary are documented in the public headers. No new
configuration option is introduced.

### Generative AI tooling

Generated-by: Codex (GPT-5)

### DIFF
--- a/include/paimon/table/source/plan.h
+++ b/include/paimon/table/source/plan.h
@@ -23,6 +23,7 @@
 #include <optional>
 #include <vector>
 
+#include "paimon/table/source/snapshot_read_view.h"
 #include "paimon/table/source/split.h"
 
 namespace paimon {
@@ -34,5 +35,13 @@ class PAIMON_EXPORT Plan {
     virtual const std::vector<std::shared_ptr<Split>>& Splits() const = 0;
     /// Snapshot id of this plan, return `std::nullopt` if the table is empty.
     virtual std::optional<int64_t> SnapshotId() const = 0;
+    /// Immutable snapshot metadata used to build this plan.
+    ///
+    /// The returned view can be injected into a later batch scan to reuse the already resolved and
+    /// parsed snapshot. Only non-streaming, non-real-time latest-snapshot plans publish a reusable
+    /// view; other plans return `nullptr`.
+    virtual std::shared_ptr<const SnapshotReadView> GetSnapshotReadView() const {
+        return nullptr;
+    }
 };
 }  // namespace paimon
--- a/src/paimon/core/table/source/data_evolution_batch_scan.cpp
+++ b/src/paimon/core/table/source/data_evolution_batch_scan.cpp
@@ -25,6 +25,7 @@
 #include "paimon/core/global_index/global_index_scan_impl.h"
 #include "paimon/core/global_index/indexed_split_impl.h"
 #include "paimon/core/table/source/data_split_impl.h"
+#include "paimon/core/table/source/snapshot_read_view_impl.h"
 #include "paimon/core/utils/snapshot_manager.h"
 #include "paimon/global_index/bitmap_global_index_result.h"
 
@@ -44,6 +45,11 @@ DataEvolutionBatchScan::DataEvolutionBatchScan(
       executor_(executor) {}
 
 Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::CreatePlan() {
+    std::shared_ptr<const SnapshotReadView> snapshot_read_view =
+        snapshot_reader_->GetSnapshotReadView();
+    if (snapshot_read_view && !snapshot_read_view->SnapshotId()) {
+        return PlanImpl::EmptyPlan(snapshot_read_view);
+    }
     std::optional<std::vector<Range>> row_ranges;
     std::shared_ptr<GlobalIndexResult> final_global_index_result = global_index_result_;
     if (!final_global_index_result) {
@@ -52,6 +58,9 @@ Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::CreatePlan() {
             final_global_index_result = index_result;
             PAIMON_ASSIGN_OR_RAISE(row_ranges, index_result->ToRanges());
         }
+        // EvalGlobalIndex captures the latest snapshot before consulting its index. Refresh the
+        // local copy so both negative and positive results publish and reuse that exact view.
+        snapshot_read_view = snapshot_reader_->GetSnapshotReadView();
     } else {
         PAIMON_ASSIGN_OR_RAISE(row_ranges, final_global_index_result->ToRanges());
     }
@@ -59,7 +68,12 @@ Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::CreatePlan() {
         return batch_scan_->CreatePlan();
     }
     if (row_ranges.value().empty()) {
-        return PlanImpl::EmptyPlan();
+        if (snapshot_read_view && snapshot_read_view->SnapshotId()) {
+            return std::make_shared<PlanImpl>(snapshot_read_view->SnapshotId(),
+                                              std::vector<std::shared_ptr<Split>>(),
+                                              snapshot_read_view);
+        }
+        return PlanImpl::EmptyPlan(snapshot_read_view);
     }
     PAIMON_ASSIGN_OR_RAISE(RowRangeIndex row_range_index,
                            RowRangeIndex::Create(row_ranges.value()));
@@ -133,7 +147,8 @@ Result<std::shared_ptr<Plan>> DataEvolutionBatchScan::WrapToIndexedSplits(
         }
         indexed_splits.push_back(std::make_shared<IndexedSplitImpl>(data_split, expected, scores));
     }
-    return std::make_shared<PlanImpl>(data_plan->SnapshotId(), indexed_splits);
+    return std::make_shared<PlanImpl>(data_plan->SnapshotId(), indexed_splits,
+                                      data_plan->GetSnapshotReadView());
 }
 
 Result<std::shared_ptr<GlobalIndexResult>> DataEvolutionBatchScan::EvalGlobalIndex() const {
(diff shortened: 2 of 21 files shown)


wangyong9999:
@lxy-9602 A review-follow-up update is now pushed as `e13ba033`. The fork pre-commit check is green and its build matrices are running; the Apache workflows are waiting for approval. When convenient, could you please approve the workflows and review the PR?

The cache-ownership boundary remains unchanged: paimon-cpp exposes only a caller-owned immutable snapshot read view, while TTL, eviction, stale-read, singleflight, and quota policies remain outside paimon-cpp. No process-global cache or downstream SCM changes are included.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
```diff
@@ -145,26 +160,46 @@ Result<std::shared_ptr<GlobalIndexResult>> DataEvolutionBatchScan::EvalGlobalInd
...
-    const std::shared_ptr<SnapshotManager>& snapshot_manager =
-        snapshot_reader_->GetSnapshotManager();
-    if (const std::optional<int64_t>& snapshot_id = core_options_.GetScanSnapshotId()) {
-        PAIMON_ASSIGN_OR_RAISE(Snapshot loaded_snapshot,
-                               snapshot_manager->LoadSnapshot(snapshot_id.value()));
-        snapshot = std::move(loaded_snapshot);
-    } else {
-        PAIMON_ASSIGN_OR_RAISE(snapshot, snapshot_manager->LatestSnapshot());
+    StartupMode startup_mode = core_options_.GetStartupMode();
+    if (!(startup_mode == StartupMode::LatestFull() || startup_mode == StartupMode::Latest())) {
```
wangyong9999, line 165: Both branches end with the same GlobalIndexScanImpl::Create + Scan; only where the snapshot comes from differs. Resolve `snapshot` in the if/else and keep a single Create/Scan below it, which also removes the `index_scan` / `index_scan_impl` split.

src/paimon/core/table/system/read_optimized_system_table.cpp
```diff
@@ -26,14 +26,57 @@
...
 #include "paimon/defs.h"
 #include "paimon/read_context.h"
 #include "paimon/scan_context.h"
 #include "paimon/table/source/table_read.h"
 #include "paimon/table/source/table_scan.h"
 
 namespace paimon {
+namespace {
+
+class ReadOptimizedTableScan final : public TableScan {
```
wangyong9999: This wrapper, `Rebind` and `ReadOptimizedBranch` exist only so the view's path string says `$ro` instead of the base table path. If the view stays bound to the base table path and the system-table branch of `TableScan::Create` validates against `system_table_path->table_path`, all three go away, and a view from a base-table plan can feed a `$ro` scan as well (same snapshot, same schema).

src/paimon/core/table/source/snapshot/snapshot_reader.cpp
```diff
@@ -47,7 +88,16 @@ Result<std::shared_ptr<Plan>> SnapshotReader::Read() const {
...
         std::vector<std::shared_ptr<Split>> data_splits,
         GenerateSplits(snapshot, scan_mode_ != ScanMode::ALL, split_generator_, std::move(files)));
-    return std::make_shared<PlanImpl>(raw_plan->SnapshotId(), data_splits);
+    std::shared_ptr<const SnapshotReadView> read_view = snapshot_read_view_;
+    if (!read_view && table_schema_) {
+        const std::shared_ptr<SnapshotManager>& snapshot_manager = GetSnapshotManager();
+        read_view = SnapshotReadViewImpl::Create(
+            snapshot_manager->RootPath(), snapshot_manager->Branch(), snapshot, table_schema_);
+    }
+    if (read_view && read_view->SnapshotId() != raw_plan->SnapshotId()) {
```
wangyong9999: With a view injected, FullStartingScanner already pins WithSnapshot to that view's snapshot, and non-latest startup modes are rejected in NewDataTableScan, so this mismatch cannot occur. Drop the check. The view construction here and in EmptyPlan() is the same five lines; a small helper would cover both.

src/paimon/core/table/source/data_evolution_batch_scan.cpp
wangyong9999: Agreed. The snapshot is now resolved in the startup-mode branches, followed by one `GlobalIndexScanImpl::Create` and `Scan` path. The explicit/non-latest behavior is unchanged.

src/paimon/core/table/system/read_optimized_system_table.cpp
wangyong9999: Agreed. The view identity is now the normalized base-table path plus normalized branch. The system-table entry validates against the parsed base path and still accepts views only for `$ro`; the wrapper, `Rebind`, and branch helper are removed. I also added coverage showing that a base-table view feeds `$ro` without metadata reload, while branch/table isolation remains enforced.

src/paimon/core/table/source/snapshot/snapshot_reader.cpp
wangyong9999: Agreed. Under the accepted-view contract, `FullStartingScanner` pins the injected snapshot and non-latest modes are rejected, so the mismatch check was redundant. It is removed, and a private const helper now shares view construction between `Read()` and `EmptyPlan()` without introducing mutable state.

include/paimon/table/source/plan.h
```diff
@@ -34,5 +35,13 @@ class PAIMON_EXPORT Plan {
...
     /// Snapshot id of this plan, return `std::nullopt` if the table is empty.
     virtual std::optional<int64_t> SnapshotId() const = 0;
+    /// Immutable snapshot metadata used to build this plan.
+    ///
+    /// The returned view can be injected into a later batch scan to reuse the already resolved and
+    /// parsed snapshot. Only non-streaming, non-real-time latest-snapshot plans publish a reusable
+    /// view; other plans return `nullptr`.
+    virtual std::shared_ptr<const SnapshotReadView> GetSnapshotReadView() const {
+        return nullptr;
+    }
```
lxy-9602, line 45: Thanks for the work. I am not yet convinced that the performance benefit justifies the new public API and ABI impact.

A caller can already avoid resolving `LATEST` by specifying `scan.snapshot-id`, and can avoid schema metadata I/O through `SetTableSchema()`. Compared with that approach, `SnapshotReadView` mainly avoids:

- loading and deserializing the snapshot JSON;
- repeatedly deserializing the schema JSON passed through `SetTableSchema()`;
- some associated object construction.

The existing Paimon cache does not cache parsed snapshot or schema objects. However, when the underlying filesystem already caches metadata and small files, the remaining benefit may mostly be JSON parsing and object construction. It is unclear whether that is significant compared with manifest planning and data access.

Could we provide a benchmark for the intended point-lookup workload, comparing:

1. normal latest-snapshot scan;
2. `scan.snapshot-id` + `SetTableSchema()` + the existing cache;
3. `SnapshotReadView` + the same cache?

It would be helpful to report planning latency (P50/P99), metadata I/O/listing counts, and ideally end-to-end QPS. In particular, the comparison between cases 2 and 3 would isolate the incremental benefit that requires this new API.

wangyong9999: Thanks — the case 2 vs. case 3 comparison was the right boundary to test. I ran a controlled point-lookup planning benchmark at the current PR head (`e13ba033`) and the incremental hot-cache benefit is modest.

```mermaid
pie title Case 2 P50 share with warmed local metadata
    "Planning unchanged by the View — 1.965 ms" : 93.51
    "Eliminated by SnapshotReadView — 0.136 ms" : 6.49
```

### Setup

- Release build: GCC 8.3, `-O3 -DNDEBUG`.
- ORC primary-key fixture, full-PK equality predicate, bucket 0, one resulting split.
- The same warmed 64 MiB LRU cache and `scan.manifest-entry-cache.max-snapshots=8` in all three modes.
- 200 warm-up iterations, then 3,000 measured iterations per mode and scenario, interleaved round-robin and pinned to one CPU.
- The filesystem wrapper counted logical metadata method calls by path category. All three modes produced the same split count; after warm-up, manifest/index I/O was zero.

### Warm local metadata

| Mode | P50 | P99 | Single-thread planning ops/s | Snapshot calls | Schema calls | List calls |
|---|---:|---:|---:|---:|---:|---:|
| 1. normal latest scan | 2.119 ms | 2.234 ms | 470.9 | 4 | 3 | 1 |
| 2. `scan.snapshot-id` + `SetTableSchema()` | 2.101 ms | 2.211 ms | 474.5 | 7 | 0 | 0 |
| 3. `SnapshotReadView` | 1.965 ms | 2.078 ms | 507.5 | 0 | 0 | 0 |

The incremental case 2 → case 3 result is:

- P50: **-6.49%** (136 μs)
- P99: **-5.99%** (133 μs)
- single-thread planning throughput: **+6.95%**

This is planning throughput, not end-to-end serving QPS. Since the View does not change data-file access, adding the common read/decode work would make its relative end-to-end share smaller in this hot-metadata boundary.

### Controlled latency sensitivity

I also injected delay before each counted logical metadata operation to identify whether the benefit is local parsing/object construction or avoided filesystem round trips:

| Injected delay / metadata call | Case 2 P50 / P99 | Case 3 P50 / P99 | P50 / P99 reduction |
|---|---:|---:|---:|
| 0 | 2.101 / 2.211 ms | 1.965 / 2.078 ms | 6.49% / 5.99% |
| 100 μs | 3.152 / 3.241 ms | 1.961 / 2.026 ms | 37.78% / 37.48% |
| 1 ms | 9.455 / 9.563 ms | 1.962 / 2.023 ms | 79.24% / 78.85% |

The 100 μs and 1 ms rows are synthetic sensitivity tests, **not** measured HDFS latency. They show that the proposal becomes valuable when those seven logical metadata calls carry real RTT, but they do not establish that production boundary.

My conclusion is that the measured hot-cache incremental gain does not justify the current composite public API/ABI by itself. I will stop pursuing this version and close the PR. If this is revisited, the prerequisite should be a real remote-filesystem end-to-end measurement, and I would first explore a narrower immutable snapshot-metadata cache rather than exposing `SnapshotReadView` as a public cross-scan object.

lxy-9602: Thanks for your reply!

PR #280, opened 2026-09-04
feat(realtime): unify Arrow storage and support PK deletion vectors

lxy-9602:
### Purpose

Linked issue: #158 

This change unifies real-time storage and schema handling for append-only and primary-key tables, adds PK + deletion-vector real-time reads, and supports querying an explicit snapshot.
The main changes are:
- Introduce RealtimeSchemaLayout for user input, store write, commit, and query schemas.
- Use ArrowRealtimeStore for both append-only and primary-key tables.
- Support real-time scans with an explicit scan.snapshot-id. Snapshots older than the context’s durable progress are rejected because the required memory range may already have been reclaimed.
- Add public APIs to read all or per-partition-bucket durable recovery offsets from an exact snapshot, using table options or a supplied file system. Missing per-bucket progress returns -1.
- Support PK + deletion-vector reads by:
  - pushing value predicates into level > 0 files;
  - reading level-0 files and memory without predicate pushdown;
  - merging high-level, level-0, and memory readers in the final merge-on-read.

#### Writer lifecycle and recovery
- PrepareCommitWithProgress() seals current real-time data and returns its commit messages together with offset ranges.
- Before writer handoff or shutdown, all accepted data must be covered by a successful prepare. Closing with building or sealed-but-unprepared data returns an error and invalidates the RealtimeContext.
- Recovery recreates both RealtimeContext and FileStoreWrite from the durable offset stored in the committed snapshot.
- A Write, Seal, or PrepareCommitWithProgress failure must be treated as terminal by upstream: discard the writer and context, then replay from the previous durable checkpoint. This follows the Java-style fail-and-recreate recovery contract instead of retrying partially mutated writer state.

Note: The previous implementation allowed a writer containing only an unsealed building segment to be closed while keeping its context reusable. This can be unsafe because a replacement writer cannot enumerate or prepare that segment if there is no more data write to that bucket. This change intentionally tightens the lifecycle validation to avoid silently losing commit progress.

### Tests
Coverage includes explicit snapshot scans, PK + deletion-vector merge-on-read, level-aware predicate pushdown, multiple level-0 runs, writer close validation, context recovery, and pinned read-view lifetime.

### API and Format

API impact:

- Adds `FileStoreWrite::Seal()`.
- Real-time scans now support the existing `scan.snapshot-id` option; no new scan API is introduced.
- Add public APIs to read all or per-partition-bucket durable recovery offsets from an exact snapshot.
- `RealtimeCommitProgress` may represent one or more sealed segments.
- `FileStoreWrite::Close()` now returns an error and invalidates its `RealtimeContext` when building or sealed data has not been covered by a successful `PrepareCommitWithProgress()`.
- After a write, seal, or prepare failure, upstream must discard the writer and context and recover from the durable snapshot offset.
- Normalizes the primary-key `RealtimeStore` write schema to:
  `[_SEQUENCE_NUMBER, _VALUE_KIND, _REALTIME_OFFSET, user fields]`.

### Generative AI tooling


zjw1111 (commented):
Found one correctness issue with explicit snapshot validation.

src/paimon/core/table/source/table_scan.cpp
```diff
@@ -255,8 +255,12 @@ Status ValidateRealtimeScan(const TableSchema& table_schema, const CoreOptions&
         return Status::Invalid("real-time union read does not support global index splits");
     }
     StartupMode startup_mode = core_options.GetStartupMode();
-    if (!(startup_mode == StartupMode::LatestFull() || startup_mode == StartupMode::Latest())) {
-        return Status::Invalid("real-time union read requires the latest snapshot");
+    const bool specified_snapshot =
```
zjw1111, line 339: [P1] Reject a nonexistent explicit snapshot

When the table has no disk snapshots, `StaticFromSnapshotStartingScanner` returns `NoSnapshot`. The resulting empty plan has no `SnapshotId`, so `RealtimeTableScan::LoadCommittedOffsets` treats it as offset 0 and unions all current memory rows. As a result, setting any nonexistent `scan.snapshot-id` can succeed and return uncommitted data while the plan's snapshot ID is empty.

Please reject this case when an explicit snapshot was requested (or make the static scanner fail when no snapshot exists), and add coverage for an empty table with in-memory rows plus `scan.snapshot-id`. Java, Python, and Rust all reject an explicit ID that does not exist.

include/paimon/realtime/realtime_snapshot_offsets.h
```diff
@@ -0,0 +1,63 @@
...
+ public:
+    RealtimeSnapshotOffsets() = delete;
+    ~RealtimeSnapshotOffsets() = delete;
+
+    /// Loads every partition-bucket offset recorded by an exact snapshot.
+    ///
+    /// An empty branch selects the main branch. A non-null `file_system` takes precedence;
+    /// otherwise it is resolved by `CoreOptions` from `options`. A snapshot without real-time
+    /// progress returns an empty map.
+    static Result<RealtimeOffsetMap> ReadAll(const std::string& table_path,
```
zjw1111: ReadAllOffsets

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #281, opened 2026-09-04
fix(aggregate): preserve listagg result ownership

ChaomingZhangCN:
### Purpose

Linked issue: N/A

`FieldListaggAgg` returned a `std::string_view` backed by a mutable member string shared by all calls to the same aggregator instance.

The merge path reuses one field aggregator across primary keys, while `SortMergeReaderBatchProducer` may retain results for multiple keys before converting the batch to Arrow. As a result, aggregating a later key could overwrite or invalidate the string data referenced by an earlier key, causing cross-key data corruption.

This change:

- builds each concatenated result independently;
- stores the result in a pool-backed `BinaryString`;
- removes the shared mutable result buffer;
- preserves the lifetime of previously returned aggregation results.

### Tests

Added:

- `FieldListaggAggTest.TestResultOwnershipAcrossAggregations` for the aggregator result-lifetime contract;
- `WriteAndReadInteTest.TestPKListAggPreservesResultsAcrossKeys` for the real PK merge path in `write_and_read_inte_test.cpp`.

The integration test creates a PK aggregation table, writes two keys over two commits, and verifies that each key retains its independently accumulated `listagg` value.

Before the production fix, the integration test fails deterministically for all three enabled formats (Parquet, ORC, and Avro):

```text
Expected: alpha,beta
Actual:   one,two\0\0\0
```

After the fix:

- `paimon-write-and-read-inte-test` builds successfully;
- `TestPKListAggPreservesResultsAcrossKeys`: 3/3 format variants passed;
- neighboring `TestPKSimple`: 3/3 format variants passed;
- `paimon-core-test` builds and links successfully on macOS arm64 after replacing the non-portable BIGINT test literals;
- clang-format, cpplint, codespell, and `git diff --check` passed.

The focused `FieldListaggAggTest.*` execution is still blocked in this local build by an existing macOS static/shared plugin registration issue (`Could not find a FileFormatFactory implementation class for format 'parquet'`). This does not affect the integration-test target above.

### API and Format

### Documentation

No documentation changes are required.

### Generative AI tooling

### DIFF
--- a/src/paimon/core/mergetree/compact/aggregate/field_listagg_agg_test.cpp
+++ b/src/paimon/core/mergetree/compact/aggregate/field_listagg_agg_test.cpp
@@ -47,13 +47,13 @@ class FieldListaggAggTest : public testing::Test {
 TEST_F(FieldListaggAggTest, TestSimple) {
     ASSERT_OK_AND_ASSIGN(auto agg, MakeAgg());
     auto ret = agg->Agg(std::string_view("hello"), std::string_view(" world")).value();
-    ASSERT_EQ(DataDefine::GetVariantValue<std::string_view>(ret), "hello, world");
+    ASSERT_EQ(DataDefine::GetStringView(ret), "hello, world");
 }
 
 TEST_F(FieldListaggAggTest, TestDelimiter) {
     ASSERT_OK_AND_ASSIGN(auto agg, MakeAgg("-"));
     auto ret = agg->Agg(std::string_view("user1"), std::string_view("user2")).value();
-    ASSERT_EQ(DataDefine::GetVariantValue<std::string_view>(ret), "user1-user2");
+    ASSERT_EQ(DataDefine::GetStringView(ret), "user1-user2");
 }
 
 TEST_F(FieldListaggAggTest, TestNull) {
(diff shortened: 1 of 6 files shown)


src/paimon/core/mergetree/compact/aggregate/field_listagg_agg_test.cpp
```diff
@@ -148,39 +160,39 @@ TEST_F(FieldListaggAggTest, TestDistinctIgnoresBlankTokens) {
...
 }
 
 TEST_F(FieldListaggAggTest, TestDistinctFalse) {
     ASSERT_OK_AND_ASSIGN(auto agg, MakeAgg(";", false));
 
     // "a;b" + "b;c" -> "a;b;b;c" (no dedup)
     auto ret = agg->Agg(std::string_view("a;b"), std::string_view("b;c")).value();
-    ASSERT_EQ(DataDefine::GetVariantValue<std::string_view>(ret), "a;b;b;c");
+    ASSERT_EQ(DataDefine::GetStringView(ret), "a;b;b;c");
 }
```
lxy-9602, line 196: Nice catch! Was this issue exposed by an integration test? Could you add an integration test in `write_and_read_inte_test.cpp` to help prevent similar issues in the future? Thanks!

ChaomingZhangCN: ok!

ChaomingZhangCN: Added `TestPKListAggPreservesResultsAcrossKeys` in `write_and_read_inte_test.cpp` in 76aee1a. It writes two PKs across two commits and verifies `alpha,beta` / `one,two`; it fails deterministically before the fix and passes for Parquet, ORC, and Avro after the fix. Thanks!

lxy-9602 (approved):
+1

PR #282, opened 2026-09-04
fix(core): resolve ScanContextBuilder executor lazily in Finish()

### LINKED ISSUE #287: [Bug] ScanContextBuilder spawns eight executor threads per builder even when an executor is supplied
### Search before asking

### Paimon-cpp version

main (7351dad0)

### Minimal reproduce step

1. Build a `ScanContextBuilder` per point lookup, as serving code does, and call `Finish()`.
2. Trace thread creation, for example `strace -f -c -e trace=clone` around the lookups.

Each `ScanContextBuilder` costs eight `clone` calls: `ScanContextBuilder::Impl` calls `CreateDefaultExecutor()` in its member initializer and again in `Reset()`, which `Finish()` invokes right after building the context, and every `DefaultExecutor` spawns `DEFAULT_EXECUTOR_THREAD_COUNT` (4) worker threads in its constructor and joins them on destruction. This happens even when the caller supplies its own executor through `WithExecutor()`. Measured on a point-lookup path, about nine `clone` calls per lookup, eight of them from this builder.

Constructing or resetting a builder should not create threads. A context that is given an executor through `WithExecutor()` should create none at all, and the default executor's workers should only exist once the scan actually submits work.

### Anything else?

`ReadContextBuilder` already resolves its default executor lazily in `Finish()`; `ScanContextBuilder` should behave the same. Falling back to the process-wide `GetGlobalDefaultExecutor()` is not an option: it sizes its pool by CPU count and `FileStoreScan::ReadFileEntries()` blocks in `CollectAll()` on the same pool, which can deadlock once every worker is inside a scan.

- [x] I'm willing to submit a PR! (#282)

wangyong9999:
### Purpose

Linked issue: close #287

`ScanContextBuilder::Impl` eagerly calls `CreateDefaultExecutor()` twice per builder
lifetime: once in the member initializer and once more in `Reset()`, which `Finish()`
invokes right after building the context. `DefaultExecutor` spawns
`DEFAULT_EXECUTOR_THREAD_COUNT` (4) worker threads in its constructor and joins them on
destruction, so every `ScanContextBuilder` costs eight thread creations even when the
caller supplies its own executor through `WithExecutor()`.

Point-lookup style serving code builds one `ScanContextBuilder` per key. Measured with
`strace -f -c -e trace=clone` on such a path, each lookup performed about nine `clone`
calls and eight of them came from this builder.

This change keeps every context on a `DefaultExecutor` of its own (no process wide
executor is introduced) and removes the thread churn in two places:

- `DefaultExecutor` starts its worker threads on the first `Add()` instead of in the
  constructor. An executor that never receives a task never spawns a thread, and
  shutting down or destroying such an executor only cleans up state. Executors that do
  receive tasks behave as before.
- `ScanContextBuilder::Impl::executor_` defaults to null and `Reset()` only clears it.
  `Finish()` resolves the executor: the one set through `WithExecutor()` if any,
  otherwise a fresh `CreateDefaultExecutor()` owned by the returned context. This
  matches what `ReadContextBuilder` already does.

With both changes a lookup that supplies its own executor creates no thread at all,
and one that relies on the default scan executor creates the four workers once, when
the scan actually submits work.

### Tests

- `DefaultExecutorTest.TestWorkersStartOnFirstTask` (new): `CreateDefaultExecutor(4)`
  creates no thread; four appear after the first tasks run; all are gone after the
  executor is destroyed (thread count read from `/proc/self/task`, Linux only).
- `DefaultExecutorTest.TestShutdownWithoutTasks` (new): `ShutdownNow()` and
  destruction of an executor that never ran a task.
- `ScanContextTest.TestDefaultExecutorIsCreatedPerContext` (new): contexts built
  without `WithExecutor()` get distinct non-null executors, none of them the global one, and an executor set on the
  builder only applies to the `Finish()` that follows it.
- All 12 `DefaultExecutorTest.*` cases (10 existing + 2 new) and the 4
  `ScanContextTest.*` cases pass locally in a Release build (`paimon-common-test`,
  `paimon-core-test`).

### API and Format

No public API or storage format change. `DefaultExecutor` worker threads now start on
the first submitted task rather than at construction; this is only observable through
thread counts.

### Documentation

No.

### Generative AI tooling

### DIFF
--- a/src/paimon/common/executor/default_executor_test.cpp
+++ b/src/paimon/common/executor/default_executor_test.cpp
@@ -26,6 +26,10 @@
 #include <thread>
 #include <vector>
 
+#ifdef __linux__
+#include <dirent.h>
+#endif
+
 #include "gtest/gtest.h"
 #include "paimon/common/executor/future.h"
 #include "paimon/executor.h"
@@ -34,6 +38,72 @@
 #include "paimon/testing/utils/testharness.h"
 
 namespace paimon::test {
+
+#ifdef __linux__
+// Number of threads of the current process according to /proc.
+int32_t CountProcessThreads() {
+    DIR* dir = opendir("/proc/self/task");
+    if (dir == nullptr) {
+        return -1;
+    }
+    int32_t count = 0;
+    while (struct dirent* entry = readdir(dir)) {
+        if (entry->d_name[0] != '.') {
+            ++count;
+        }
+    }
+    closedir(dir);
+    return count;
+}
+#endif
+
+TEST(DefaultExecutorTest, TestWorkersStartOnFirstTask) {
+#ifdef __linux__
+    int32_t threads_before = CountProcessThreads();
+    ASSERT_GT(threads_before, 0);
+#endif
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<Executor> executor, CreateDefaultExecutor(4));
+    ASSERT_EQ(4u, executor->GetThreadNum());
+#ifdef __linux__
+    // Constructing the executor does not spawn any worker thread.
+    ASSERT_EQ(threads_before, CountProcessThreads());
+#endif
+
+    std::atomic<int64_t> sum = {0};
+    std::vector<std::future<void>> futures;
+    for (int32_t index = 0; index < 8; ++index) {
+        futures.push_back(Via(executor.get(), [&sum]() { sum++; }));
+    }
+    Wait(futures);
+    ASSERT_EQ(8, sum.load());
+#ifdef __linux__
+    ASSERT_EQ(threads_before + 4, CountProcessThreads());
+#endif
+    executor.reset();
+#ifdef __linux__
+    // A joined worker may trail in /proc for a moment, so poll briefly.
+    int32_t threads_after = CountProcessThreads();
+    for (int32_t i = 0; i < 100 && threads_after != threads_before; ++i) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        threads_after = CountProcessThreads();
+    }
+    ASSERT_EQ(threads_before, threads_after);
+#endif
+}
+
+TEST(DefaultExecutorTest, TestShutdownWithoutTasks) {
+    // Shutting down or destroying an executor that never ran a task must not
+    // block or touch workers that were never started.
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<Executor> executor, CreateDefaultExecutor(2));
+    executor->ShutdownNow();
+    std::atomic<bool> ran = {false};
+    executor->Add([&ran]() { ran = true; });
+    std::this_thread::sleep_for(std::chrono::milliseconds(50));
+    ASSERT_FALSE(ran.load());
+    executor.reset();
+    std::unique_ptr<Executor> idle_executor = CreateDefaultExecutor();
+    idle_executor.reset();
+}
 
 TEST(DefaultExecutorTest, TestViaVoidFunc) {
     auto executor = GetGlobalDefaultExecutor();
--- a/src/paimon/common/executor/default_executor_test.cpp
+++ b/src/paimon/common/executor/default_executor_test.cpp
@@ -55,18 +55,33 @@ int32_t CountProcessThreads() {
     closedir(dir);
     return count;
 }
+
+// A worker joined by an earlier test can trail in /proc for a moment, so take
+// the count only once two consecutive reads agree.
+int32_t StableProcessThreadCount() {
+    int32_t last = CountProcessThreads();
+    for (int32_t i = 0; i < 100; ++i) {
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+        int32_t current = CountProcessThreads();
+        if (current == last) {
+            return current;
+        }
+        last = current;
+    }
+    return last;
+}
 #endif
 
 TEST(DefaultExecutorTest, TestWorkersStartOnFirstTask) {
 #ifdef __linux__
-    int32_t threads_before = CountProcessThreads();
+    const int32_t threads_before = StableProcessThreadCount();
     ASSERT_GT(threads_before, 0);
 #endif
     ASSERT_OK_AND_ASSIGN(std::unique_ptr<Executor> executor, CreateDefaultExecutor(4));
     ASSERT_EQ(4u, executor->GetThreadNum());
 #ifdef __linux__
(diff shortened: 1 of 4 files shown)


src/paimon/core/operation/scan_context.cpp
```diff
@@ -178,13 +179,14 @@ Result<std::unique_ptr<ScanContext>> ScanContextBuilder::Finish() {
     if (impl_->path_.empty()) {
         return Status::Invalid("cannot scan with empty table path");
     }
+    std::shared_ptr<Executor> executor =
+        impl_->executor_ ? impl_->executor_ : GetGlobalDefaultExecutor();
```
lxy-9602: Please do not use `GlobalDefaultExecutor`. It creates a number of threads proportional to the number of CPU cores and may also introduce potential locking issues.

wangyong9999: One concrete failure mode is `FileStoreScan::ReadFileEntries()`: it submits manifest work to the context executor and then blocks in `CollectAll()`. If all global workers are already running scans, every worker can wait for child work queued to the same pool and deadlock. Please keep the fallback lazy but private (`CreateDefaultExecutor()` in `Finish()`), while leaving `Reset()` null; that removes the eager thread creation without changing executor isolation.

src/paimon/core/operation/scan_context_test.cpp
```diff
@@ -101,4 +101,25 @@ TEST(ScanContextTest, TestSetOptionsOverridesAddedOptions) {
     ASSERT_EQ(expected_options, ctx->GetOptions());
 }
 
+TEST(ScanContextTest, TestDefaultExecutorIsGlobalSingleton) {
```
wangyong9999: Once the fallback becomes a private `CreateDefaultExecutor()` (per the thread above), this test pins the opposite of the new behavior. Worth flipping it to: `Finish()` without `WithExecutor()` yields a non-null executor that is not `GetGlobalDefaultExecutor()` and differs between two contexts; an executor passed to `WithExecutor()` is used for the next `Finish()` only. That checks the isolation the maintainer asked for instead of the singleton.

src/paimon/core/operation/scan_context.cpp
wangyong9999: Thanks, agreed: a CPU-proportional shared pool is the wrong tool for a serving path, and `FileStoreScan::ReadFileEntries()` blocking in `CollectAll()` on that same pool could deadlock once every worker is inside a scan. Pushed 3475256, which drops `GetGlobalDefaultExecutor()` entirely:

- `DefaultExecutor` now starts its workers on the first `Add()` instead of in the constructor. An executor that never receives a task never spawns a thread, and shutting down or destroying such an executor only cleans up state; executors that do run tasks behave exactly as before. `workers_` stays guarded by the existing `state_->mutex`, no new lock.
- `ScanContextBuilder::Finish()` falls back to a fresh `CreateDefaultExecutor()` owned by the returned context, the same as `ReadContextBuilder` does. Nothing is shared across contexts.

Net effect on the per-lookup path: constructing or resetting a builder creates no thread (previously 8), a context that supplies its own executor creates none at all, and the four default workers only appear once the scan actually submits work. Covered by `DefaultExecutorTest.TestWorkersStartOnFirstTask` / `TestShutdownWithoutTasks` and `ScanContextTest.TestDefaultExecutorIsCreatedPerContext`; the PR description is updated to match.

src/paimon/common/executor/default_executor_test.cpp
```diff
@@ -35,6 +39,72 @@
...
+
+    std::atomic<int64_t> sum = {0};
+    std::vector<std::future<void>> futures;
+    for (int32_t index = 0; index < 8; ++index) {
+        futures.push_back(Via(executor.get(), [&sum]() { sum++; }));
+    }
+    Wait(futures);
+    ASSERT_EQ(8, sum.load());
+#ifdef __linux__
+    ASSERT_EQ(threads_before + 4, CountProcessThreads());
```
wangyong9999: `add_test` registers the whole `paimon-common-test` binary, so every `DefaultExecutorTest.*` (and whatever ran before it) shares this process. A worker joined by the previous test can still show up in `/proc/self/task` when `threads_before` is read at line 62 and be gone by here, and the exact `== threads_before + 4` / `== threads_before` checks then fail on an unrelated off-by-one. Capture the baseline the same way the tail does (poll until two reads agree), and relax the comparisons to `>= threads_before + 4` after the tasks and `<= threads_before` after reset; that still proves "no thread at construction, four after the first task, none after destruction" without depending on other tests being quiet.

wangyong9999: Good catch, and it is not hypothetical: with the exact comparisons the first local run of this test failed on precisely that window (a worker joined by the previous executor was still listed in `/proc/self/task`). Addressed in c2d25dc7:

- the baseline is now taken with `StableProcessThreadCount()`, which returns only once two consecutive reads agree;
- the checks are `<= threads_before` after construction, `>= threads_before + 4` after the first tasks, and `<= threads_before` after destruction (the last one polls briefly, as before).

Validated by running the whole `paimon-common-test` binary, the same way `add_test` registers it, plus three filtered `DefaultExecutorTest.*` runs.

src/paimon/core/operation/scan_context_test.cpp
wangyong9999: Done. `TestDefaultExecutorIsGlobalSingleton` was replaced by `TestDefaultExecutorIsCreatedPerContext` in 3475256: `Finish()` without `WithExecutor()` yields a non-null executor that differs between two contexts, and an executor passed to `WithExecutor()` applies to the next `Finish()` only. 5ca87e38 adds the remaining piece, an explicit `ASSERT_NE(GetGlobalDefaultExecutor(), ...)` for both contexts, so a regression back to the singleton is caught directly. `ScanContextTest.*` passes locally.

SteNicholas:
Please create new isssue for each pull request. Meanwhile, please bind your email to your github account.

wangyong9999:
Thanks for the pointers, both done:

- Opened #287 for this change and linked it in the description.
- The commits carried an author email that is not attached to any GitHub account. I rewrote the four commits with the address bound to this account and rebased them onto the current `main` (c65cd40); the content is unchanged.

wangyong9999:
Hi @lucasfang , could u pls take a look when convenient, thanks!

lucasfang:
LGTM

lxy-9602 (approved):
+1

PR #284, opened 2026-09-05
fix(scan): share bucket pruning across append and key-value scans

### LINKED ISSUE #288: [Bug] Fix predicate bucket pruning for append scans and rescaled KV reads
### Search before asking

- [x] I searched the existing issues and found no duplicate.

### Paimon-cpp version

Upstream main at `ead92078` (before PR #284).

### Minimal reproduce step

1. Write a fixed-bucket primary-key table with two buckets.
2. Read the same snapshot with the current `bucket` option set to four and an equality predicate covering the bucket key. Choose a key whose computed bucket differs between the two counts. Do not set an explicit bucket filter.
3. Compare the returned rows with the same lookup using the original bucket option.

The existing KV scan derives one bucket from the current option and applies it to historical files. A matching file can be discarded because its stored bucket was calculated with a different count.

For an append table, write keys across several fixed buckets with overlapping file statistics, then scan with an equality predicate on the complete bucket key. The scan retains files from other buckets even though the predicate determines which bucket can contain the key.

Equality-based pruning should work for both append and KV scans, and changing the current bucket option must not hide matching historical data. The bucket comparison needs each manifest entry's recorded total bucket count. Historical schemas with compatible bucket-key identities, types and hash functions should remain eligible.

### Anything else?

PR #284 implements a shared selector in `FileStoreScan`, preserves explicit-filter semantics, and keeps inferred manifest-cache candidates safe across historical bucket counts and schemas. It includes unit and integration coverage for rescaled options, schema compatibility and repeated cached lookups.

- [x] I'm willing to submit a PR! See #284.

wangyong9999:
### Purpose

Closes #288.

Fixed-bucket append scans retain other buckets even when equality predicates fully specify the bucket key. The existing KV scan has the opposite problem after rescaling: it applies a single bucket computed from the current table option to files written with other bucket counts, which can drop matching data.

Both scans now use a shared selector in `FileStoreScan`. The converter builds the key row and bucket function once; each manifest entry is checked using its own `TotalBuckets()`. Explicit bucket filters keep their existing semantics and take precedence. Historical schemas remain eligible when their ordered bucket-key field IDs, Arrow types and bucket function match. Only incompatible bucket schemas or nonpositive total bucket counts fall back to the existing filtering behavior. Incomplete/non-equality predicates and bucket-unaware tables do not enable inference.

Inferred pruning runs at the manifest-entry level for both scan types. Manifest min/max-bucket skipping still requires an explicit filter. Inferred scans now also use the snapshot live-entry cache: its key includes the selected bucket, current bucket count and current schema ID, and its candidates retain files from other counts or schemas. Query predicates and the selector run after cache lookup. This restores repeated-lookup caching without losing historical files or leaking one query's filtering into another.

Bucket-schema compatibility is memoized per file schema ID. Historical schema reads use the existing `ConcurrentHashMap` utility because manifest filtering runs in parallel.

Decimal literals are converted to the field's precision and scale only when their value is preserved exactly. NaN literals disable inference because equal NaNs can have different stored hashes.

### Tests

Review follow-up `e918bd48`: the build, 56 core tests, 15 cache tests, 81 integration tests, all pre-commit hooks and `git diff --check` passed. The build produced no warnings. Local clang-tidy 11 with the repository configuration checked the three modified C++ files; it reported one `clang-analyzer-cplusplus.NewDelete` warning in the unchanged LRU test's `ASSERT_OK` expansion (`include/paimon/status.h:335`), with no diagnostics on the changed code. This does not establish a clean full CI clang-tidy run. Latest-head CI awaits maintainer approval.

```bash
/opt/tiger/typhoon-blade/gccs/x86_64-x86_64-clang-1101/bin/clang-tidy -p build-reuse --extra-arg=-Wno-error=unknown-warning-option src/paimon/common/io/cache/cache_key.cpp src/paimon/core/operation/file_store_scan.cpp src/paimon/common/io/cache/lru_cache_test.cpp
```

- 56 focused core unit tests and 15 cache unit tests passed, covering per-entry bucket counts, compatible/incompatible historical bucket schemas, concurrent schema reads, and cache-key isolation. The two historical-schema pruning regressions failed on `9aca26db` before this follow-up.
- 81 integration tests passed across Parquet/ORC, prefetch and manifest-cache settings. Append and KV scans read historical files with current bucket options of 2, 4, 8 and 17. Fresh scans of the same snapshot report cache hits and return the correct rows.
- Two append keys share a current two-bucket selection but need different historical buckets; both return correctly through the shared cache. Explicit-bucket entries are seeded first to check isolation from inferred candidates.
- A negative control disabled only inferred cache eligibility, reproducing the old gate: all 8 cache-enabled Append/KV integration cases failed at the cache-enabled assertion. Restored the final source, rebuilt and reran the passing tests.

```bash
cmake --build build-reuse --target paimon-core-test paimon-scan-and-read-inte-test paimon-common-test -j12
build-reuse/release/paimon-core-test --gtest_filter='AppendBucketPruningTest.*:AppendOnlyFileStoreScanTest.*:BucketSelectConverterTest.*:KeyValueBucketPruningTest.*:KeyValueFileStoreScanTest.*:SchemaManagerTest.*:CastExecutorTest.TestDecimalToDecimal*'
build-reuse/release/paimon-common-test --gtest_filter='LruCacheTest.*'
build-reuse/release/paimon-scan-and-read-inte-test --gtest_filter='*Append*:*TestWithPKBucketSelectByPredicate*'
/home/wangyong.alen/.local/bin/uv run --no-project --python 3.11 --with pre-commit==3.8.0 pre-commit run --all-files
git diff --check
```

The incremental build passed with GCC 8, C++17 and default `-Wall`, without warnings. All pre-commit hooks passed. The full test suite, aarch64 and downstream KV RPC E2E were not run. Upstream CI requires maintainer approval; local checks are not a substitute for it.

Earlier measurement at `5ebbaf33`, before the shared-selector follow-up (Release, local disk, warm OS cache, Parquet, prefetch off): 64 buckets, 4 manifests, 256 files, 16,384 rows with 512-byte seeded random values. Both versions used the same persisted table; each configuration warmed up for 40 lookups and measured 40. All lookups returned the expected four rows. The baseline restores only the append scan implementation from `ead92078`; the fixed version was measured both before and after baseline with similar results.

| Mode | Manifest cache | Planned files | Planning p50 ms | Read p50 ms | Scan + read p50 ms | Scan + read p95 ms |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| Before append inference | Off | 256 | 3.972 | 73.118 | 77.087 | 87.784 |
| Append inference at `5ebbaf33` | Off | 4 | 3.904 | 2.019 | 5.936 | 6.589 |
| Before append inference | On | 256 | 3.884 | 74.083 | 78.008 | 79.683 |
| Append inference at `5ebbaf33` | On | 4 | 3.849 | 2.010 | 5.853 | 5.953 |
| Explicit bucket filter | On | 4 | 1.301 | 1.976 | 3.281 | 3.385 |

The improvement is in data-file reads. Planning time is essentially unchanged. This is a small local comparison with four manifests, not a many-manifest or downstream RPC benchmark.

### API and Format

Removed the public `CacheKey::ForSnapshotLiveManifestEntries` factory as requested in review. Repository callers now use the internal `SnapshotLiveManifestEntriesCacheKey::ForExplicit` and `ForInferred` factories. External callers of the removed symbol would need migration; this is a source/ABI change. No protocol or storage format changes. Explicit bucket filtering retains its existing semantics.

### Documentation

Updated `docs/source/api/scan.rst` with shared pruning, per-entry bucket counts, compatible historical schemas and inferred cache behavior.

### Generative AI tooling

Generated-by: Codex (GPT-6)

### DIFF
--- a/include/paimon/cache/cache.h
+++ b/include/paimon/cache/cache.h
@@ -48,6 +48,16 @@ class PAIMON_EXPORT CacheKey {
                                                                     const std::string& branch,
                                                                     int32_t bucket);
 
+    /// Cache candidates for an inferred bucket, including files with other bucket counts or
+    /// schemas.
+    /// @param total_buckets Positive bucket count used to compute the inferred bucket.
+    /// @param schema_id Schema used to build the selector.
+    static std::shared_ptr<CacheKey> ForSnapshotLiveManifestEntries(const std::string& table_path,
+                                                                    const std::string& branch,
+                                                                    int32_t bucket,
+                                                                    int32_t total_buckets,
+                                                                    int64_t schema_id);
+
  public:
     virtual ~CacheKey() = default;
 
--- a/include/paimon/cache/cache.h
+++ b/include/paimon/cache/cache.h
@@ -44,9 +44,6 @@ class PAIMON_EXPORT CacheKey {
                                                  int32_t length, bool is_index);
     static std::shared_ptr<CacheKey> ForKind(const std::string& file_path, int64_t position,
                                              int32_t length, CacheKind kind);
-    static std::shared_ptr<CacheKey> ForSnapshotLiveManifestEntries(const std::string& table_path,
-                                                                    const std::string& branch,
-                                                                    int32_t bucket);
 
  public:
     virtual ~CacheKey() = default;
--- a/include/paimon/cache/cache.h
+++ b/include/paimon/cache/cache.h
@@ -48,16 +48,6 @@ class PAIMON_EXPORT CacheKey {
                                                                     const std::string& branch,
                                                                     int32_t bucket);
 
-    /// Cache candidates for an inferred bucket, including files with other bucket counts or
-    /// schemas.
-    /// @param total_buckets Positive bucket count used to compute the inferred bucket.
-    /// @param schema_id Schema used to build the selector.
-    static std::shared_ptr<CacheKey> ForSnapshotLiveManifestEntries(const std::string& table_path,
-                                                                    const std::string& branch,
-                                                                    int32_t bucket,
-                                                                    int32_t total_buckets,
-                                                                    int64_t schema_id);
-
  public:
     virtual ~CacheKey() = default;
 
--- a/src/paimon/common/io/cache/cache_key.cpp
+++ b/src/paimon/common/io/cache/cache_key.cpp
@@ -24,11 +24,14 @@ namespace {
 class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
  public:
     SnapshotLiveManifestEntriesCacheKey(const std::string& table_path, const std::string& branch,
-                                        int32_t bucket)
+                                        int32_t bucket, int32_t total_buckets = 0,
+                                        int64_t schema_id = 0)
         : CacheKey(CacheKind::SNAPSHOT_LIVE_MANIFEST),
           table_path_(table_path),
           branch_(branch),
-          bucket_(bucket) {}
+          bucket_(bucket),
+          total_buckets_(total_buckets),
+          schema_id_(schema_id) {}
 
     bool IsIndex() const override {
         return false;
@@ -40,7 +43,8 @@ class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
             return false;
         }
         return table_path_ == rhs->table_path_ && branch_ == rhs->branch_ &&
-               bucket_ == rhs->bucket_ && GetKind() == rhs->GetKind();
+               bucket_ == rhs->bucket_ && total_buckets_ == rhs->total_buckets_ &&
+               schema_id_ == rhs->schema_id_ && GetKind() == rhs->GetKind();
     }
 
     size_t HashCode() const override {
@@ -50,6 +54,8 @@ class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
         seed ^= std::hash<int32_t>{}(bucket_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
         seed ^= std::hash<int32_t>{}(static_cast<int32_t>(GetKind())) + HASH_CONSTANT +
                 (seed << 6) + (seed >> 2);
+        seed ^= std::hash<int32_t>{}(total_buckets_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
+        seed ^= std::hash<int64_t>{}(schema_id_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
         return seed;
     }
 
@@ -59,6 +65,8 @@ class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
     const std::string table_path_;
     const std::string branch_;
     const int32_t bucket_;
+    const int32_t total_buckets_;
+    const int64_t schema_id_;
 };
 
 }  // namespace
@@ -80,6 +88,15 @@ std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::st
                                                                    const std::string& branch,
                                                                    int32_t bucket) {
     return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket);
+}
+
+std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
+                                                                   const std::string& branch,
+                                                                   int32_t bucket,
+                                                                   int32_t total_buckets,
+                                                                   int64_t schema_id) {
+    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
+                                                                 total_buckets, schema_id);
 }
 
 bool PositionCacheKey::IsIndex() const {
--- a/src/paimon/common/io/cache/cache_key.cpp
+++ b/src/paimon/common/io/cache/cache_key.cpp
@@ -33,12 +33,6 @@ std::shared_ptr<CacheKey> CacheKey::ForKind(const std::string& file_path, int64_
     return key;
 }
 
-std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
-                                                                   const std::string& branch,
-                                                                   int32_t bucket) {
-    return SnapshotLiveManifestEntriesCacheKey::ForExplicit(table_path, branch, bucket);
-}
-
 std::shared_ptr<CacheKey> SnapshotLiveManifestEntriesCacheKey::ForExplicit(
     const std::string& table_path, const std::string& branch, int32_t bucket) {
     return std::shared_ptr<CacheKey>(new SnapshotLiveManifestEntriesCacheKey(
--- a/src/paimon/common/io/cache/cache_key.cpp
+++ b/src/paimon/common/io/cache/cache_key.cpp
@@ -24,8 +24,7 @@ namespace {
 class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
  public:
     SnapshotLiveManifestEntriesCacheKey(const std::string& table_path, const std::string& branch,
-                                        int32_t bucket, int32_t total_buckets = 0,
-                                        int64_t schema_id = 0)
+                                        int32_t bucket, int32_t total_buckets, int64_t schema_id)
         : CacheKey(CacheKind::SNAPSHOT_LIVE_MANIFEST),
           table_path_(table_path),
           branch_(branch),
@@ -87,14 +86,14 @@ std::shared_ptr<CacheKey> CacheKey::ForKind(const std::string& file_path, int64_
 std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
                                                                    const std::string& branch,
                                                                    int32_t bucket) {
-    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket);
+    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
+                                                                 /*total_buckets=*/0,
+                                                                 /*schema_id=*/0);
 }
 
-std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
-                                                                   const std::string& branch,
-                                                                   int32_t bucket,
-                                                                   int32_t total_buckets,
-                                                                   int64_t schema_id) {
+std::shared_ptr<CacheKey> CreateInferredSnapshotLiveManifestEntriesCacheKey(
+    const std::string& table_path, const std::string& branch, int32_t bucket, int32_t total_buckets,
+    int64_t schema_id) {
     return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
                                                                  total_buckets, schema_id);
 }
--- a/src/paimon/common/io/cache/cache_key.cpp
+++ b/src/paimon/common/io/cache/cache_key.cpp
@@ -19,56 +19,6 @@
 #include "paimon/common/io/cache/cache_key.h"
 
 namespace paimon {
-namespace {
-
-class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
- public:
-    SnapshotLiveManifestEntriesCacheKey(const std::string& table_path, const std::string& branch,
-                                        int32_t bucket, int32_t total_buckets, int64_t schema_id)
-        : CacheKey(CacheKind::SNAPSHOT_LIVE_MANIFEST),
-          table_path_(table_path),
-          branch_(branch),
-          bucket_(bucket),
-          total_buckets_(total_buckets),
-          schema_id_(schema_id) {}
-
-    bool IsIndex() const override {
-        return false;
-    }
-
-    bool Equals(const CacheKey& other) const override {
-        const auto* rhs = dynamic_cast<const SnapshotLiveManifestEntriesCacheKey*>(&other);
-        if (!rhs) {
-            return false;
-        }
-        return table_path_ == rhs->table_path_ && branch_ == rhs->branch_ &&
-               bucket_ == rhs->bucket_ && total_buckets_ == rhs->total_buckets_ &&
-               schema_id_ == rhs->schema_id_ && GetKind() == rhs->GetKind();
-    }
-
-    size_t HashCode() const override {
-        size_t seed = 0;
-        seed ^= std::hash<std::string>{}(table_path_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
-        seed ^= std::hash<std::string>{}(branch_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
-        seed ^= std::hash<int32_t>{}(bucket_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
-        seed ^= std::hash<int32_t>{}(static_cast<int32_t>(GetKind())) + HASH_CONSTANT +
-                (seed << 6) + (seed >> 2);
-        seed ^= std::hash<int32_t>{}(total_buckets_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
-        seed ^= std::hash<int64_t>{}(schema_id_) + HASH_CONSTANT + (seed << 6) + (seed >> 2);
-        return seed;
-    }
-
- private:
-    static constexpr uint64_t HASH_CONSTANT = 0x9e3779b97f4a7c15ULL;
-
-    const std::string table_path_;
-    const std::string branch_;
-    const int32_t bucket_;
-    const int32_t total_buckets_;
-    const int64_t schema_id_;
-};
-
-}  // namespace
 
 std::shared_ptr<CacheKey> CacheKey::ForPosition(const std::string& file_path, int64_t position,
                                                 int32_t length, bool is_index) {
@@ -86,16 +36,20 @@ std::shared_ptr<CacheKey> CacheKey::ForKind(const std::string& file_path, int64_
 std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
                                                                    const std::string& branch,
                                                                    int32_t bucket) {
-    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
-                                                                 /*total_buckets=*/0,
-                                                                 /*schema_id=*/0);
+    return SnapshotLiveManifestEntriesCacheKey::ForExplicit(table_path, branch, bucket);
 }
 
-std::shared_ptr<CacheKey> CreateInferredSnapshotLiveManifestEntriesCacheKey(
+std::shared_ptr<CacheKey> SnapshotLiveManifestEntriesCacheKey::ForExplicit(
+    const std::string& table_path, const std::string& branch, int32_t bucket) {
+    return std::shared_ptr<CacheKey>(new SnapshotLiveManifestEntriesCacheKey(
+        table_path, branch, bucket, Mode::kExplicit, std::nullopt, std::nullopt));
+}
+
+std::shared_ptr<CacheKey> SnapshotLiveManifestEntriesCacheKey::ForInferred(
     const std::string& table_path, const std::string& branch, int32_t bucket, int32_t total_buckets,
     int64_t schema_id) {
-    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
-                                                                 total_buckets, schema_id);
+    return std::shared_ptr<CacheKey>(new SnapshotLiveManifestEntriesCacheKey(
+        table_path, branch, bucket, Mode::kInferred, total_buckets, schema_id));
 }
 
 bool PositionCacheKey::IsIndex() const {
--- a/src/paimon/common/io/cache/cache_key.h
+++ b/src/paimon/common/io/cache/cache_key.h
@@ -26,6 +26,11 @@
 
 namespace paimon {
 
+// Cache inferred candidates separately from explicit buckets and other bucket layouts.
+std::shared_ptr<CacheKey> CreateInferredSnapshotLiveManifestEntriesCacheKey(
+    const std::string& table_path, const std::string& branch, int32_t bucket, int32_t total_buckets,
+    int64_t schema_id);
+
 class PositionCacheKey : public CacheKey {
  public:
     PositionCacheKey(const std::string& file_path, int64_t position, int32_t length, bool is_index,
--- a/src/paimon/common/io/cache/cache_key.h
+++ b/src/paimon/common/io/cache/cache_key.h
@@ -20,16 +20,76 @@
 
 #include <cstdint>
 #include <memory>
+#include <optional>
 #include <string>
 
 #include "paimon/cache/cache.h"
 
 namespace paimon {
 
-// Cache inferred candidates separately from explicit buckets and other bucket layouts.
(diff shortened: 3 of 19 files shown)


src/paimon/core/operation/append_only_file_store_scan.cpp
```diff
@@ -86,6 +104,12 @@ Result<bool> AppendOnlyFileStoreScan::FilterByStats(const ManifestEntry& entry)
     if (!predicates_) {
         return true;
     }
+    // A historical file may use a different schema or bucket count after a rescale.
+    // Keep the inferred bucket separate from the caller's explicit bucket filter.
+    if (predicate_bucket_ && entry.TotalBuckets() == core_options_.GetBucket() &&
```
wangyong9999: Both guards give up more than they need to. A different `TotalBuckets()` does not make the file unprunable: its bucket is just `function.Bucket(key_row, entry.TotalBuckets())`. `Convert()` already builds the key row and the bucket function, so keep those two (or memoize per total-bucket count) and recompute for that count instead of retaining every file of a rescaled partition. Likewise the `schema_id` check disables pruning for every file older than the last schema change, and adding a nullable column is enough to trigger that, while the hash only depends on the bucket-key columns. `FilterByStats` already loads the file's data schema a few lines below; compare the bucket-key field types there and skip only when they differ.

src/paimon/core/operation/append_only_file_store_scan.cpp
```diff
@@ -66,6 +69,21 @@ Result<std::unique_ptr<AppendOnlyFileStoreScan>> AppendOnlyFileStoreScan::Create
                                     table_schema, arrow_schema, core_options, executor, pool));
     PAIMON_RETURN_NOT_OK(
         scan->SplitAndSetFilter(table_schema->PartitionKeys(), arrow_schema, scan_filters));
+    const auto& bucket_keys = table_schema->BucketKeys();
```
wangyong9999: Worth spelling out in the description: since the inferred bucket never reaches `bucket_filter_`, planning still opens every manifest. Both the min/max-bucket manifest skip in `FilterManifest` and the snapshot live-manifest-entry cache are gated on `bucket_filter_`, and for the point-lookup case that motivates this change they are most of the planning cost. The entry-level check is the safe choice for mixed layouts, but a number for planning time with inferred vs explicit bucket on a many-manifest table would show whether this delivers the win by itself.

src/paimon/core/operation/append_only_file_store_scan.cpp
```diff
@@ -66,6 +69,21 @@ Result<std::unique_ptr<AppendOnlyFileStoreScan>> AppendOnlyFileStoreScan::Create
...
+        std::vector<std::shared_ptr<arrow::DataType>> bucket_key_types;
+        bucket_key_types.reserve(bucket_keys.size());
+        for (const auto& key : bucket_keys) {
+            PAIMON_ASSIGN_OR_RAISE(DataField field, table_schema->GetField(key));
+            bucket_key_types.push_back(field.Type());
+        }
+        PAIMON_ASSIGN_OR_RAISE(scan->predicate_bucket_,
+                               BucketSelectConverter::Convert(
+                                   scan->predicates_, bucket_keys, bucket_key_types,
+                                   core_options.GetBucketFunctionType(), num_buckets, pool.get()));
```
wangyong9999: Please rescale DECIMAL literals to the bucket field's scale before inferring the bucket, or skip pruning if exact conversion is not possible. With `DECIMAL(10,2)` and 17 buckets, stored `1.20` hashes to bucket 9, but `Decimal::FromUnscaledLong(12, 10, 1)` (`1.2`) infers bucket 0. This predicate passes validation and matches the file stats, yet the matching file is discarded. A cross-scale regression reproduces this on the PR and passes with the previous scan implementation.

src/paimon/core/operation/append_only_file_store_scan.cpp
```diff
@@ -86,6 +104,12 @@ Result<bool> AppendOnlyFileStoreScan::FilterByStats(const ManifestEntry& entry)
     if (!predicates_) {
         return true;
     }
+    // A historical file may use a different schema or bucket count after a rescale.
+    // Keep the inferred bucket separate from the caller's explicit bucket filter.
+    if (predicate_bucket_ && entry.TotalBuckets() == core_options_.GetBucket() &&
+        entry.File()->schema_id == table_schema_->Id() && entry.Bucket() != *predicate_bucket_) {
+        return false;
```
wangyong9999: Please skip inferred pruning for NaN bucket-key literals. `Literal::CompareTo()` treats all NaNs as equal, but the default bucket hash uses raw bits. With 17 buckets, DOUBLE payloads `0x7ff8000000000000` and `0x7ff8000000000001` map to buckets 11 and 14. Querying the former now drops a matching-schema file containing the latter, even though its stats match. I reproduced this against both versions: the previous scan retains the file. Canonicalizing only the query cannot fix existing stored keys; retain all buckets for NaN and add a regression.

src/paimon/core/operation/append_only_file_store_scan.cpp
wangyong9999: Keeping these fallbacks in this PR. They are conservative but do not lose rows. Pruning across layouts would need to use each file's bucket count and compatible bucket-key schema. That is a separate optimization; the description now explicitly says that even an unrelated schema change disables inference for older files.

src/paimon/core/operation/append_only_file_store_scan.cpp
wangyong9999: Updated the description and scan docs. This prunes data files at the manifest-entry level; it still opens manifests and does not activate min/max-bucket manifest skipping or the bucket-specific live-entry cache. I have not run a many-manifest planning benchmark and am not claiming a planning-time speedup.

src/paimon/core/operation/append_only_file_store_scan.cpp
wangyong9999: Fixed in 5ebbaf33. The shared converter now uses DecimalToDecimalCastExecutor and checks that the converted value equals the original before hashing. Rounding and overflow fall back to no inferred bucket. Added DEFAULT/HIVE tests for both compact and noncompact decimals, plus the append regression for 1.2 against stored 1.20. The regression failed before the fix and passes now.

src/paimon/core/operation/append_only_file_store_scan.cpp
wangyong9999: Fixed in 5ebbaf33. FLOAT and DOUBLE NaN bucket-key literals now disable inference in the shared converter. Added a regression with the two DOUBLE payloads above: their hashes differ, their stats match, and the file is retained. The test failed before the fix and passes now.

lxy-9602:
C++ already performs predicate-to-bucket conversion for `KeyValueFileStoreScan`, while this PR introduces a second append-specific path with different semantics. Java installs a shared total-aware selector for both scan types and computes the bucket using each manifest entry’s totalBucket. Could we centralize this in `FileStoreScan` and make the selector total-aware like java? Besides avoiding duplication, this would prevent the existing KV path from pruning historical files incorrectly after bucket rescaling. Thanks!

wangyong9999:
@lxy-9602 Fixed in 9aca26db. I reproduced the KV bug: with the current option set to 4 buckets, a matching file written with 2 buckets was discarded.

Selector setup and entry filtering now live in `FileStoreScan`. Both scans reuse the key row and bucket function, then compute the bucket from each entry's `TotalBuckets()`. The append-specific path and KV's inferred global bucket filter are removed.

Explicit bucket filters still take precedence. Different schema IDs retain the conservative fallback. Inferred KV scans now also skip the manifest-level bucket optimization and bucket-specific cache, since a single inferred bucket is unsafe across rescaling. The description calls out this tradeoff.

Added regressions for both scans and extended the real-file scan/read tests to current bucket options of 2, 4, 8 and 17. The two unit regressions failed before the fix; 46 unit tests and 81 integration tests pass now. The build and pre-commit checks pass too.

wangyong9999:
Fixed the clang-tidy failure and the latest cache-key review comments in `a824461a`. The new workflows are waiting for approval again (`action_required`). Could a maintainer approve them? [Build and Test run](https://github.com/apache/paimon-cpp/actions/runs/34185536124).

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -557,6 +557,13 @@ Result<bool> FileStoreScan::FilterManifestEntry(const ManifestEntry& entry) cons
     if (bucket_filter_ != std::nullopt && entry.Bucket() != bucket_filter_.value()) {
         return false;
     }
+    // Hash with the file's bucket count, never the current table's count. An older schema
+    // may encode bucket keys differently, so leave those files to the existing stats filter.
+    if (bucket_selector_ && entry.TotalBuckets() > 0 &&
+        entry.File()->schema_id == table_schema_->Id() &&
```
wangyong9999: This guard is new for the key-value path. Before this PR the KV scan pruned by the inferred bucket regardless of `schema_id`, so a table that ever added a column now loses primary-key bucket pruning for every file written before that change, and a time-travel scan against an older snapshot loses it entirely. The hash only depends on the bucket-key columns. `FilterByStats` in both scans already reads the file's schema when the id differs; compare the bucket-key field types there (memoized per schema id) and skip only when they differ. Java's `BucketSelector` applies to every entry with no schema check at all, so that would still be the stricter of the two.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -602,6 +609,19 @@ Status FileStoreScan::SplitAndSetFilter(const std::vector<std::string>& partitio
         }
     }
     bucket_filter_ = scan_filters->GetBucketFilter();
+    const auto& bucket_keys = table_schema_->BucketKeys();
+    if (predicates_ && !bucket_filter_ && core_options_.GetBucket() > 0 && !bucket_keys.empty()) {
```
wangyong9999, line 653: Removing the global filter also removes the two planning optimizations the KV path had before this PR: the min/max-bucket skip in `FilterManifestFileMeta` and the snapshot live-manifest-entry cache are both gated on `bucket_filter_`, so an inferred point lookup now reads every manifest on every plan even with `scan.manifest-entry-cache.max-snapshots` set. The cache part does not need the rescale-unsafe global filter: key the cache by the inferred bucket under the current count together with that count, have `ReadAndMergeBucketFileEntries` keep entries whose `TotalBuckets()` differs from the current count in addition to `Bucket() == bucket`, and let the selector filter those at entry level as it already does. Manifest skipping can stay explicit-only. Without this, tables that never rescaled pay a full manifest read on every lookup, which is most of the planning cost this pruning is meant to remove.

src/paimon/core/operation/file_store_scan.cpp
wangyong9999: Fixed in 7c9545a4. Bucket compatibility now compares the ordered field IDs, Arrow types and bucket function, and caches the result per schema ID. An unrelated column change no longer disables pruning; incompatible bucket-key layouts still fall back. Added Append/KV regressions for compatible historical schemas and an incompatible STRING/BINARY case. Historical schema reads now use the existing concurrent map, since manifest filtering can load them in parallel.

src/paimon/core/operation/file_store_scan.cpp
wangyong9999: Fixed in 7c9545a4. Inferred scans can use the live-entry cache again. The key includes the current schema ID as well as bucket and bucket count: files needing a schema fallback must not disappear when another query fills the cache. Candidates retain matching buckets, other bucket counts and other schema IDs; the selector and predicates run after lookup. Cold inferred reads keep all manifests, while explicit filtering keeps its existing path.

The tests verify cache hits across fresh scans, explicit/inferred isolation, and two keys that share the current bucket but need different historical buckets. Disabling inferred cache eligibility makes all 8 cache-enabled integration cases fail. With the fix, 56 core tests, 15 cache tests and 81 integration tests pass. Build and pre-commit checks pass; upstream CI still needs maintainer approval.

include/paimon/cache/cache.h
```diff
@@ -48,6 +48,16 @@ class PAIMON_EXPORT CacheKey {
...
+    /// Cache candidates for an inferred bucket, including files with other bucket counts or
+    /// schemas.
+    /// @param total_buckets Positive bucket count used to compute the inferred bucket.
+    /// @param schema_id Schema used to build the selector.
+    static std::shared_ptr<CacheKey> ForSnapshotLiveManifestEntries(const std::string& table_path,
+                                                                    const std::string& branch,
+                                                                    int32_t bucket,
+                                                                    int32_t total_buckets,
+                                                                    int64_t schema_id);
+
```
lxy-9602, line 47: Putting these two `ForSnapshotLiveManifestEntries` functions in `include` exposes too much internal meaning, and from the external API it is not clear when the 3-parameter version should be used versus the 5-parameter version.

src/paimon/common/io/cache/cache_key.cpp
```diff
@@ -24,11 +24,14 @@ namespace {
 class SnapshotLiveManifestEntriesCacheKey : public CacheKey {
  public:
     SnapshotLiveManifestEntriesCacheKey(const std::string& table_path, const std::string& branch,
-                                        int32_t bucket)
+                                        int32_t bucket, int32_t total_buckets = 0,
+                                        int64_t schema_id = 0)
```
lxy-9602: Please avoid using default functions in production code.

src/paimon/core/operation/key_value_file_store_scan_test.cpp
```diff
@@ -58,6 +60,51 @@ class Schema;
...
+            TableSchema::Create(current_schema_id, arrow_schema, {}, {"rowkey"}, options));
+        ASSERT_OK_AND_ASSIGN(auto core_options, CoreOptions::FromMap(options));
+        auto predicate = PredicateBuilder::Equal(0, "rowkey", FieldType::STRING,
+                                                 Literal(FieldType::STRING, "key", 3));
+        auto filters = std::make_shared<ScanFilter>(
+            predicate, std::vector<std::map<std::string, std::string>>(), std::nullopt);
+        ASSERT_OK_AND_ASSIGN(
+            auto scan,
+            KeyValueFileStoreScan::Create(nullptr, manager, nullptr, nullptr, schema, arrow_schema,
+                                          filters, core_options, GetGlobalDefaultExecutor(), pool));
```
lxy-9602: Please do not use the global executor.

src/paimon/core/operation/append_only_file_store_scan_test.cpp
```diff
@@ -46,10 +51,172 @@
...
+        }
+        auto arrow_schema = DataField::ConvertDataFieldsToArrowSchema(fields);
+        PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<TableSchema> schema,
+                               TableSchema::Create(schema_id_, arrow_schema, {}, {}, options_));
+        PAIMON_ASSIGN_OR_RAISE(CoreOptions options, CoreOptions::FromMap(options_));
+        auto filters = std::make_shared<ScanFilter>(
+            predicate, std::vector<std::map<std::string, std::string>>(), bucket);
+        return AppendOnlyFileStoreScan::Create(nullptr, schema_manager_, nullptr, nullptr, schema,
+                                               arrow_schema, filters, options,
+                                               GetGlobalDefaultExecutor(), pool_);
```
lxy-9602: Please do not use the global executor.

src/paimon/core/operation/append_only_file_store_scan_test.cpp
```diff
@@ -46,10 +51,172 @@
...
+
+TEST_F(AppendBucketPruningTest, PreservesDifferentNaNPayloadMatch) {
+    rowkey_type_ = arrow::float64();
+    uint64_t query_bits = 0x7ff8000000000000ULL;
+    uint64_t stored_bits = 0x7ff8000000000001ULL;
+    double query_value;
+    double stored_value;
+    std::memcpy(&query_value, &query_bits, sizeof(query_value));
+    std::memcpy(&stored_value, &stored_bits, sizeof(stored_value));
+    CheckMatchingValue(FieldType::DOUBLE, query_value, stored_value);
```
lxy-9602, line 213: Consider use `CanonicalizeFloatingPoint` in math rather than `memcpy`.

include/paimon/cache/cache.h
wangyong9999: Moved the new five-argument factory to the internal cache header in e6bb0bfb and named it CreateInferredSnapshotLiveManifestEntriesCacheKey. Kept the existing three-argument public factory for compatibility. The PR no longer adds a public overload; cache isolation and Append/KV integration tests pass.

src/paimon/common/io/cache/cache_key.cpp
wangyong9999: Removed the constructor default arguments in e6bb0bfb. The explicit-bucket factory now passes both values explicitly.

src/paimon/core/operation/key_value_file_store_scan_test.cpp
wangyong9999: Replaced the global executor with CreateDefaultExecutor() in e6bb0bfb. The focused KV scan tests pass.

src/paimon/core/operation/append_only_file_store_scan_test.cpp
wangyong9999: Replaced the global executor with CreateDefaultExecutor() in e6bb0bfb. The focused Append scan tests pass.

src/paimon/core/operation/append_only_file_store_scan_test.cpp
wangyong9999: Replaced memcpy with FloatingPointFromBits from math.h in e6bb0bfb. CanonicalizeFloatingPoint would collapse the two NaN payloads and remove the regression this test covers. The test still checks that a matching file with a different stored NaN payload is retained, and passes.

src/paimon/common/io/cache/cache_key.h
```diff
@@ -26,6 +26,11 @@
 
 namespace paimon {
 
+// Cache inferred candidates separately from explicit buckets and other bucket layouts.
+std::shared_ptr<CacheKey> CreateInferredSnapshotLiveManifestEntriesCacheKey(
+    const std::string& table_path, const std::string& branch, int32_t bucket, int32_t total_buckets,
+    int64_t schema_id);
+
```
lxy-9602, line 93: Please avoid introducing global static functions in Paimon.

src/paimon/common/io/cache/cache_key.cpp
```diff
@@ -87,14 +86,14 @@ std::shared_ptr<CacheKey> CacheKey::ForKind(const std::string& file_path, int64_
 std::shared_ptr<CacheKey> CacheKey::ForSnapshotLiveManifestEntries(const std::string& table_path,
                                                                    const std::string& branch,
                                                                    int32_t bucket) {
-    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket);
+    return std::make_shared<SnapshotLiveManifestEntriesCacheKey>(table_path, branch, bucket,
+                                                                 /*total_buckets=*/0,
+                                                                 /*schema_id=*/0);
 }
```
lxy-9602, line 40: Could we explicitly distinguish explicit and inferred cache keys instead of using `0` as a placeholder? `schema_id = 0` is valid and looks like real metadata.

For example, the internal key could provide `ForExplicit` and `ForInferred` static factories, store an explicit mode, and use `std::optional` for inferred-only fields. This avoids both placeholder metadata values and the namespace-level factory function.

```cpp

class SnapshotLiveManifestEntriesCacheKey {
 public:
    static std::shared_ptr<CacheKey> ForExplicit(
        const std::string& path, const std::string& branch, int32_t bucket);

    static std::shared_ptr<CacheKey> ForInferred(
        const std::string& path, const std::string& branch, int32_t bucket,
        int32_t total_buckets, int64_t schema_id);
};
```

src/paimon/common/io/cache/cache_key.h
wangyong9999: Replaced the namespace-level factory with SnapshotLiveManifestEntriesCacheKey::ForExplicit and ForInferred in a824461a. Both stay in the internal cache header.

src/paimon/common/io/cache/cache_key.cpp
wangyong9999: Done in a824461a. The key now stores an explicit mode and optional bucket-count/schema fields. ForExplicit leaves the fields empty; ForInferred keeps the supplied metadata, including schema 0. Equality and hashing include the mode. Updated the isolation test; 15 cache tests and 81 Append/KV integration tests pass.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -376,7 +376,7 @@ Status FileStoreScan::ReadManifestEntriesWithCache(
 
 std::shared_ptr<CacheKey> FileStoreScan::SnapshotLiveManifestEntriesCacheKey(int32_t bucket) const {
     if (!bucket_filter_ && bucket_selector_) {
-        return CreateInferredSnapshotLiveManifestEntriesCacheKey(
+        return paimon::SnapshotLiveManifestEntriesCacheKey::ForInferred(
             table_path_, BranchManager::NormalizeBranch(core_options_.GetBranch()), bucket,
```
lxy-9602, line 381: In `src`, we probably don’t need the `paimon::` qualifier here? Also, can `CacheKey::ForSnapshotLiveManifestEntries` be removed directly?

wangyong9999: Done in e918bd48. Removed the public forwarding factory and switched its callers to the internal factories. Renamed the scan helper to CreateSnapshotLiveManifestEntriesCacheKey so it no longer shadows the key class, then removed the paimon:: qualifier. The PR description notes the public API removal. Build, 56 core tests, 15 cache tests and 81 integration tests pass.

wangyong9999:
Following up on the suggestion to centralize a total-aware selector in `FileStoreScan`: #302 does that for the key-value scan with a smaller footprint (no changes under `include/`, the entry cache stays keyed per bucket, candidate buckets come from the bucket numbers in the schema history, and every entry is checked against its own `total_buckets`). Once #302 lands I will rebase this PR onto it so that `AppendOnlyFileStoreScan` only installs a selector the same way and the remaining diff is the append coverage.

wangyong9999:
One small follow-up on the inferred cache rebuild in `ReadAndMergeBucketFileEntries`. Entries are retained when `entry.File()->schema_id != table_schema_->Id()`, so after a schema change that keeps the bucket keys (an added column, say) every bucket's cached set carries all pre-change entries of the whole table until compaction rewrites them, and each lookup filters them again. `HasCompatibleBucketKeys()` already tells whether an older schema hashes the key the same way, and a compatible entry with the current bucket count sits in exactly `Bucket(count)`, so it only needs retaining when `entry.Bucket() == bucket`:

```cpp
bool retain = entry.Bucket() == bucket;
if (!retain && inferred_bucket) {
    if (entry.TotalBuckets() != core_options_.GetBucket()) {
        retain = true;
    } else {
        PAIMON_ASSIGN_OR_RAISE(bool compatible,
                               HasCompatibleBucketKeys(entry.File()->schema_id));
        retain = !compatible;
    }
}
```

Measured on a two-bucket primary key table with partition `p1` written before an ADD COLUMN and `p2` after it: each bucket's cached set goes from 3 entries (its own file plus both `p1` files) to 2 (its own two files), all 80 lookups still return their row, and the scan suites pass. Entries written with another bucket count are unchanged, `TotalBuckets() != count` keeps them. Happy to send it as a follow-up if you would rather keep this PR as it is.

lxy-9602 (approved):
+1

PR #285, opened 2026-09-05
perf(read): reduce point-lookup metadata I/O

### LINKED ISSUE #290: [Feature] Reduce redundant metadata I/O in Parquet point lookups
### Search before asking

- [x] I searched existing issues. #225 concerns page row-count limits and #214 concerns index pushdown; neither covers this metadata-I/O reuse improvement.

### Motivation

Repeated full-primary-key lookups against Parquet files on remote storage still perform avoidable metadata I/O:

- Readers reuse cached footers but reload ColumnIndex/OffsetIndex bytes across reader instances.
- Arrow's default ColumnIndex read envelope can include large payload-column min/max values even when only primary-key columns participate in the predicate.
- Latest-snapshot discovery probes the snapshot directory before reading `LATEST`, although the successful hint path does not require this probe and the listing fallback already checks directory existence.

These are performance opportunities, not a claim that the current lookup returns incorrect rows.

### Solution

Reuse immutable page-index bytes through the existing caller-provided cache, retain the allocation's memory pool for the cached bytes' lifetime, and limit predicate index reads to predicate columns. Keep query-local readers and clear restricted read hints before loading projected payload offset indexes. Remove the redundant snapshot-directory probe while preserving the next-snapshot check and listing fallback.

No process-wide resource manager, snapshot TTL, result cache, public API change or file-format change is proposed. Implementation: #285.

### Anything else?

A synthetic remote-HDFS point-lookup experiment measured the incremental snapshot-probe change on an unchanged small-page table. At eight concurrent KV clients, mean latency decreased from 51.24 to 47.79 ms (6.7%), P99 from 90.18 to 81.31 ms, and throughput increased from 155.93 to 167.22 queries/s (7.2%). Both revisions already included page-index reuse and predicate-only index reads; these numbers are not the total PR speedup. The PR records the configuration and measurement limits.

A focused regression fixture also reduces predicate-index bytes read from 6,651 to 2,212 while preserving the projected payload result. This is an index-read byte count, not total query I/O.

- [x] I'm willing to submit a PR! See #285.

wangyong9999:
### Purpose

Closes #290.

Parquet point readers reuse footers but reload page indexes. Arrow also coalesces the ColumnIndex ranges of all columns by default, including large payload min/max values unrelated to the predicate.

**Reuse immutable page-index bytes through the caller-provided cache, and restrict predicate index reads to predicate columns.** URI/offset/length identify entries in the existing `DATA_FILE_FOOTER` budget. A query-local Arrow read hint is cleared before projected payload OffsetIndex reads. No shared readers, process-wide manager, new capacity setting, result cache or snapshot freshness change.

**Cached footer/index allocations retain their memory pool until eviction.** Missing cache/URI uses storage; errors propagate. No separate design document.

**Latest-snapshot discovery no longer probes the parent directory before reading LATEST.** The next-snapshot check is retained, and the existing listing fallback already checks directory existence. This removes one redundant filesystem call without caching the hint or changing snapshot freshness.

#### Performance (anonymized integration experiment)

The comparison below isolates the **last incremental change: removing the redundant snapshot-directory probe**. Both revisions already include page-index reuse and predicate-only index reads. It is not a measurement of the entire PR against unmodified main.

| Concurrent clients | Mean latency, before → after (ms) | P99, before → after (ms) | Throughput, before → after (queries/s) |
|---|---:|---:|---:|
| 1 | 46.05 → 39.69 | 79.52 → 71.36 | 21.70 → 25.18 |
| 8 | 51.24 → 47.79 | 90.18 → 81.31 | 155.93 → 167.22 |
| 16 | 55.70 → 51.20 | 102.26 → 96.90 | 286.90 → 312.14 |

At C8, mean latency decreased **6.7% (3.45 ms)** and throughput increased **7.2%**. The two post-change rounds averaged 47.71/47.87 ms, with 10,047 measured requests in total.

Setup: a native KV service embedding Paimon-CPP, with no SQL planning; remote SSD-backed HDFS; a 96-vCPU, approximately 1-TB-RAM serving host. The unchanged synthetic PK table contains 2.1 million rows / 3.11 GB across two partitions, one bucket per partition and 22 L2 files, with no L0 or concurrent writes. Queries specify the complete three-field primary key and return five fields, including 512/768-byte string payloads. Parquet is uncompressed; target page size is 16 KiB, write batch 16 rows and dictionary target 16 KiB; representative payload pages are approximately 16/24 KiB. There is no global secondary index in this experiment.

Each configuration uses the same 512-key working set, 512 warm-up requests per round and two 30-second measurement rounds. Mean is request-weighted; P99 is calculated from pooled raw samples. Every result is validated after the RPC timer, while validation time remains included in throughput. Profiling runs are separate; no slow requests are discarded. C16 mean remains above 50 ms, and its post-change maximum is 1.14 s versus 0.52 s before. These short, warm-cache runs do not establish statistical significance, a tail-latency SLA or random full-table performance. Data-layout rewrite results are excluded from the PR's code-speedup comparison.

Separately, the predicate-index regression fixture reads **2,212 B instead of 6,651 B (66.7% fewer index bytes)** and returns the full projected payload. This is not total query I/O. In the warm integration workload, predicate-only index reads changed C8 mean from 52.34 to 51.24 ms; that small difference alone is insufficient to claim a significant steady-state improvement.

#### Regression coverage

- `TestPageIndexBytesSurviveReaderClose`: multi-RG index reuse, eviction/reload, data reads bypass cache, temporary allocator lifetime.
- `TestPointReadReusesFooterAndPageIndexes`: independent point readers return the same row without reloading metadata.
- `TestCachedFooterKeepsAllocatorAliveUntilEviction`: footer allocator lifetime.
- `PredicateReadsOnlyItsPageIndexesAndKeepsPayloadReadable`: exact predicate-index byte count and complete projected row results.
- `LatestSnapshotWithStaleMissingOrInvalidHint`: valid and stale hints, newly committed snapshots, missing and malformed hints. All 19 SnapshotManager tests passed locally, including missing-table cases.

Local current `FileReaderWrapper` and its 14 tests compiled and passed against the matching Arrow 17/SCM dependency set. C++17 syntax and clang-format 20.1.8 checks passed. All GitHub checks passed for 526576e0, including x86/aarch64 GCC/Clang and sanitizers. The review follow-up removes filesystem-call-count assertions and retains result checks for valid, stale, missing and malformed hints; all 19 SnapshotManager tests and repository-wide pre-commit hooks passed locally. CI for the follow-up commit is pending.

### API and Format

No public API or storage-format change. Data-page reads and snapshot freshness are unchanged. Full-projection OffsetIndex readers remain separate from the predicate-restricted reader. Missing-hint retries and listing error propagation are retained.

### Documentation

Inline comments explain range selection and lifetime. No new user option. The integration measurements above are separate from unit-test evidence.

### Generative AI tooling

Generated-by: OpenAI Codex


lucasfang:
LGTM

src/paimon/core/utils/snapshot_manager_test.cpp
```diff
@@ -139,6 +141,51 @@ TEST(SnapshotManagerTest, TestPathNotExist) {
...
+    ASSERT_OK(fs->Delete(hint_path));
+    fs->exists_paths.clear();
+    ASSERT_OK_AND_ASSIGN(latest, mgr.LatestSnapshotId());
+    ASSERT_EQ(latest, 2);
+    ASSERT_EQ(std::count(fs->exists_paths.begin(), fs->exists_paths.end(), mgr.SnapshotDirectory()),
+              1);
+
+    ASSERT_OK(fs->WriteFile(hint_path, "invalid", true));
+    ASSERT_OK_AND_ASSIGN(latest, mgr.LatestSnapshotId());
+    ASSERT_EQ(latest, 2);
```
lxy-9602, line 167: I’m not sure this test is necessary. It is tightly coupled to the current implementation flow, so it may easily become a false alarm if `FindLatest` is refactored later.

wangyong9999: Agreed—the filesystem call counts tie this test to the current implementation. Removed the tracking filesystem and those assertions in 3ce3aba9. Kept only the returned snapshot checks for valid, stale, missing and malformed hints. All 19 SnapshotManager tests pass locally. No production code changed.

lxy-9602 (approved):
+1

PR #286, opened 2026-09-06
perf(read): support warmup between data files

### LINKED ISSUE #289: [Feature] Overlap the first read of merge-on-read files with a reader warmup
## Search before asking

## Motivation

Reading a primary-key table in merge-on-read mode serializes the first read of every data file behind the consumption of the previous one, so on remote storage the merge pays one network round trip per file instead of overlapping them.

A split is partitioned into sections of overlapping files, and a section is merged by a `LoserTree` (the default sort engine) whose leaves are the section's sorted runs. Each run is a `ConcatKeyValueRecordReader` over its files, and each file is a `KeyValueDataFileRecordReader` over a stack of `FileBatchReader` wrappers that ends in `DelegatingPrefetchReader` and `PrefetchFileBatchReaderImpl`. `PrefetchFileBatchReaderImpl` issues its prefetch reads from a background thread that it starts lazily, inside the first `NextBatchWithBitmap()` call, so nothing of a file is in flight before the consumer actually asks that file for its first batch. That costs at two levels:

- Within a run the files are read strictly one after another, and a filtered read of a keyed table yields about one batch per file. The consumer therefore reaches the end of one file and only then starts the next file's thread and first read, so the two never overlap.
- Across the runs of a section, `LoserTree::InitializeIfNeeded` advances its leaves one at a time and each advance blocks on that leaf's own first read. The leaves are independent sorted runs, so with `k` runs in a section the merge waits for `k` first reads in sequence even though they could all be in flight at once.

The latency being serialized is dominated by remote storage access, so on a section with several runs and several files per run it is a multiple of a single round trip spent doing nothing but waiting.

## Solution

Add one optional `Warmup()` verb to the two reader abstractions and call it where the consumer already knows what it is going to read next, so the first read of that reader is issued while the current one is still being consumed.

- Declare `virtual Status Warmup()` with a default `Status::OK()` body on both `FileBatchReader` and the internal `KeyValueRecordReader`. A reader with nothing to start, or one not yet ready to start it, is a no-op, so no existing implementation has to change and warming can never fail a read.
- Implement it once, in `PrefetchFileBatchReaderImpl`, by extracting the lazy background-thread start into `EnsureBackgroundThread()` and sharing that between `NextBatchWithBitmap()` and `Warmup()`, so warming and reading cannot diverge. Warmup returns OK without starting anything when the read ranges are not refreshed yet or the reader is shut down; the read path keeps rejecting the former, so a genuinely unprepared read is not hidden.
- Forward the call through the wrappers a data file's stack is built from (`FieldMappingReader`, `CompleteRowTrackingFieldsBatchReader`, `ApplyDeletionVectorBatchReader`, `ApplyBitmapIndexBatchReader`, `ShreddingFileReader`, `VectorFileBatchReader`, `DelegatingPrefetchReader`) and through `KeyValueDataFileRecordReader` on the KeyValue side. `FieldMappingReader` is the outermost wrapper of every file's stack, so swallowing the call there would leave the whole stack cold, and `DelegatingPrefetchReader` returns OK when prefetching is off because it bypasses the prefetch layer entirely in that case and a bypassed reader has no background thread to start.
- Warm one file ahead in `ConcatKeyValueRecordReader::NextBatch()`, which is idempotent and so costs one virtual call plus a pointer test per already-warm reader. The lookahead is one file because each warm file holds a prefetch queue of its own, and a strictly sequential consumer cannot use more overlap than the file it will read next.
- Warm every leaf in `LoserTree::InitializeIfNeeded()` before the advance loop, in the same order the loop consumes them, so the leaf the merge blocks on first is the one whose read was started first and the loop then collects reads that are already in flight.
- Build a section's readers in one pass: flatten the section's runs into one file list, call `CreateRawFileReaders` once for it, and hand the readers back to their run by position, with both section read paths (merge and raw) going through that one build. This drops `CreateReaderForRun`, gives one `FieldMappingBuilder` per section instead of one per run, and keeps the two paths from producing different reader stacks for the same section. Since readers are matched to runs positionally, a build that dropped a reader would shift every following file into the wrong run, so the existing `assert` becomes an explicit `Status::Invalid`.

## Anything else?

Warming changes when a file is opened, never what it reads: ordering, merging, filtering, deletion vectors and metrics are untouched, and warmup is unconditional with no configuration knob.

Two known limits, both left as follow-ups. Warming does not cross a section boundary yet, because sections are wrapped in `SortMergeKeyValueRecordReader` inside the outer `ConcatKeyValueRecordReader` and that wrapper holds a `SortMergeReader`, which has no warmup verb to forward to. And the `min-heap` sort engine gets the intra-run warmup but not the cross-run warmup at merge initialization, since only `LoserTree::InitializeIfNeeded` warms its leaves.

`FileBatchReader` is a public exported class, so the new virtual is additive and source compatible for every existing subclass, but it does change the vtable layout of that class.

lucasfang:
### Purpose

Linked issue: close #289

Reading a primary-key table in merge-on-read mode serializes the first read of every data file behind the consumption of the previous one, because `PrefetchFileBatchReaderImpl` starts its background thread lazily, inside the first `NextBatchWithBitmap()`. Within a sorted run the next file's first read starts only once the current file is exhausted, and across a section's `k` runs `LoserTree::InitializeIfNeeded` blocks on each leaf's first read in turn. On remote storage that is one round trip per file instead of overlapping them.

This PR adds one optional `Warmup()` verb to the two reader abstractions and calls it where the consumer already knows what it is going to read next, so that reader's first read is issued while the current one is still being consumed. Warming changes when a file is opened, never what it reads: ordering, merging, filtering, deletion vectors and metrics are untouched.

How far it prepares the next file is a public option, because it trades memory for latency and a scan that stops early - a `LIMIT`, or a predicate selective enough to empty a split - pays for whatever it already warmed. `WarmupLevel` sits in `include/paimon/utils/prefetch_cache_config.h` next to the `CacheConfig` it interacts with, set through `ReadContextBuilder::SetWarmupLevel` and read back through `ReadContext::GetWarmupLevel`. Each level takes the next file one step further along the read pipeline, so a higher one hides more latency and commits more memory:

- `NONE` - the behavior from before warmup existed: no extra memory, no background thread.
- `RAW` - fetch only the next file's raw, still-compressed bytes, which overlaps the remote fetch without materializing a decoded batch ahead of the read.
- `DECODED` (default) - also start the background decode loop, so decoded batches are ready before the file is read.

Changes:

- `virtual void Warmup()` with an empty default body on `BatchReader` (public) and on the internal `KeyValueRecordReader`, so no existing implementation has to change. It returns `void` on purpose: the caller may warm a reader it never ends up reading, and a hint about a file nobody reads must not be able to fail the read in progress.
- `PrefetchFileBatchReaderImpl::Warmup()` is the only real implementation and interprets the level: `NONE` returns at once, `RAW` calls the new `WarmCacheOnce()`, `DECODED` calls `EnsureBackgroundThread()`, into which the lazy thread start moved so the read path and the warmup path share one start. All three return early when the read ranges are not fresh, a state `NextBatchWithBitmap()` still rejects, so an unprepared read is reported there and not swallowed here. `ReadAheadCache::Init()` is not idempotent, so `WarmCacheOnce()` guards on `cache_warmed_`, which `CleanUp()` clears for the next read-range generation.
- Seven `FileBatchReader` wrappers forward the call - `FieldMappingReader`, `CompleteRowTrackingFieldsBatchReader`, `ApplyDeletionVectorBatchReader`, `ApplyBitmapIndexBatchReader`, `ShreddingFileReader`, `VectorFileBatchReader` and `DelegatingPrefetchReader` - plus `KeyValueDataFileRecordReader` on the KeyValue side. `DelegatingPrefetchReader` skips the call when `NeedPrefetch()` is false, because `GetReader()` then bypasses the prefetch layer and forwarding would decode the reader the consumer is itself reading.
- `ConcatKeyValueRecordReader` warms the reader it is consuming plus one file ahead (`kWarmupLookahead = 1`) on every `NextBatch()`, and `LoserTree::InitializeIfNeeded()` calls the new `WarmupLeaves()` before its advance loop, in consumption order, so a section of `k` runs has `k` files in flight at once - one per leaf, not two.
- `MergeFileSplitRead` now builds a section's runs with one `CreateRawFileReaders` call over the flattened file list and hands the readers back by position, the merge path and the raw path both going through it; `CreateReaderForRun` is removed. Because the matching is positional, the existing `assert` on the two counts becomes an explicit `Status::Invalid`: a release build would otherwise shift files into the wrong run instead of failing.
- The level travels with the rest of the read settings through `DataFileReadOptions::warmup_level`, filled by `AbstractSplitRead` and `FormatTableRead` from the `ReadContext` and forwarded by `DataFileReaderFactory::Open()`, the one place a production read builds the prefetching reader. `Create()` no longer defaults the parameter, so a dropped link is a compile error instead of a silent fallback to `DECODED`.
- System tables rebuild a `ReadContext` for the data table underneath them and copied every prefetch and cache setting but not this one, so a caller's `NONE` or `RAW` came back as `DECODED` on the table that does the actual reading; each chain is now a named `CreateDataReadContext()`, which is also what makes it testable.

Two known limits, left as follow-ups: warming does not cross a section boundary, because sections sit inside a `SortMergeKeyValueRecordReader` holding a `SortMergeReader` with no warmup verb to forward to; and the `min-heap` engine gets the intra-run warmup but not the cross-run one, since only `LoserTree` warms its leaves - `loser-tree` is the default.

One build fix rides along, plus the CI gap that hid it; both can be split into their own PR if preferred. `PAIMON_RETURN_NOT_OK_FROM_ARROW` and `PAIMON_ASSIGN_OR_RAISE_IMPL_FROM_ARROW` called `ToPaimonStatus` by unqualified name, which does not resolve at global scope, where `benchmark/parquet_format_benchmark.cpp` keeps its helpers, so every expansion site there failed to compile; both now spell `::paimon::ToPaimonStatus`. CI never saw it because it builds nothing under `benchmark/`, the top-level `CMakeLists.txt` having added it only under `PAIMON_BUILD_BENCHMARKS`; it is now added under `PAIMON_BUILD_TESTS OR PAIMON_BUILD_BENCHMARKS`, so a tests-only build registers the two gtest targets already gated there (ctest count 28 to 30). Building them exposed two dormant defects, fixed here too: the benchmark test linked only the Parquet format library, so the `CoreOptions` it constructs could not resolve `manifest.format` - default `avro` - and all 8 cases failed, so it now links the benchmark's own `PAIMON_BENCHMARK_STATIC_LINK_LIBS`; and its negative codec assertion wrote a real `"lz4"` file, which Parquet rejects only mid-write through a Thrift conversion whose `default` branch is `DCHECK(false)`, an abort in the six Debug and sanitizer CI jobs, so it now pins the same fact at name resolution with no write.

### Tests

18 new cases: 17 appended to test files that already build in an existing target, plus a new `data_file_reader_factory_test.cpp` registered in `src/paimon/CMakeLists.txt`. `MockFileBatchReader` gained a `Warmup()` override and a `GetWarmupCount()` accessor, which is what makes the forwarding cases possible without a new mock per wrapper.

- Forwarding, one case per link, each asserting the inner reader's count goes 0 to 1: `ConcatKeyValueRecordReaderTest.TestWarmupLooksOneReaderAhead`, one case per forwarding wrapper, and `KeyValueDataFileRecordReaderTest.TestWarmupForwardsToInnerReader`. The Concat case also pins that nothing is warmed before a read asks for it, and that a warmed Concat read to the end yields the same KeyValues as the unwarmed one.
- Merge: `SortMergeReaderTest.TestLoserTreeWarmsAllLeavesBeforeAdvancingAny` asserts a three-leaf tree records exactly `warm2, warm1, warm0, read2, read1, read0`, and that a second `InitializeIfNeeded()` records nothing.
- Levels: `PrefetchFileBatchReaderImplTest.TestWarmupLevelNone`, `.TestWarmupLevelRaw` and `.TestWarmupLevelDecoded`, each over `{parquet, orc} x {read-ahead cache on, off}`, assert whether the background decode loop started and then read every row against the unwarmed expectation; two more cover both branches of the `NeedPrefetch()` guard; `.TestWarmupCacheRearmsForNewReadRangeGeneration` covers the `cache_warmed_` lifecycle across a `RefreshReadRanges()`.
- Option plumbing: `ReadContextTest.TestSetWarmupLevel`, `SystemTableTest.TestNewReadPropagatesWarmupLevel` - for `$ro`, `$audit_log` and `$binlog`, as `EXPECT_EQ` so one broken chain does not hide another - and `DataFileReaderFactoryTest.OpenForwardsWarmupLevelToPrefetchReader`.

The assertions are shaped by mutation testing: seven deletions that all compile - the `WarmupLeaves()` call, all seven wrapper forwards, `DECODED` degraded to `RAW`, `cache_warmed_.store(false)`, the `NeedPrefetch()` guard, and `SetWarmupLevel` from each system-table chain - each left the full core and common suites green before, and each is now killed by the case that pins it. The changed paths also stay covered by the existing `ConcatKeyValueRecordReaderTest`, `SortMergeReaderTest` (both engines), `MergeFileSplitReadTest` and `PrefetchFileBatchReaderImplTest`.

Validation, local Release build with `PAIMON_BUILD_TESTS=ON` and `PAIMON_BUILD_BENCHMARKS=ON`, all passed:

```bash
cmake --build build --target unittest -j 96
cd build && ctest -j 96 -L unittest --output-on-failure
./build/release/paimon-core-test
./build/release/paimon-common-test
./build/release/paimon-read-inte-test
./build/release/paimon-write-and-read-inte-test
./build/release/paimon-pk-compaction-inte-test
```

30 of 30 targets under the `unittest` label, including the two the `CMakeLists.txt` change brings into a tests-only build; `paimon-core-test` 2053; `paimon-common-test` 1572 ran, 1570 passed and 2 skipped - the cache-disabled parameters of `TestWarmupCacheRearmsForNewReadRangeGeneration`, and the only skips in any run; `paimon-read-inte-test` 294; `paimon-write-and-read-inte-test` 198; `paimon-pk-compaction-inte-test` 38. Not run locally, left to CI: the sanitizer jobs and `ci/scripts/build_paimon.sh`. The sanitizers matter more than usual here, because `DECODED` warmup and the `NeedPrefetch()` guard both start work on a background thread ahead of the reader's own thread.

### API and Format

Three public API changes under `include/paimon/`, all additive except one constructor. No storage format and no protocol change: nothing written to disk, to a manifest, to a file footer or over the wire is touched, and warming changes when a file is opened, never what is read.

- `reader/batch_reader.h`: `virtual void Warmup()` with an empty default body on the exported `BatchReader`. Source compatible, but not binary compatible - a new virtual changes the vtable layout and shifts the slot of every virtual after it - so everything must be recompiled together.
- `utils/prefetch_cache_config.h`: new `enum class PAIMON_EXPORT WarmupLevel { NONE, RAW, DECODED }`. Purely additive and binary compatible; a scoped enum carries no data and no functions.
- `read_context.h`: `SetWarmupLevel()` and `GetWarmupLevel()` are new non-virtual members and the builder is a pimpl, so `sizeof(ReadContextBuilder)` is unchanged. `ReadContext`'s **public** constructor gained a trailing `WarmupLevel`, source-breaking for a caller that constructs it directly - the class documents going through `ReadContextBuilder`, and nothing here does - and it gained a private member, so `sizeof(ReadContext)` changes; no virtuals, so layout only.

Everything else is under `src/paimon/` and not exported: `KeyValueRecordReader::Warmup()`, `LoserTree::WarmupLeaves()`, `WarmCacheOnce()`, `EnsureBackgroundThread()`, the `MergeFileSplitRead` signature changes, `DataFileReadOptions::warmup_level`, the two `CreateDataReadContext()` methods, and `PrefetchFileBatchReaderImpl::Create()`'s new trailing `WarmupLevel` parameter, which has no default.

### Documentation

The three new methods need no `.rst` edit of their own: `docs/source/api/read.rst` already pulls `paimon::ReadContextBuilder`, `paimon::ReadContext` and `paimon::BatchReader` in through `doxygenclass` with `:members: :undoc-members:`, and `api/file_format.rst` does the same for `paimon::FileBatchReader`, so they render automatically with the doc comments this PR adds.

One documentation line is added: `.. doxygenenum:: paimon::WarmupLevel` in `api/read.rst`. Without it no `.rst` references the new public enum, so the `@see WarmupLevel` in `SetWarmupLevel`'s doc comment is a dangling reference and the meaning of the three levels appears nowhere in the rendered docs even though the option is public. The repository already uses `doxygenenum` for `paimon::FieldType`, `paimon::ByteOrder` and `paimon::SeekOrigin`. No user-guide page is affected: `user_guide/prefetch.rst` documents no user-facing option at all.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/include/paimon/read_context.h
+++ b/include/paimon/read_context.h
@@ -62,7 +62,8 @@ class PAIMON_EXPORT ReadContext {
                 const std::shared_ptr<RealtimeContext>& realtime_context,
                 const std::map<std::string, std::string>& options, bool read_ahead_cache_enabled,
                 const CacheConfig& cache_config, const std::shared_ptr<Cache>& cache,
-                const std::shared_ptr<FormatTable>& format_table);
+                const std::shared_ptr<FormatTable>& format_table, WarmupMode warmup_mode);
+
     ~ReadContext();
 
     const std::string& GetPath() const {
@@ -150,6 +151,10 @@ class PAIMON_EXPORT ReadContext {
         return format_table_;
     }
 
+    WarmupMode GetWarmupMode() const {
+        return warmup_mode_;
+    }
+
     /// Whether a read schema (C ArrowSchema) for nested column pruning was provided.
     bool HasReadSchema() const {
         return read_schema_ != nullptr && read_schema_->release != nullptr;
@@ -190,6 +195,7 @@ class PAIMON_EXPORT ReadContext {
     CacheConfig cache_config_;
     std::shared_ptr<Cache> cache_;
     std::shared_ptr<FormatTable> format_table_;
+    WarmupMode warmup_mode_;
     // Owns schema resources and releases ArrowSchema::release in destructor.
     std::unique_ptr<ArrowSchema> read_schema_;
 };
@@ -350,6 +356,16 @@ class PAIMON_EXPORT ReadContextBuilder {
     /// @param config The cache configuration to use.
     /// @return Reference to this builder for method chaining.
     ReadContextBuilder& WithCacheConfig(const CacheConfig& config);
+
+    /// Set how aggressively the reader warms up the next file before it is read.
+    ///
+    /// Warmup overlaps remote-storage latency with the read of the current file. More aggressive
+    /// modes hide more latency but use more memory, and may warm files that a query never reads
+    /// (for example when a LIMIT stops the scan early).
+    /// @param mode The warmup aggressiveness to use (default: WarmupMode::FULL).
+    /// @return Reference to this builder for method chaining.
+    /// @see WarmupMode
+    ReadContextBuilder& SetWarmupMode(WarmupMode mode);
 
     /// Set the total number of batches to prefetch across all files.
     ///
--- a/include/paimon/read_context.h
+++ b/include/paimon/read_context.h
@@ -364,6 +364,9 @@ class PAIMON_EXPORT ReadContextBuilder {
     /// when a LIMIT stops the scan early).
     /// @param level The warmup level to use (default: WarmupLevel::DECODED).
     /// @return Reference to this builder for method chaining.
+    /// @note WarmupLevel::RAW warms the read-ahead cache, so it has no effect and behaves like
+    /// WarmupLevel::NONE when the cache is off (see SetReadAheadCacheEnabled()).
+    /// WarmupLevel::DECODED does not depend on the cache.
     /// @see WarmupLevel
     ReadContextBuilder& SetWarmupLevel(WarmupLevel level);
 
--- a/include/paimon/read_context.h
+++ b/include/paimon/read_context.h
@@ -62,7 +62,7 @@ class PAIMON_EXPORT ReadContext {
                 const std::shared_ptr<RealtimeContext>& realtime_context,
                 const std::map<std::string, std::string>& options, bool read_ahead_cache_enabled,
                 const CacheConfig& cache_config, const std::shared_ptr<Cache>& cache,
-                const std::shared_ptr<FormatTable>& format_table, WarmupMode warmup_mode);
+                const std::shared_ptr<FormatTable>& format_table, WarmupLevel warmup_level);
 
     ~ReadContext();
 
@@ -151,8 +151,8 @@ class PAIMON_EXPORT ReadContext {
         return format_table_;
     }
 
-    WarmupMode GetWarmupMode() const {
-        return warmup_mode_;
+    WarmupLevel GetWarmupLevel() const {
+        return warmup_level_;
     }
 
     /// Whether a read schema (C ArrowSchema) for nested column pruning was provided.
@@ -195,7 +195,7 @@ class PAIMON_EXPORT ReadContext {
     CacheConfig cache_config_;
     std::shared_ptr<Cache> cache_;
     std::shared_ptr<FormatTable> format_table_;
-    WarmupMode warmup_mode_;
+    WarmupLevel warmup_level_;
     // Owns schema resources and releases ArrowSchema::release in destructor.
     std::unique_ptr<ArrowSchema> read_schema_;
 };
@@ -357,15 +357,15 @@ class PAIMON_EXPORT ReadContextBuilder {
     /// @return Reference to this builder for method chaining.
     ReadContextBuilder& WithCacheConfig(const CacheConfig& config);
 
-    /// Set how aggressively the reader warms up the next file before it is read.
+    /// Set how far the reader prepares the next file before it is read.
     ///
-    /// Warmup overlaps remote-storage latency with the read of the current file. More aggressive
-    /// modes hide more latency but use more memory, and may warm files that a query never reads
-    /// (for example when a LIMIT stops the scan early).
-    /// @param mode The warmup aggressiveness to use (default: WarmupMode::FULL).
+    /// Warmup overlaps remote-storage latency with the read of the current file. Higher levels hide
+    /// more latency but use more memory, and may warm files that a query never reads (for example
+    /// when a LIMIT stops the scan early).
+    /// @param level The warmup level to use (default: WarmupLevel::DECODED).
     /// @return Reference to this builder for method chaining.
-    /// @see WarmupMode
-    ReadContextBuilder& SetWarmupMode(WarmupMode mode);
+    /// @see WarmupLevel
+    ReadContextBuilder& SetWarmupLevel(WarmupLevel level);
 
     /// Set the total number of batches to prefetch across all files.
     ///
--- a/include/paimon/reader/batch_reader.h
+++ b/include/paimon/reader/batch_reader.h
@@ -87,15 +87,16 @@ class PAIMON_EXPORT BatchReader {
     ///
     /// This is an optional hint and never changes what the reader returns: ordering, filtering and
     /// metrics are the same with or without it. A reader with nothing to start, or one that is not
-    /// yet ready to start it, returns OK unchanged, so an OK result does not mean work was
-    /// started. Calling it before the reader is configured (for example before `SetReadSchema()` on
-    /// a file reader), or after `Close()`, is such a no-op.
+    /// yet ready to start it, does nothing. Calling it before the reader is configured (for example
+    /// before `SetReadSchema()` on a file reader), or after `Close()`, is such a no-op.
+    ///
+    /// It reports no error on purpose: the caller may warm up a reader it never ends up reading,
+    /// and a hint about a file nobody reads must not fail the read in progress. An implementation
+    /// that cannot start its work leaves it to be started by the first read, which reports the
+    /// failure itself.
     /// @warning The call starts background work owned by this reader, so it must be made from the
     /// same thread that reads this reader, and not concurrently with any other call on it.
-    /// @return The status of the operation.
-    virtual Status Warmup() {
-        return Status::OK();
-    }
+    virtual void Warmup() {}
 
     /// Retrieves the reader's metrics.
     /// Note that calling this method frequently may incur significant performance overhead.
--- a/include/paimon/reader/batch_reader.h
+++ b/include/paimon/reader/batch_reader.h
@@ -81,23 +81,6 @@ class PAIMON_EXPORT BatchReader {
     /// bitmap. Noted that the returned bitmap has at least one valid row id.
     virtual Result<ReadBatchWithBitmap> NextBatchWithBitmap();
 
-    /// Starts whatever background work this reader would otherwise start on its first read, so a
-    /// caller that knows this reader is next can pay that startup while still consuming the
-    /// previous one.
-    ///
-    /// This is an optional hint and never changes what the reader returns: ordering, filtering and
-    /// metrics are the same with or without it. A reader with nothing to start, or one that is not
-    /// yet ready to start it, does nothing. Calling it before the reader is configured (for example
-    /// before `SetReadSchema()` on a file reader), or after `Close()`, is such a no-op.
-    ///
-    /// It reports no error on purpose: the caller may warm up a reader it never ends up reading,
-    /// and a hint about a file nobody reads must not fail the read in progress. An implementation
-    /// that cannot start its work leaves it to be started by the first read, which reports the
-    /// failure itself.
-    /// @warning The call starts background work owned by this reader, so it must be made from the
-    /// same thread that reads this reader, and not concurrently with any other call on it.
-    virtual void Warmup() {}
-
     /// Retrieves the reader's metrics.
     /// Note that calling this method frequently may incur significant performance overhead.
     /// @return A shared pointer to the `Metrics` object.
--- a/include/paimon/reader/batch_reader.h
+++ b/include/paimon/reader/batch_reader.h
@@ -81,6 +81,22 @@ class PAIMON_EXPORT BatchReader {
     /// bitmap. Noted that the returned bitmap has at least one valid row id.
     virtual Result<ReadBatchWithBitmap> NextBatchWithBitmap();
 
+    /// Starts whatever background work this reader would otherwise start on its first read, so a
+    /// caller that knows this reader is next can pay that startup while still consuming the
+    /// previous one.
+    ///
+    /// This is an optional hint and never changes what the reader returns: ordering, filtering and
+    /// metrics are the same with or without it. A reader with nothing to start, or one that is not
+    /// yet ready to start it, returns OK unchanged, so an OK result does not mean work was
+    /// started. Calling it before the reader is configured (for example before `SetReadSchema()` on
+    /// a file reader), or after `Close()`, is such a no-op.
+    /// @warning The call starts background work owned by this reader, so it must be made from the
+    /// same thread that reads this reader, and not concurrently with any other call on it.
+    /// @return The status of the operation.
+    virtual Status Warmup() {
+        return Status::OK();
+    }
+
     /// Retrieves the reader's metrics.
     /// Note that calling this method frequently may incur significant performance overhead.
     /// @return A shared pointer to the `Metrics` object.
--- a/include/paimon/reader/file_batch_reader.h
+++ b/include/paimon/reader/file_batch_reader.h
@@ -63,6 +63,14 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
 
     /// Get whether or not support read precisely while bitmap pushed down.
     virtual bool SupportPreciseBitmapSelection() const = 0;
+
+    /// Starts whatever background work the reader would otherwise start on its
+    /// first read, so a caller that knows this reader is next can pay that
+    /// startup while still consuming the previous one. Optional: a reader with
+    /// nothing to start, or one not yet ready to start it, returns OK unchanged.
+    virtual Status Warmup() {
+        return Status::OK();
+    }
 };
 
 }  // namespace paimon
--- a/include/paimon/reader/file_batch_reader.h
+++ b/include/paimon/reader/file_batch_reader.h
@@ -63,6 +63,23 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
 
     /// Get whether or not support read precisely while bitmap pushed down.
     virtual bool SupportPreciseBitmapSelection() const = 0;
+
+    /// Starts whatever background work this reader would otherwise start on its first read, so a
+    /// caller that knows this reader is next can pay that startup while still consuming the
+    /// previous one.
+    ///
+    /// This is an optional hint and never changes what the reader returns: ordering, filtering and
+    /// metrics are the same with or without it. A reader with nothing to start, or one that is not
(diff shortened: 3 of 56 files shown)


include/paimon/reader/batch_reader.h
```diff
@@ -81,6 +81,23 @@ class PAIMON_EXPORT BatchReader {
...
+    /// yet ready to start it, does nothing. Calling it before the reader is configured (for example
+    /// before `SetReadSchema()` on a file reader), or after `Close()`, is such a no-op.
+    ///
+    /// It reports no error on purpose: the caller may warm up a reader it never ends up reading,
+    /// and a hint about a file nobody reads must not fail the read in progress. An implementation
+    /// that cannot start its work leaves it to be started by the first read, which reports the
+    /// failure itself.
+    /// @warning The call starts background work owned by this reader, so it must be made from the
+    /// same thread that reads this reader, and not concurrently with any other call on it.
+    virtual void Warmup() {}
```
lxy-9602: Consider move to file batch reader?

src/paimon/common/reader/data_file_reader_factory.cpp
```diff
@@ -77,15 +77,16 @@ Result<std::unique_ptr<FileBatchReader>> DataFileReaderFactory::Open(
...
+        PAIMON_ASSIGN_OR_RAISE(
+            std::unique_ptr<PrefetchFileBatchReaderImpl> prefetch_reader,
+            PrefetchFileBatchReaderImpl::Create(
+                file_path, file_size, reader_builder, file_system,
+                read_options.prefetch_max_parallel_num, read_options.read_batch_size,
+                read_options.prefetch_batch_count, read_options.adaptive_prefetch_strategy,
+                executor,
+                /*initialize_read_ranges=*/false, read_options.read_ahead_cache_enabled,
+                read_options.cache_config, read_options.prefetch_io_metrics_enabled, pool,
+                arrow_pool, read_options.warmup_level));
```
lxy-9602: Please don’t put the new parameter at the end. Keeping `pool` as the last parameter is more consistent.

include/paimon/reader/file_batch_reader.h
```diff
@@ -63,6 +63,23 @@ class PAIMON_EXPORT FileBatchReader : public BatchReader {
...
+    /// metrics are the same with or without it. A reader with nothing to start, or one that is not
+    /// yet ready to start it, does nothing. Calling it before the reader is configured (for example
+    /// before `SetReadSchema()`), or after `Close()`, is such a no-op.
+    ///
+    /// It reports no error on purpose: the caller may warm up a reader it never ends up reading,
+    /// and a hint about a file nobody reads must not fail the read in progress. An implementation
+    /// that cannot start its work leaves it to be started by the first read, which reports the
+    /// failure itself.
+    /// @warning The call starts background work owned by this reader, so it must be made from the
+    /// same thread that reads this reader, and not concurrently with any other call on it.
```
lxy-9602: Is this something that can be customized by plugins? If not, should we clarify that in the comment? The behavior described by “The call starts background work owned by this reader” seems to be fixed by the implementation.

include/paimon/read_context.h
```diff
@@ -351,6 +357,16 @@ class PAIMON_EXPORT ReadContextBuilder {
...
+    /// Set how far the reader prepares the next file before it is read.
+    ///
+    /// Warmup overlaps remote-storage latency with the read of the current file. Higher levels hide
+    /// more latency but use more memory, and may warm files that a query never reads (for example
+    /// when a LIMIT stops the scan early).
+    /// @param level The warmup level to use (default: WarmupLevel::DECODED).
+    /// @return Reference to this builder for method chaining.
+    /// @see WarmupLevel
+    ReadContextBuilder& SetWarmupLevel(WarmupLevel level);
+
```
lxy-9602, line 372: `Warmup` and cache seem to be tightly coupled in the current implementation. `SetWarmupLevel(WarmupLevel::RAW)` with `SetReadAheadCacheEnabled(false)` appears to be effectively equivalent to `SetWarmupLevel(WarmupLevel::None)`. Please document this limitation in the comments.

src/paimon/common/reader/prefetch_file_batch_reader_impl.h
```diff
@@ -123,10 +125,19 @@ class PrefetchFileBatchReaderImpl : public PrefetchFileBatchReader {
...
     void Workloop();
+    /// Starts the background prefetch thread if it is not running yet. The first read does this
+    /// itself; Warmup() is the same call made earlier, so the two must not diverge.
+    void EnsureBackgroundThread();
+    /// Initializes and warms the read-ahead cache at most once per read-range generation, using the
+    /// first reader's PreBufferRange(). Shared by Workloop() (DECODED) and Warmup() (RAW), so the
+    /// two never Init the cache twice. A no-op when there is no cache. Errors are recorded via
+    /// SetReadStatus() rather than returned: a warmup hint for a file that may never be read must
+    /// not fail an in-flight read.
+    void WarmCacheOnce();
```
lxy-9602, line 143: `exchange(true)` only elects one initializer; it does not make other threads wait for `cache_->Init()` to finish. If RAW warmup and `Workloop()` race, the losing thread may continue with an uninitialized cache. Please either rely on the documented non-concurrent contract and remove this race claim, or synchronize initialization completion.

src/paimon/core/io/complete_row_tracking_fields_reader_test.cpp
```diff
@@ -486,4 +486,21 @@ TEST_F(CompleteRowTrackingFieldsBatchReaderTest, TestOnlyReadRowTrackingFields)
     CheckResult(src_array, /*first_row_id=*/100, /*snapshot_id=*/4, read_schema, target_array);
 }
 
+// Warmup() is only a hint, but it must reach the wrapped reader: this wrapper sits between the
+// split read and the file, and swallowing the hint here would leave the whole stack below it cold
+// no matter which layer asked for the warmup.
+TEST_F(CompleteRowTrackingFieldsBatchReaderTest, TestWarmupForwardsToInnerReader) {
```
lxy-9602, line 489: It doesn’t seem necessary to add a comment like this to every case.

src/paimon/core/io/key_value_record_reader.h
```diff
@@ -39,6 +39,13 @@ class KeyValueRecordReader {
 
     virtual std::shared_ptr<Metrics> GetReaderMetrics() const = 0;
 
+    /// Starts whatever background work this reader would otherwise start on its first read, so a
+    /// caller that knows this reader is next can pay that startup while still consuming the
+    /// previous one. Optional and reporting no error, like `FileBatchReader::Warmup()`: a reader
+    /// with nothing to start does nothing, and a hint about a file nobody reads must not fail the
+    /// read in progress.
+    virtual void Warmup() {}
```
lxy-9602: Is it a good idea to keep an empty `{}` here? Could this cause a newly added `KeyValueRecordReader` in the future to miss forwarding `Warmup()`?

src/paimon/core/operation/merge_file_split_read.cpp
```diff
@@ -662,14 +662,46 @@ MergeFileSplitRead::CreateRecordReadersForSection(
...
+    for (const SortedRun& run : section) {
+        const std::vector<std::shared_ptr<DataFileMeta>>& run_files = run.Files();
+        section_files.insert(section_files.end(), run_files.begin(), run_files.end());
+    }
+    PAIMON_ASSIGN_OR_RAISE(
+        std::vector<std::unique_ptr<FileBatchReader>> raw_file_readers,
+        CreateRawFileReaders(partition, section_files, read_schema_, predicate, dv_factory,
+                             /*row_ranges=*/{}, data_file_path_factory,
+                             /*extra_format_options=*/{}));
+    // A build that dropped a reader would silently shift every following file into the wrong run.
```
lxy-9602: I’m not following how the refactoring of `CreateReaderForRun` is related to the current warmup support. As far as I see, `CreateRawFileReaders` doesn’t introduce any additional concurrency. The current changes also add more validation, so it doesn’t seem like they would have much impact on the final performance?

include/paimon/reader/file_batch_reader.h
lucasfang: Good catch — that wording described one implementation's behavior as if it were the base-class contract. Plugins are free to define what Warmup() starts, including doing nothing at all.

src/paimon/common/reader/prefetch_file_batch_reader_impl.h
lucasfang: You're right, exchange(true) only elects an initializer and never makes the loser wait — the race claim was wrong.Went with the first option: relying on the documented non-concurrent contract. Warmup() must be called from the thread that reads the reader, so a RAW warmup runs WarmCacheOnce() on that thread before the same thread starts the background thread running Workloop(). Thread creation gives happens-before, so the two calls are ordered rather than concurrent, and the loser of the exchange can never proceed on a cache that another thread is still initializing. Comments in both the header and the implementation now say exactly that, and state that the flag's only job is to avoid a second non-idempotent Init(). No logic change.

lxy-9602 (approved):
+1

Issue #197, opened 2026-08-11
[Feature] Support richer data types, starting with VECTOR<T, N> based on PIP-40


ChaomingZhangCN:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-cpp/issues) and found nothing similar.

### Motivation

Apache Paimon Java has introduced the VECTOR<T, N> data type based on [PIP-40](https://cwiki.apache.org/confluence/spaces/PAIMON/pages/399279132/PIP-40+Introduce+a+new+Vector+data+type). It supports schema representation, regular data-file storage such as Parquet, dedicated vector storage, reads, writes, and Data Evolution.

Paimon C++ currently recognizes .vector. file names for some file-level bookkeeping, but it does not yet provide a VECTOR logical type, Arrow mapping, serialization, storage, or end-to-end read and write support.

This issue implements the VECTOR roadmap item tracked in [#186](https://github.com/apache/paimon-cpp/issues/186).

### Solution

Introduce `VECTOR<T, N>` support incrementally, while keeping the schema and storage behavior compatible with Apache Paimon Java.

#### Phase 1: Schema and regular Parquet storage

- [x] Add `VECTOR<T, N>` to the Paimon C++ logical type system.
- [x] Implement schema JSON serialization and deserialization compatible with Paimon Java.
- [x] Map `VECTOR<T, N>` to Arrow `FixedSizeList<T, N>`.
- [x] Support the element types defined by PIP-40:
  - `BOOLEAN`
  - `TINYINT`
  - `SMALLINT`
  - `INT`
  - `BIGINT`
  - `FLOAT`
  - `DOUBLE`
- [x] Validate that the dimension is positive and fixed.
- [x] Validate that the written vector length equals `N`.
- [x] Reject null vector elements.
- [x] Support reading and writing VECTOR columns in regular Parquet data files.
- [x] Add end-to-end append-table tests.
- [x] Add Java/C++ schema and file compatibility tests.

##### Schema Evolution

- [x] Support adding and dropping VECTOR columns.
- [x] Support reading files written with previous table schemas.
- [x] Preserve VECTOR values when projecting files across schema versions.
- [x] Reject incompatible dimension changes, such as
      `VECTOR<FLOAT, 3>` to `VECTOR<FLOAT, 5>`.
- [x] Reject incompatible element-type changes.
- [x] Add schema-evolution integration tests for both append-only and
      supported primary-key tables.

##### Data Evolution Read/Write

[...]

PR #295, opened 2026-09-08
feat(vector): support data evolution reads and writes

ChaomingZhangCN:
### Purpose

Related issue: #197

Enable VECTOR columns in Parquet-backed, row-tracking append-only tables
with data evolution.

The existing Parquet VECTOR read/write path can support data evolution,
but schema validation currently rejects VECTOR fields in these tables.
This change removes that restriction and adds end-to-end coverage for
full-row writes, partial-column updates, and reads across schema versions.

This PR also:
- Includes VECTOR dimensions in nested type equality checks.
- Rejects incompatible VECTOR dimensions, element types, and conversions
  between VECTOR and non-VECTOR types when projecting historical schemas,
  including nested fields.
- Replaces two long-suffixed BIGINT literals in the existing Data Evolution
  predicate tests with explicit int64_t values to fix macOS linking.

Support remains limited to Parquet. Dedicated vector storage and
schema-commit validation for VECTOR type changes are outside this PR.

### Tests

Added end-to-end tests:
- `TestVectorReadWrite`: full writes, appended row groups, partial VECTOR
  updates, NULL transitions, all-NULL updates, and projected row-range reads.
- `TestNestedVectorReadWrite`: full and partial writes of a nullable STRUCT
  containing a VECTOR field.
- `TestVectorSchemaEvolution`: adding and dropping VECTOR columns,
  null-filling added columns in historical files, and partial writes
  across schema versions.
- `TestVectorSchemaEvolutionRejectsTypeChange`: read-time rejection of
  incompatible VECTOR dimensions and element types.

Added projection unit tests:
- `PruneDataTypeVector`
- `PruneDataTypeRejectsIncompatibleVectorEvolution`
- `PruneDataTypeRejectsNestedVectorEvolution`

Updated schema-validation coverage to accept VECTOR in Parquet Data
Evolution tables while retaining the ORC restriction.

Local validation on macOS arm64:
- Standard CMake builds of `paimon-core-test` and
  `paimon-data-evolution-table-test` succeeded.
- Related core tests: 108 passed.
- Full Data Evolution suite: 204 passed, 18 skipped.
- Related ordinary read/write tests: 33 passed.
- Related BLOB tests: 48 passed.
- cpplint, codespell, sphinx-lint, changed-line clang-format checks,
  and `git diff --check` passed.

Runtime results above use statically relinked test executables, without
a Literal compatibility shim. The default mixed-link Data Evolution
executable still aborts during Arrow JSON builder construction.
The full repository suite and Linux CI were not run locally.

New VECTOR integration tests explicitly skip non-Parquet configurations.

### API and Format

VECTOR values continue to use the existing Parquet LIST representation
and are restored as Arrow FixedSizeList values on read.

Changing a VECTOR's dimension or element type is not supported;
incompatible historical schemas are rejected during read projection.

### Documentation

Updated the VECTOR section of the data types user guide to document
Data Evolution read/write support, adding and dropping VECTOR columns,
and the restrictions on dimension and element-type changes.

### Generative AI tooling

Generated-by: OpenAI Codex CLI 0.153.4

### DIFF
--- a/src/paimon/core/utils/nested_projection_utils.cpp
+++ b/src/paimon/core/utils/nested_projection_utils.cpp
@@ -165,6 +165,13 @@ Result<bool> EqualWithFieldIds(const std::shared_ptr<arrow::DataType>& a,
     if (a->id() != b->id() || a->num_fields() != b->num_fields()) {
         return false;
     }
+    if (a->id() == arrow::Type::FIXED_SIZE_LIST) {
+        const auto& vector_a = checked_cast<const arrow::FixedSizeListType&>(*a);
+        const auto& vector_b = checked_cast<const arrow::FixedSizeListType&>(*b);
+        if (vector_a.list_size() != vector_b.list_size()) {
+            return false;
+        }
+    }
     if (a->num_fields() == 0) {
         return a->Equals(*b);
     }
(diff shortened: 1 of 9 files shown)


test/inte/data_evolution_table_test.cpp
```diff
@@ -1651,6 +1651,251 @@ TEST_P(DataEvolutionTableTest, TestPartitionWithPredicate) {
...
+        {Options::ROW_TRACKING_ENABLED, "true"},
+        {Options::DATA_EVOLUTION_ENABLED, "true"},
+        {Options::READ_BATCH_SIZE, "2"},
+    };
+    CreateTable(fields, /*partition_keys=*/{}, options);
+    std::string table_path = PathUtil::JoinPath(dir_->Str(), "foo.db/bar");
+    auto field_names = arrow::schema(fields)->field_names();
+    auto initial_array = std::dynamic_pointer_cast<arrow::StructArray>(
+        arrow::ipc::internal::json::ArrayFromJSON(arrow::struct_(fields), R"([
+            [1, [1, 2, 3], "a"],
```
lxy-9602, line 1677: Consider use `checked_pointer_cast`

src/paimon/core/utils/nested_projection_utils.cpp
```diff
@@ -193,10 +200,31 @@ Result<bool> EqualWithFieldIds(const std::shared_ptr<arrow::DataType>& a,
...
+        if (read_vector.list_size() == data_vector.list_size() &&
+            read_vector.value_type()->Equals(data_vector.value_type())) {
+            return Status::OK();
+        }
+    }
+    return Status::Invalid(
+        fmt::format("VECTOR type mismatch during schema evolution: data {} vs read {}",
+                    data_type->ToString(), read_type->ToString()));
+}
+
```
lxy-9602: Should we consider moving this function into `VectorUtils`?

ChaomingZhangCN: ok

test/inte/data_evolution_table_test.cpp
ChaomingZhangCN: ok

lxy-9602 (approved):
+1

Issue #298, opened 2026-09-08
[Bug] Snapshot live manifest entry cache serves stale entries after a rollback


wangyong9999:
### Search before asking

- [x] I searched in the [issues](https://github.com/apache/paimon-cpp/issues) and found nothing similar.

### Paimon-cpp version

main (ded7cfb8)

### Minimal reproduce step

1. Primary key table with one bucket, three commits (snapshots 1 to 3): `(k1, v1), (k2, v2)`, then `(k3, v3)`, then `(k1, v1b)`.
2. Scan it with a `Cache` on the `ScanContext` and `scan.manifest-entry-cache.max-snapshots = 2`; the live manifest entries of snapshot 3 are now cached for bucket 0.
3. Roll back and commit twice more, for example from Flink:

   ```sql
   CALL sys.rollback_to(`table` => 'default.t', snapshot_id => 1);
   INSERT INTO t VALUES ('k4', 'v4');
   INSERT INTO t VALUES ('k1', 'v1c');
   ```
   The rollback deletes `snapshot-2`, `snapshot-3` and the data files only they referenced; the inserts create a new `snapshot-2` and `snapshot-3`.
4. Scan again through the same `Cache`.

### What doesn't meet your expectations?

The second scan plans from the cached entries of the deleted snapshot 3: reading `k1` or `k3` fails with `Not exist: File .../bucket-0/data-....parquet not exists`, `k4` returns no row although it was committed, only `k2` is correct. It recovers once a fourth snapshot moves the ids past the cached ones.

`FileStoreScan::ReadManifestEntriesWithCache` treats `cached->snapshot_id == snapshot.Id()` as a hit (`file_store_scan.cpp:343`); nothing else identifies the snapshot, so an id reused after a rollback (or after dropping and recreating a table at the same path) matches entries of a different snapshot. The cache is off by default (`max-snapshots = 0`); a long-lived process that enables it for point lookups is exposed.

### Anything else?

Storing the snapshot's delta manifest list name with the cached entries and requiring it to match on a hit is enough: the name carries a UUID, so a rewritten snapshot never matches and falls through to the existing rebuild.

### Are you willing to submit a PR?

- [x] I'm willing to submit a PR!

wangyong9999:
Not pursuing this for now, see #302. Rolling back a table that is being served through the manifest entry cache is rare enough; if it comes up, the fix in #302 applies.

PR #300, opened 2026-09-08
fix(scan): ignore cached manifest entries of a rewritten snapshot

### LINKED ISSUE #299: [Bug] Snapshot live manifest entry cache matches on snapshot id only and serves stale entries after a rollback
### Search before asking

### Paimon-cpp version

main (ded7cfb8)

### Minimal reproduce step

1. Create a primary key table with one bucket and commit three times (snapshots 1 to 3), for example `(k1, v1), (k2, v2)`, then `(k3, v3)`, then `(k1, v1b)`.
2. Scan it with a `Cache` set on the `ScanContext` and `scan.manifest-entry-cache.max-snapshots = 2`, so the live manifest entries of snapshot 3 are cached for bucket 0.
3. Roll the table back and commit twice more, for example from Flink:

   ```sql
   CALL sys.rollback_to(`table` => 'default.t', snapshot_id => 1);
   INSERT INTO t VALUES ('k4', 'v4');
   INSERT INTO t VALUES ('k1', 'v1c');
   ```
   The rollback deletes `snapshot-2`, `snapshot-3` and the data files only they referenced; the two inserts create a new `snapshot-2` and `snapshot-3`.
4. Scan again through the same `Cache`.

The second scan plans from the cached entries of the deleted snapshot 3. Reading `k1` or `k3` fails with `Not exist: File .../bucket-0/data-....parquet not exists`, `k4` returns no row although it was committed, and only `k2` is correct. The table recovers once a fourth snapshot is committed and the cached ids fall behind.

`FileStoreScan::ReadManifestEntriesWithCache` treats `cached->snapshot_id == snapshot.Id()` as a hit (`file_store_scan.cpp:343`). Nothing else identifies the snapshot, so a snapshot id that is reused after a rollback (or after dropping and recreating a table at the same path) matches entries that belong to a different snapshot. With the default `max-snapshots = 0` the cache is off; any deployment that enables it for point lookups is exposed.

### Anything else?

Storing the snapshot's delta manifest list name next to the cached entries and requiring it to match on a hit is enough: the name carries a UUID, so a rewritten snapshot never matches, and a mismatch falls through to the existing rebuild path.

wangyong9999:
### Purpose

Linked issue: close #299

`FileStoreScan::ReadManifestEntriesWithCache` caches the merged live manifest entries of a bucket per snapshot and takes `cached->snapshot_id == snapshot.Id()` as a hit. A snapshot id does not identify a snapshot, though: `rollback_to` deletes the newest snapshots together with the data files only they referenced, and the next commits reuse the same ids for different content (dropping and recreating a table at the same path does the same). A process that keeps the cache across such a rollback then plans the rewritten snapshot from the entries of the deleted one: reads of the deleted files fail with `Not exist`, rows committed after the rollback are missing, and this lasts until the ids move past the cached ones.

The cached snapshot now records the delta manifest list it was built from. Its name carries a UUID, so a rewritten snapshot never matches; a mismatch falls through to the existing rebuild, which replaces the stale entries in place.

- `SnapshotLiveManifestEntries` keeps the delta manifest list per cached snapshot, `Put()` takes it, and the new `Find(snapshot_id, delta_manifest_list)` reports a hit only when both match. `LatestBeforeOrEqual()` is unchanged apart from carrying the name.
- `FileStoreScan::ReadManifestEntriesWithCache` looks the snapshot up with `Find()` and stores the name with the rebuilt entries.
- The serialized form gains the name per snapshot; the magic changes from `SMEC` to `SMED`, so bytes written by the old layout fail deserialization and are rebuilt instead of being misread.

### Tests

- `FileStoreScanTest.TestSnapshotLiveManifestEntries` (extended): `Find()` misses on another delta manifest list or another id, and putting the same id again replaces the stale entries in place.
- `FileStoreScanTest.TestSnapshotLiveManifestEntriesSerialization` (extended): the name round-trips.
- `TableScanTest.TestManifestEntryCacheIgnoresRewrittenSnapshot` (new): three commits are planned through an `LruCache` with `scan.manifest-entry-cache.max-snapshots = 2`, the table is rolled back to snapshot 1 the way `rollback_to` does it (snapshot files removed, `LATEST` moved), two more commits create a new snapshot 2 and 3, and the next plan through the same cache reads `k1=v1c, k2=v2, k4=v4` rather than the rows of the deleted snapshot; the plan after that is a cache hit.

### API and Format

No change under `include/`. The in-memory cache value format of `SnapshotLiveManifestEntries` changes (new magic); it is never written to storage.

### Documentation

No.

### Generative AI tooling

### DIFF
--- a/src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
+++ b/src/paimon/core/manifest/snapshot_live_manifest_entries.cpp
@@ -21,6 +21,7 @@
 
 #include <algorithm>
 #include <cstring>
+#include <string>
 #include <utility>
 
 #include "paimon/common/io/memory_segment_output_stream.h"
@@ -34,7 +35,9 @@
 namespace paimon {
 namespace {
 
-constexpr int32_t kMagic = 0x534d4543;  // SMEC
+// Bumped from SMEC when the delta manifest list was added to every cached snapshot; older bytes
+// fail the magic check and are rebuilt.
+constexpr int32_t kMagic = 0x534d4544;  // SMED
 
 size_t NormalizeMaxSnapshots(int32_t max_snapshots) {
     return static_cast<size_t>(std::max(0, max_snapshots));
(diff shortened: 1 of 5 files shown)


wangyong9999:
Folded into #302, which carries this commit together with the bucket selection fix so that #298 is fixed by one PR.

PR #301, opened 2026-09-08
perf(predicate): reduce redundant batch evaluation and materialization

gripleaf:
### Purpose

Linked issue: #240 

Reduce redundant work in batch predicate evaluation through three optimizations:

- **Candidate-row evaluation:** Evaluate AND children only on remaining matches and OR children only on remaining non-matches. Gather only the referenced column and skip index rewriting when the candidate set remains unchanged. Preserve eager evaluation where skipping rows could hide errors.
- **Field bounds checks:** Use `StructArray::num_fields()` instead of `fields().size()` to avoid materializing unrelated column wrappers.
- **Typed equality:** Compare supported Arrow values directly without constructing per-row Literals. Preserve the existing path for unsupported types and cases requiring the original conversion semantics.

Synthetic benchmark results: 8,192 rows, an equality predicate combined with 16 OR equality predicates; median of six rounds, ten evaluations per round.

| Leading predicate match rate | Original | Optimized | Speedup |
| --- | ---: | ---: | ---: |
| 1 / 8192 | 16.388 ms | 0.086 ms | 191× |
| 1 / 16 | 16.126 ms | 0.215 ms | 75× |
| All rows | 16.084 ms | 0.584 ms | 27.5× |

These are batch-evaluation microbenchmarks, not end-to-end read latencies. Candidate selection still adds approximately 8.5% overhead compared with eager typed equality when all rows match. No separate numerical benefit is claimed for the bounds-check optimization.

### Tests

Passed:

- 1,555 common tests, including 57 predicate tests.
- 220 Parquet tests.
- 3 cross-module Predicate ABI tests.
- Opt-in batch-evaluation microbenchmark, checking result equivalence on every iteration.
- Changed-file pre-commit checks and `git diff --check`.

Regression coverage includes nested AND/OR predicates, empty selections, slices, NULLs, dictionary columns, field bounds, error preservation, typed equality equivalence, binary values and floating-point fallbacks.

Validation used Linux x86_64, Release, C++17 and GCC 13.3. An unchanged `fields_comparator_test.cpp` triggered a GCC internal compiler error and was compiled separately with Clang 18. The build environment also required shared libunwind. Neither workaround modifies repository files. Debug, sanitizer and aarch64 builds were not run locally.

### API and Format

No changes to public headers under `include/paimon/`, storage formats or protocols. Selection-aware methods are internal implementation details.

### Documentation

No new user-facing configuration or feature. Internal comments document selection semantics, eager fallbacks and supported typed equality paths.

### Generative AI tooling

### DIFF
--- a/src/paimon/common/predicate/leaf_predicate_impl.h
+++ b/src/paimon/common/predicate/leaf_predicate_impl.h
@@ -45,10 +45,10 @@ class LeafPredicateImpl : public LeafPredicate, public PredicateFilter {
     Result<std::vector<char>> Test(const arrow::Array& array,
                                    arrow::MemoryPool* pool) const override {
         const auto& struct_array = checked_cast<const arrow::StructArray&>(array);
-        if (field_index_ >= static_cast<int32_t>(struct_array.fields().size())) {
+        if (field_index_ >= struct_array.num_fields()) {
             return Status::Invalid(
                 fmt::format("field index {} exceed field count {} in struct array", field_index_,
-                            struct_array.fields().size()));
+                            struct_array.num_fields()));
         }
         const auto& field_array = struct_array.field(field_index_);
         return leaf_function_.Test(*field_array, literals_, pool);
@@ -58,10 +58,10 @@ class LeafPredicateImpl : public LeafPredicate, public PredicateFilter {
                                            const std::vector<int64_t>& selection,
                                            arrow::MemoryPool* pool) const override {
         const auto& struct_array = checked_cast<const arrow::StructArray&>(array);
-        if (field_index_ >= static_cast<int32_t>(struct_array.fields().size())) {
+        if (field_index_ >= struct_array.num_fields()) {
             return Status::Invalid(
                 fmt::format("field index {} exceed field count {} in struct array", field_index_,
-                            struct_array.fields().size()));
+                            struct_array.num_fields()));
         }
         const auto& field_array = struct_array.field(field_index_);
         // Preserve eager conversion/comparison errors, even when the offending row is not
--- a/src/paimon/common/predicate/leaf_predicate_impl.h
+++ b/src/paimon/common/predicate/leaf_predicate_impl.h
@@ -66,11 +66,8 @@ class LeafPredicateImpl : public LeafPredicate, public PredicateFilter {
         const auto& field_array = struct_array.field(field_index_);
         // Preserve eager conversion/comparison errors, even when the offending row is not
         // selected. Dictionary and unsupported physical types also retain the original path.
-        // LIKE validates escapes while examining values; temporal/decimal conversions may fail
-        // on rows outside the candidate set as well.
-        if (leaf_function_.GetType() == Function::Type::LIKE ||
-            field_array->type_id() == arrow::Type::TIMESTAMP ||
-            field_array->type_id() == arrow::Type::DECIMAL128) {
+        // LIKE validates escapes while examining values, including unselected rows.
+        if (leaf_function_.GetType() == Function::Type::LIKE) {
             return PredicateFilter::TestSelected(array, selection, pool);
         }
         auto field_type = FieldTypeUtils::ConvertToFieldType(field_array->type_id());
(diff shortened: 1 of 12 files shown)


src/paimon/common/predicate/leaf_predicate_impl.h
```diff
@@ -44,15 +45,49 @@ class LeafPredicateImpl : public LeafPredicate, public PredicateFilter {
...
+                fmt::format("field index {} exceed field count {} in struct array", field_index_,
+                            struct_array.num_fields()));
+        }
+        const auto& field_array = struct_array.field(field_index_);
+        // Preserve eager conversion/comparison errors, even when the offending row is not
+        // selected. Dictionary and unsupported physical types also retain the original path.
+        // LIKE validates escapes while examining values; temporal/decimal conversions may fail
+        // on rows outside the candidate set as well.
+        if (leaf_function_.GetType() == Function::Type::LIKE ||
+            field_array->type_id() == arrow::Type::TIMESTAMP ||
```
wangyong9999: The TIMESTAMP and DECIMAL128 arms guard a per-row failure that cannot happen. `GetLiteralFromTimestampArray` and `GetLiteralFromDecimalArray` return a plain `std::vector<Literal>` with no failure path, `Decimal`'s constructor does not validate, and `Literal::CompareTo` on two same-type non-null TIMESTAMP/DECIMAL values returns a value rather than a Status. The only per-row Status a comparison leaf can raise is the type mismatch, and the literal loop just below already routes that to the eager path.

So these two arms only cost: any AND/OR with a timestamp or decimal child evaluates that child over the whole batch even when the candidate set is down to a handful of rows. Suggest dropping both and keeping the LIKE arm, whose escape validation really is reached per value.

src/paimon/common/predicate/predicate_test.cpp
```diff
@@ -1149,6 +1250,245 @@ TEST_F(PredicateTest, TestOr) {
...
+                ASSERT_EQ(actual, expected);
+            }
+        }
+    }
+}
+
+// Opt-in microbenchmark: no timing thresholds in CI. Compare the original eager Literal path,
+// bounds checks alone, bounds checks plus typed equality, and all three optimizations together.
+// These are incremental variants, not additive or independently attributable speedups.
+TEST_F(PredicateTest, DISABLED_BenchmarkBatchEvaluation) {
```
wangyong9999: This benchmark never runs in CI, prints to std::cout, and pulls `<chrono>` and `<iostream>` into the unit-test binary. It is also the only caller of `TestEager`'s `literal_equal` and `box_all_fields` parameters, so the shared correctness helper carries two extra modes purely to feed it.

The repo already has a `benchmark/` target for this. Suggest moving it there, or dropping it and keeping the numbers in the PR description.

lucasfang:
Thanks, the selection logic itself is correct and the test matrix is thorough. Both earlier review comments are addressed at this head: the TIMESTAMP/DECIMAL128 eager arms are gone and the benchmark moved to `benchmark/`. My concerns are about the performance model rather than correctness, and most of them come from the interaction with #306.

## The three commits have very different value

`b9061c0` (`num_fields()`) is independent, tiny and obviously correct. Please split it out with `TestLeafFieldBoundsOnWideSlicedBatch` and the integration test into its own PR so it can merge on its own; there is no reason to couple it to the selection work.

`4785b98` (typed equality) is a strict subset of #306 and should be dropped, see below.

The selection commits are correct but their payoff shrinks a lot once #306 lands, and the benchmark as posted does not isolate the cost that decides whether they are net positive.

gripleaf:
> Thanks, the selection logic itself is correct and the test matrix is thorough. Both earlier review comments are addressed at this head: the TIMESTAMP/DECIMAL128 eager arms are gone and the benchmark moved to `benchmark/`. My concerns are about the performance model rather than correctness, and most of them come from the interaction with #306.
> 
> ## The three commits have very different value
> `b9061c0` (`num_fields()`) is independent, tiny and obviously correct. Please split it out with `TestLeafFieldBoundsOnWideSlicedBatch` and the integration test into its own PR so it can merge on its own; there is no reason to couple it to the selection work.
> 
> `4785b98` (typed equality) is a strict subset of #306 and should be dropped, see below.
> 
> The selection commits are correct but their payoff shrinks a lot once #306 lands, and the benchmark as posted does not isolate the cost that decides whether they are net positive.

Thanks for the review. I’ve split the num_fields() optimization and its unit and integration tests into a separate branch as requested (#318). I’ll close this PR for now and evaluate the selection optimization against the latest main, including #306.
  I’ll open a new PR once I have benchmark results that isolate the selection overhead and demonstrate a net performance benefit.

PR #302, opened 2026-09-08
fix(scan): ignore cached manifest entries of a rewritten snapshot

### LINKED ISSUE #298: [Bug] Snapshot live manifest entry cache serves stale entries after a rollback

wangyong9999:
### Purpose

Linked issue: close #298

`FileStoreScan::ReadManifestEntriesWithCache` treated a matching snapshot id as a cache hit. Snapshot ids get reused: `rollback_to` deletes the newest snapshots together with the data files only they referenced, and the following commits recreate the same ids with different content (recreating a table at the same path does the same). A process that keeps the cache across such a rollback then plans the rewritten snapshot from the entries of the deleted one: reads of the deleted files fail with `Not exist`, rows committed after the rollback are missing, until the ids move past the cached ones.

- `SnapshotLiveManifestEntries` records the delta manifest list per cached snapshot; its name carries a UUID, so a rewritten snapshot never matches. `Put()` takes it, and the new `Find(snapshot_id, delta_manifest_list)` is the only hit path in `ReadManifestEntriesWithCache`. A miss falls through to the existing rebuild, which replaces the stale entries in place.
- The serialized form gains the name per snapshot and the magic moves from `SMEC` to `SMED`, so bytes of the old layout are rebuilt instead of misread.

### Tests

- `FileStoreScanTest.TestSnapshotLiveManifestEntries` and `TestSnapshotLiveManifestEntriesSerialization` (extended): `Find()` misses on another delta manifest list or another id, putting the same id again replaces the stale entries, and the name round-trips.
- `TableScanTest.TestManifestEntryCacheIgnoresRewrittenSnapshot` (new): three commits are planned through an `LruCache` with `scan.manifest-entry-cache.max-snapshots = 2`; the table is rolled back to snapshot 1 the way `rollback_to` does it (snapshot files removed, `LATEST` moved); two more commits create a new snapshot 2 and 3; the next plan through the same cache reads `k1=v1c, k2=v2, k4=v4` (without the fix: `k1=v1b, k2=v2, k3=v3`), and the plan after that is a cache hit.

### API and Format

No change under `include/`. The cache value of `SnapshotLiveManifestEntries` changes layout (new magic); it lives in memory only and is never written to storage.

### Documentation

No.

### Generative AI tooling

### DIFF
--- a/src/paimon/core/manifest/snapshot_live_manifest_entries.h
+++ b/src/paimon/core/manifest/snapshot_live_manifest_entries.h
@@ -24,6 +24,7 @@
 #include <map>
 #include <memory>
 #include <optional>
+#include <string>
 #include <vector>
 
 #include "paimon/core/manifest/manifest_entry.h"
@@ -38,17 +39,26 @@ class MemorySegment;
 ///
 /// This value object owns merged live manifest entries by snapshot id. It does not own or access a
 /// cache; callers are responsible for storing the serialized bytes in the cache layer.
+///
+/// A snapshot id alone does not identify a snapshot: after a rollback, or after a table is dropped
+/// and recreated at the same path, later commits reuse the ids of the deleted snapshots. Every
+/// cached snapshot therefore also records the delta manifest list it was built from, whose name is
+/// unique per commit, and `Find()` only reports a hit when both match.
 class SnapshotLiveManifestEntries {
  public:
     struct Entry {
         int64_t snapshot_id;
+        std::string delta_manifest_list;
         std::shared_ptr<const std::vector<ManifestEntry>> entries;
     };
 
     explicit SnapshotLiveManifestEntries(int32_t max_snapshots);
 
     std::optional<Entry> LatestBeforeOrEqual(int64_t snapshot_id) const;
-    void Put(int64_t snapshot_id, std::vector<ManifestEntry>&& entries);
+    /// The entries cached for exactly this snapshot: same id, same delta manifest list.
+    std::optional<Entry> Find(int64_t snapshot_id, const std::string& delta_manifest_list) const;
+    void Put(int64_t snapshot_id, const std::string& delta_manifest_list,
+             std::vector<ManifestEntry>&& entries);
     size_t Size() const;
 
     Result<std::shared_ptr<Bytes>> Serialize(const std::shared_ptr<MemoryPool>& pool) const;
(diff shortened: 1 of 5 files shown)


src/paimon/core/manifest/snapshot_live_manifest_entries.h
```diff
@@ -38,17 +39,26 @@ class MemorySegment;
...
  public:
     struct Entry {
         int64_t snapshot_id;
+        std::string delta_manifest_list;
         std::shared_ptr<const std::vector<ManifestEntry>> entries;
     };
 
     explicit SnapshotLiveManifestEntries(int32_t max_snapshots);
 
     std::optional<Entry> LatestBeforeOrEqual(int64_t snapshot_id) const;
```
wangyong9999, line 57: After this change nothing outside the tests calls `LatestBeforeOrEqual()`, and it is the id-only lookup that produced the stale hit in the first place. Unless an incremental rebuild from an older cached snapshot is planned, drop it in this PR; the eviction checks in the tests can go through `Find()` and `Size()`.

wangyong9999:
Closing for now: the rollback scenario is rare enough that we are not pursuing this change at the moment. The branch stays available if it becomes relevant.

PR #304, opened 2026-09-08
perf(parquet): avoid duplicate pre-buffering for mixed row groups

### LINKED ISSUE #240: [Feature] Improve scan performance in hot read paths

gripleaf:
### Purpose

  Linked issue: close #240 

  When a scan includes both fully matched and page-filtered row groups, Arrow’s automatic pre-buffering and Paimon’s range pre-buffering can read the same data twice.

  Let FileReaderWrapper manage a shared pre-buffer cache for both kinds of row groups. Refresh cache coverage when seeking, preserve Arrow’s standard handling for full-only scans, and retain existing fallback behavior.

  ### Tests

  Added regression tests in FileReaderWrapperTest:

  - ManagedPreBufferUsesSingleCacheForMixedRowGroups: checks equivalent results with fewer bytes read for mixed scans and unchanged I/O for full-only and partial-only scans.
  - ManagedPreBufferBackwardSeekRefreshesCacheCoverage: checks cache coverage after backward seeks.
  - ManagedPreBufferPropagatesReadErrors: checks propagation of injected read failures.
  - DisabledPreBufferPreservesPageFilteredCache: checks page-filtered caching when automatic pre-buffering is disabled.

  Tests were not run while preparing this description. No integration tests were added.

  ### API and Format

  No changes to public APIs under include/, storage formats, or protocols.

  ### Documentation

  No new user-facing features or configuration options. No documentation changes required.

  ### Generative AI tooling

### DIFF
--- a/src/paimon/format/parquet/file_reader_wrapper.cpp
+++ b/src/paimon/format/parquet/file_reader_wrapper.cpp
@@ -87,7 +87,7 @@ std::vector<::arrow::io::ReadRange> MergeOverlappingRanges(
 
 Result<std::unique_ptr<FileReaderWrapper>> FileReaderWrapper::Create(
     std::unique_ptr<::parquet::arrow::FileReader>&& file_reader, int64_t batch_size,
-    std::shared_ptr<::arrow::MemoryPool> pool) {
+    std::shared_ptr<::arrow::MemoryPool> pool, bool pre_buffer_enabled) {
     try {
         if (file_reader == nullptr) {
             return Status::Invalid("file reader wrapper create failed. file reader is nullptr");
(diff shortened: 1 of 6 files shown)


gripleaf:
Effective remote retest: QPS 77.4735, E2E mean 50.855 ms.
Compared to PR #329: QPS +1.80%, mean improvement 2.00%.

The optimization review shows limited improvement.

PR #306, opened 2026-09-09
perf(predicate): evaluate comparison predicates with arrow::compute kernels

### LINKED ISSUE #307: [Feature] Evaluate comparison predicates with arrow::compute kernels
## Search before asking

## Motivation

Evaluating a comparison predicate (`=`, `<>`, `<`, `<=`, `>`, `>=`) on a batch goes through `NullFalseLeafBinaryFunction::Test`, which materializes the whole column into `Literal` objects — one heap allocation and one value hash per row — and then compares each row against the literal one at a time. That is `O(rows)` allocations plus `O(rows)` scalar comparisons per batch, even though a comparison only needs one vectorized pass over the column. On a wide scan with a selective filter this per-row `Literal` churn is a measurable part of predicate evaluation. This is the sibling of the `IN` / `NOT IN` optimization that already probes a batch with `arrow::compute::is_in`; the comparison functions are the other `LeafFunction` family still on the row-by-row path.

## Solution

Compare the whole batch against the literal with one `arrow::compute` comparison kernel (`equal`, `not_equal`, `less`, `less_equal`, `greater`, `greater_equal`) whenever the literal and the column allow it, which turns evaluation into `O(1)` setup — build one scalar per batch — plus an `O(rows)` vectorized compare with no `Literal` per row. Every `LeafFunction` is a shared stateless singleton, so the scalar cannot be cached on the function and is built per batch.

- Move `NullFalseLeafBinaryFunction::Test(array, literals, pool)` out of the header into a new `null_false_leaf_binary_function.cpp`.
- Build the comparison scalar with the existing `LiteralConverter::ConvertLiteralsToArray`, the same call `IN` writes its value set with, and probe the column with the kernel that matches the function type.
- Take the kernel path only where a kernel agrees with `Literal::CompareTo`: `BOOLEAN`, `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `DATE`, `STRING`, `BINARY`, and `DECIMAL` / `TIMESTAMP` whose column carries the scale / the unit of the literal and no time zone. `FLOAT` and `DOUBLE` keep the row-by-row path because `FieldsComparator::CompareFloatingPoint` orders `-0.0 < +0.0` and makes every NaN equal to every NaN, where an IEEE-754 kernel says `-0.0 == +0.0` and that no NaN compares to anything.
- Gate the kernel path on the layouts the row-by-row path already accepts, so which columns evaluate and which report an error stays exactly the same: the string pattern functions (`STARTS_WITH`, `ENDS_WITH`, `CONTAINS`, `LIKE`), a decimal column of another scale, a timestamp column of another unit or with a time zone, a literal whose field type disagrees with the column, and the layouts `ConvertLiteralsFromArray` rejects all keep the row-by-row path and its error reporting.
- As a side effect the kernel decodes a dictionary column, so a row that points at a null dictionary value is now false for every comparison instead of reading the empty value the slot holds, which an empty literal used to match.

## Anything else?

No public API, storage format, or protocol change; the work is internal to `src/paimon/common/predicate/`. The statistics and min/max pruning path keeps the existing per-`Literal` comparison, since it compares a single min/max pair rather than a batch.

lucasfang:
### Purpose

Linked issue: close #307 

`NullFalseLeafBinaryFunction::Test` evaluates a comparison predicate (`=`, `<>`, `<`, `<=`, `>`, `>=`) by materializing the whole column into `Literal` objects, one heap allocation and one value hash per row, and comparing each row against the literal one at a time. That is `O(rows)` allocations plus `O(rows)` scalar comparisons per batch, even though a comparison needs only one vectorized pass over the column. This is the sibling of the `IN` / `NOT IN` optimization that already probes a batch with `arrow::compute::is_in`; the comparison functions are the other `LeafFunction` family still on the row-by-row path.

This PR builds the literal into a one-element scalar once per batch and compares the whole column against it with the matching `arrow::compute` kernel (`equal`, `not_equal`, `less`, `less_equal`, `greater`, `greater_equal`), which turns evaluation into `O(1)` setup plus an `O(rows)` vectorized compare with no per-row `Literal`. Every `LeafFunction` is a shared stateless singleton, so the scalar cannot be cached on the function and is built per batch.

Changes:

- `NullFalseLeafBinaryFunction::Test(array, literals, pool)` moves out of the header into a new `null_false_leaf_binary_function.cpp`. The scalar is built with the existing `LiteralConverter::ConvertLiteralsToArray`, the same call `IN` writes its value set with.
- The kernel path is taken only where a kernel agrees with `Literal::CompareTo`: `BOOLEAN`, `TINYINT`, `SMALLINT`, `INT`, `BIGINT`, `DATE`, `STRING`, `BINARY`, and `DECIMAL` / `TIMESTAMP` whose column carries the scale / the unit of the literal and no time zone. `FLOAT` and `DOUBLE` keep the row-by-row path because `FieldsComparator::CompareFloatingPoint` orders `-0.0 < +0.0` and makes every NaN equal to every NaN, where an IEEE-754 kernel says `-0.0 == +0.0` and that no NaN compares to anything.
- The kernel path is gated on the layouts the row-by-row path already accepts, so which columns evaluate and which report an error stays exactly the same: the string pattern functions (`STARTS_WITH`, `ENDS_WITH`, `CONTAINS`, `LIKE`), a decimal column of another scale, a timestamp column of another unit or with a time zone, a literal whose field type disagrees with the column, and the layouts `ConvertLiteralsFromArray` rejects all keep the row-by-row path and its error reporting.
- One behavior correction falls out of decoding the dictionary in the kernel: a dictionary row that points at a null dictionary value is now false for every comparison, where the row-by-row path read the empty value the slot holds and an empty literal used to match it.

### Tests

New unit tests in `src/paimon/common/predicate/null_false_leaf_binary_function_test.cpp`, 21 cases, each asserting the per-row result over a batch:

- Kernel path per field type: `TestInt` and `TestString` across all six comparisons, `TestTinyIntSmallIntAndBigInt` with the int64 boundaries, `TestBoolean`, `TestDate`, `TestBinary` with embedded zero bytes, `TestDecimal`, `TestTimestamp`
- Fallback parity with the row-by-row path: `TestDecimalOffTheKernelPath` for another scale and the scale pair whose cast would fail, `TestTimestampOffTheKernelPath` for a coarser unit, a finer unit that would overflow int64 and a zoned column, `TestFloatAndDoubleStayOffTheKernelPath` for `-0.0`, `+0.0` and NaN
- Dictionary columns: `TestDictionaryString`, `TestLargeStringDictionary`, `TestDictionaryWithNullValue`
- Null, empty and error parity: `TestNullLiteralIsFalseForEveryRow`, `TestEmptyLiteralsReportTheError`, `TestMismatchedLiteralTypeReportsTheError`, `TestLayoutsLiteralConversionRejects`
- Non-comparison functions and edge layouts: `TestStringPatternFunctionsKeepTheRowByRowPath`, `TestSlicedArray`, `TestEmptyArray`

Verified with the full `paimon-common-test` suite: 1572 tests from 180 test suites, all passed.

### API and Format

No. No header under `include/` is touched, and neither the storage format nor the protocol changes. `NullFalseLeafBinaryFunction::Test` keeps its signature; only its definition moves from the header into a new `.cpp`.

### Documentation

No. This is a performance optimization plus the dictionary null-value correction above, with no user-visible API or configuration change.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/src/paimon/CMakeLists.txt
+++ b/src/paimon/CMakeLists.txt
@@ -128,6 +128,7 @@ set(PAIMON_COMMON_SRCS
     common/predicate/multi_literals_leaf_function.cpp
     common/predicate/not_equal.cpp
     common/predicate/not_in.cpp
+    common/predicate/null_false_leaf_binary_function.cpp
     common/predicate/or.cpp
     common/predicate/predicate.cpp
     common/predicate/predicate_builder.cpp
@@ -627,6 +628,7 @@ if(PAIMON_BUILD_TESTS)
                     common/predicate/literal_converter_test.cpp
                     common/predicate/literal_test.cpp
                     common/predicate/multi_literals_leaf_function_test.cpp
+                    common/predicate/null_false_leaf_binary_function_test.cpp
                     common/predicate/predicate_test.cpp
                     common/predicate/predicate_utils_test.cpp
                     common/predicate/predicate_validator_test.cpp
--- a/src/paimon/common/predicate/null_false_leaf_binary_function.cpp
+++ b/src/paimon/common/predicate/null_false_leaf_binary_function.cpp
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "paimon/common/predicate/null_false_leaf_binary_function.h"
+
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <vector>
+
+#include "arrow/array/array_base.h"
+#include "arrow/array/array_primitive.h"
+#include "arrow/compute/exec.h"
+#include "arrow/datum.h"
+#include "arrow/memory_pool.h"
+#include "arrow/result.h"
+#include "arrow/scalar.h"
+#include "arrow/type.h"
+#include "paimon/common/predicate/literal_converter.h"
+#include "paimon/common/utils/arrow/status_utils.h"
+#include "paimon/common/utils/checked_cast.h"
+#include "paimon/common/utils/field_type_utils.h"
+#include "paimon/defs.h"
+#include "paimon/status.h"
+
+namespace paimon {
+namespace {
+/// The `arrow::compute` kernel that compares a column against a scalar the way `Test(field,
+/// literal)` compares two `Literal`s, or `nullptr` for a function that is not a comparison.
+///
+/// `STARTS_WITH`, `ENDS_WITH`, `CONTAINS` and `LIKE` are `NullFalseLeafBinaryFunction`s too, but
+/// they match a pattern instead of ordering two values, so they are not here and keep the row by
+/// row path.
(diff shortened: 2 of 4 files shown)


lxy-9602 (approved):
+1

PR #309, opened 2026-09-09
perf(parquet): size Arrow builders by selected rows in page-filtered reads

### LINKED ISSUE #308: [Feature] Size Arrow builders by selected rows in page-filtered Parquet reads
## Search before asking

## Motivation

A page-filtered read resets every leaf through `PageFilteredRowGroupReader::ExecuteSkipReadPattern`, which calls `parquet::arrow::ColumnReader::ResetLeaf(col_idx, total)` where `total` is the leaf's whole-row-group size in post-page-filter compressed space, and that single number drives `RecordReader::Reserve(total)` for the leaf. When a selective predicate appends only a handful of rows out of a large row group, this mis-sizes the Arrow builders in two directions at once. For a variable-width (BYTE_ARRAY) leaf, `Reserve(total)` allocates the `::arrow::BinaryBuilder` offsets buffer for the entire row group — tens of MiB of offsets that are never written, because only the selected rows are appended. At the same time the builder's value-data buffer is not reserved at all: `Reserve()` deliberately leaves it alone for BYTE_ARRAY readers (`uses_values_` is false for them), so it grows by doubling once per decoded batch, and every doubling copies everything the row group has accumulated so far, which is quadratic on a multi-batch selective read.

## Solution

Split the single `reserve` into three numbers that describe what actually happens, threaded from the format layer down into the patched Arrow, so each leaf sizes its builders once for the selection instead of for the whole row group.

- `reserve_records` is the old `total`, the leaf's compressed-space size, and still bounds the rep/def levels that `SkipRecords` walks, so it stays at the full row-group size.
- `reserve_values` is `RowRanges::RowCount()` of the effective selection — the rows actually appended — and sizes the offsets buffer for the selection rather than the group.
- `reserve_value_bytes` drives a new `parquet::RecordReader::ReserveValueBytes(num_values, num_bytes)` hook that reserves the BinaryBuilder data buffer once; only the BYTE_ARRAY readers override it, and fixed-width readers keep the default no-op because `Reserve()` already sizes their values buffer.
- The byte estimate is `min(avg_width * reserve_values, total_uncompressed_size)` with `avg_width = column_chunk->total_uncompressed_size() / column_chunk->num_values()`. `total_uncompressed_size` also covers page headers, levels, and BYTE_ARRAY's 4-byte length prefixes, so `avg_width` over-estimates — the safe direction, since reserving too much costs one allocation while reserving too little costs a doubling copy of everything read so far — and the chunk size caps it, which no selection can exceed. It is guarded by `num_values() > 0 && reserve_values > 0`.
- `ReserveValueBytes` clamps the data reservation to `kBinaryMemoryLimit - value_data_length()`, mirroring `ArrowBinaryHelper::Prepare`, because `ArrowBinaryHelper` rolls a chunk at `kBinaryMemoryLimit` and `ReserveData()` would fail its overflow check on a request past it.

## Anything else?

No paimon public API (`include/paimon/`), storage format, or protocol change. The only API touched is paimon's own vendored Arrow patch (`cmake_modules/arrow.diff`): the internal `parquet::arrow::ColumnReader::ResetLeaf` signature and the new `parquet::RecordReader::ReserveValueBytes` virtual, neither of which is upstream Arrow's public API. The reservation is a capacity hint, so output is unchanged and the existing `page_filtered_row_group_reader_test.cpp` cases already cover the touched paths — notably `StringColumnMultiBatchStreaming`, a multi-batch BYTE_ARRAY selective read with per-value assertions, plus the nested struct/list/map cases that thread `ResetLeaf` through `StructReader` and `ListReader`.

lucasfang:
### Purpose

Linked issue: close #308 

A page-filtered read resets each leaf with `ColumnReader::ResetLeaf(col_idx, total)`, where `total` is the leaf's whole-row-group size in compressed space, and that one number drives `RecordReader::Reserve(total)`. On a selective predicate that appends a few rows out of a large row group this mis-sizes the Arrow builders both ways: a variable-width (BYTE_ARRAY) leaf reserves its `::arrow::BinaryBuilder` offsets buffer for the entire row group (tens of MiB never written), while the value-data buffer is not reserved at all because `Reserve()` leaves it alone for BYTE_ARRAY readers (`uses_values_` is false), so it doubles once per decoded batch and each doubling copies everything accumulated so far.

This PR splits the single `reserve` into `reserve_records` / `reserve_values` / `reserve_value_bytes` and threads them from the format layer into the patched Arrow, so each leaf sizes its builders once for the selection. Output is unchanged — the reservation is only a capacity hint.

Changes:

- `cmake_modules/arrow.diff`: widen `parquet::arrow::ColumnReader::ResetLeaf` from `(col_idx, reserve)` to `(col_idx, reserve_records, reserve_values, reserve_value_bytes)`, threaded through `LeafReader`, `ExtensionReader`, `ListReader`, and `StructReader`; `LeafReader::ResetLeaf` now calls `Reserve(reserve_records)` plus `ReserveValueBytes(reserve_values, reserve_value_bytes)`. Add the `parquet::RecordReader::ReserveValueBytes(num_values, num_bytes)` virtual (default no-op for fixed-width readers) and a `ByteArrayChunkedRecordReader` override that reserves the builder offsets and data buffer, clamped to `kBinaryMemoryLimit - value_data_length()` to mirror `ArrowBinaryHelper::Prepare`.
- `page_filtered_row_group_reader.h` / `.cpp`: `ExecuteSkipReadPattern` gains `reserve_values` / `reserve_value_bytes`; `ReadFilteredField` computes them per leaf — `reserve_values = effective_ranges.RowCount()` and `reserve_value_bytes = min(avg_width * reserve_values, total_uncompressed_size)` from the column chunk, an over-estimate capped by the chunk and guarded by `num_values() > 0 && reserve_values > 0` — and fetches `column_chunk` whenever the selection is non-empty, not only on the page-index path.

### Tests

No new test: the change preserves output and the existing `page_filtered_row_group_reader_test.cpp` already covers every path it touches — `StringColumnPageFilter` and `StringColumnMultiBatchStreaming` (multi-batch BYTE_ARRAY selective reads with per-value content assertions, the doubling case `ReserveValueBytes` removes), the nested struct/list/map cases that thread `ResetLeaf` through `StructReader` / `ListReader`, and the dictionary cases. A reservation-size assertion would be brittle and couple to implementation details.

Verified against a bundled Arrow re-patched with the updated `arrow.diff` (Release, `-j 96`):

- `paimon-parquet-format-test`: 220 tests from 16 suites passed, including all 49 `PageFilteredRowGroupReaderTest` cases.
- `paimon-common-test`: 1551 tests from 179 suites passed.

### API and Format

No. No header under `include/paimon/` is touched and neither the storage format nor the protocol changes. The only API affected is paimon's own vendored Arrow patch (`cmake_modules/arrow.diff`): the internal `parquet::arrow::ColumnReader::ResetLeaf` signature and the new `parquet::RecordReader::ReserveValueBytes` virtual, neither of which is upstream Arrow's public API.

### Documentation

No. This is an internal read-path performance optimization with no user-visible API, configuration, or behavior change.

### Generative AI tooling

Generated-by: Qoder

### DIFF
--- a/cmake_modules/arrow.diff
+++ b/cmake_modules/arrow.diff
@@ -12,17 +12,18 @@ index e7523add27..e079a1ad41 100644
        )
      endif()
 diff --git a/cpp/cmake_modules/ThirdpartyToolchain.cmake b/cpp/cmake_modules/ThirdpartyToolchain.cmake
-index 8cb3ec83f5..0765df8fa8 100644
+index 8cb3ec83f5..9aad83ac95 100644
 --- a/cpp/cmake_modules/ThirdpartyToolchain.cmake
 +++ b/cpp/cmake_modules/ThirdpartyToolchain.cmake
-@@ -814,5 +814,6 @@ if(DEFINED ENV{ARROW_THRIFT_URL})
+@@ -814,6 +814,7 @@ if(DEFINED ENV{ARROW_THRIFT_URL})
    set(THRIFT_SOURCE_URL "$ENV{ARROW_THRIFT_URL}")
  else()
    set_urls(THRIFT_SOURCE_URL
 +           "https://archive.apache.org/dist/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
             "https://www.apache.org/dyn/closer.cgi?action=download&filename=/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
             "https://downloads.apache.org/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
-@@ -983,6 +983,11 @@ if(CMAKE_TOOLCHAIN_FILE)
+            "https://apache.claz.org/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
+@@ -983,6 +984,11 @@ if(CMAKE_TOOLCHAIN_FILE)
    list(APPEND EP_COMMON_CMAKE_ARGS -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE})
  endif()
 
@@ -34,7 +35,7 @@ index 8cb3ec83f5..0765df8fa8 100644
  # and crosscompiling emulator (for try_run() )
  if(CMAKE_CROSSCOMPILING_EMULATOR)
    string(REPLACE ";" ${EP_LIST_SEPARATOR} EP_CMAKE_CROSSCOMPILING_EMULATOR
-@@ -1716,6 +1721,7 @@ macro(build_thrift)
+@@ -1716,6 +1722,7 @@ macro(build_thrift)
        -DWITH_JAVASCRIPT=OFF
        -DWITH_LIBEVENT=OFF
        -DWITH_NODEJS=OFF
@@ -56,7 +57,7 @@ index b36c38c6d4..f974a33073 100644
    /// \brief Return zero-copy string_view to upcoming bytes.
    ///
 diff --git a/cpp/src/arrow/util/bit_run_reader.h b/cpp/src/arrow/util/bit_run_reader.h
-index a436a503a0..27d483978c 100644
+index a436a50b86..27d483978c 100644
 --- a/cpp/src/arrow/util/bit_run_reader.h
 +++ b/cpp/src/arrow/util/bit_run_reader.h
 @@ -168,6 +168,26 @@ class ARROW_EXPORT BitRunReader {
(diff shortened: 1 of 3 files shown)


cmake_modules/arrow.diff
```diff
@@ -12,17 +12,18 @@ index e7523add27..e079a1ad41 100644
...
    set(THRIFT_SOURCE_URL "$ENV{ARROW_THRIFT_URL}")
  else()
    set_urls(THRIFT_SOURCE_URL
 +           "https://archive.apache.org/dist/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
             "https://www.apache.org/dyn/closer.cgi?action=download&filename=/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
             "https://downloads.apache.org/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
-@@ -983,6 +983,11 @@ if(CMAKE_TOOLCHAIN_FILE)
+            "https://apache.claz.org/thrift/${ARROW_THRIFT_BUILD_VERSION}/thrift-${ARROW_THRIFT_BUILD_VERSION}.tar.gz"
+@@ -983,6 +984,11 @@ if(CMAKE_TOOLCHAIN_FILE)
    list(APPEND EP_COMMON_CMAKE_ARGS -DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE})
```
lxy-9602, line 27: May remove?

lxy-9602 (approved):
+1

PR #312, opened 2026-09-09
ci: print core dump backtraces

### LINKED ISSUE #311: [Feature] Print native crash backtraces from core dumps in CI
### Search before asking

### Motivation

`build_support/run-test.sh` can use a core dump to print an all-thread backtrace after a native test crash, but the Build and Test workflow does not enable core dump generation or install GDB. LLVM sanitizers also disable core dumps by default on 64-bit platforms.

As a result, CI may only report `(core dumped)` without printing the stack that caused the crash, making intermittent native test failures difficult to diagnose from the job log.

### Solution

- Install GDB in the Linux build-and-test matrix.
- Set `kernel.core_pattern=core.%e.%p` and raise `RLIMIT_CORE` in the same shell that runs the tests.
- Configure ASan and TSan to enable core dumps and abort when reporting an error.
- Keep using GDB to print all-thread backtraces when `run-test.sh` finds a core, then delete the core during test cleanup.
- Remove the `Status::Abort()` death test so successful test runs do not intentionally create core dumps.

Core dumps and matching binaries are not retained or uploaded as artifacts.

### Anything else?

The implementation is available in #312. The behavior was verified once with a temporary program that called `abort()`; the [CI log](https://github.com/apache/paimon-cpp/actions/runs/34444475601/job/102766163233) contains the real GDB backtrace. The temporary crash test was removed after validation so regular CI runs do not intentionally generate a core dump.

zjw1111:
### Purpose

Linked issue: close #311

Enable core dump generation in Linux CI so native test crashes produce actionable backtraces in the job log.

- Install GDB in the build-and-test matrix.
- Configure `kernel.core_pattern` and raise `RLIMIT_CORE` in the same shell that runs the tests.
- Override the LLVM sanitizer defaults that disable core dumps, and make ASan/TSan abort when reporting an error.
- Use GDB to print all-thread backtraces when `run-test.sh` finds a core, then delete the core during test cleanup.
- Remove the `Status::Abort()` death test so a successful test run does not intentionally create core dumps.
- Do not retain or upload core dumps, test executables, or shared libraries.

### Tests

- `bash -n build_support/run-test.sh`
- `pre-commit run --files .github/workflows/build_and_test.yaml`
- `git diff --check`
- One-time real core dump validation: [Build and Test / script-tests](https://github.com/apache/paimon-cpp/actions/runs/34444475601/job/102766163233)

The one-time CI probe compiled a small program with debug symbols that called `abort()`. Real GDB reported `SIGABRT` and the crashing `main` source line, after which the core was deleted. The probe was removed after validation so normal CI runs do not intentionally crash.

### API and Format

No public API, storage format, or protocol changes. `Status::Abort()` remains available; only its death test is removed.

### Documentation

No user-facing documentation changes. This only changes CI crash diagnostics.

### Generative AI tooling


lxy-9602 (approved):
+1

PR #314, opened 2026-09-09
perf(parquet): add opt-in parsed offset index caching

### LINKED ISSUE #328: Reuse Parquet OffsetIndex objects and direct-plan decisions
### Problem

Arrow 17 deserializes OffsetIndex on every lookup, even when filesystem/index bytes are cached. Selective reads repeatedly request the same index during bitmap trimming, prebuffer planning and decoding. Rebuilding direct-read plans also repeats work.

### Improvement

Offer reader-local parsed OffsetIndex caching through `parquet.read.enable-offset-index-cache`, disabled by default to avoid retained-memory growth for wide tables. Reuse the per-leaf direct-plan decision. Preserve page selection, validation and fallbacks.

### Evidence

A local 100,000-row, one-column, 1-KiB-value benchmark with 6,250 pages reduced single-row read latency from 2.445 to 0.892 ms with equal bytes read. With 98 pages the improvement was 3.5%; full scans were unchanged. CPU samples confirm repeated OffsetIndex deserialization as the dominant cost. These results support selective many-page workloads, not a general wide-table speedup.

Implementation, reproduction and memory trade-offs: #314.

wangyong9999:
### Purpose

Closes #328.

Reuse parsed OffsetIndex objects across bitmap trimming, prebuffer planning and decoding within a Parquet reader. Enable with `parquet.read.enable-offset-index-cache=true`; **default: false**. The separate per-leaf direct-plan decision reuse avoids constructing the same plan twice.

Filesystem/read-ahead caches retain bytes, but Arrow 17 still calls `OffsetIndex::Make` on every lookup. Parsed-object reuse avoids that deserialization. It is useful for selective reads over many pages; it is not a general scan optimization.

Page selection, validation and decoding fallbacks are unchanged. Only accessed column indexes are memoized. Retained memory scales with row groups × accessed columns × pages × `sizeof(PageLocation)`, plus overhead and serialized buffers. For a 24-byte location, 1,024 × 10 × 1,000 needs about 234 MiB for locations alone. The 1,024-reader limit is not a byte cap; wide projections should leave the option off unless measured benefits justify the memory.

### Performance

Local x86_64/GCC 8 Release; 100,000 rows, one string column of deterministic 1-KiB values, one row group, Zstd, dictionary disabled. Fresh reader per iteration; warm local filesystem; single-row bitmap with `trim`, or full scan. Same binary/file, only the cache option changes. Five randomized interleaved repetitions, minimum 0.5 s each:

| Workload | Actual pages | Off, ms | On, ms | Result |
|---|---:|---:|---:|---|
| Point read, 16 KiB pages | 6,250 | 2.445 | 0.892 | 63.5% lower latency, 2.74× |
| Point read, 1 MiB pages | 98 | 0.831 | 0.802 | 3.5% lower latency |
| Full scan, 16 KiB pages | 6,250 | 120.169 | 120.493 | Essentially unchanged |

Medians; corresponding bytes read are identical (162,389 / 703,052 / 62,410,047). A second interleaved run produced 2.470 → 0.894 ms for the many-page point read. The benchmark checks row counts; option regression tests check values.

User-space `perf` sampling at 99 Hz with DWARF stacks, collected separately from timings, attributes 79.9% of off samples and 45.0% of on samples to stacks containing `OffsetIndex::Make` (1,195 / 1,010 samples; zero lost). The first parse remains necessary. This supports a CPU parsing benefit despite warm file bytes. It does not establish remote-storage, concurrent-service or thousand-column performance.

```bash
build/release/paimon-parquet-format-benchmark \
  --benchmark_filter=BM_ParquetRead_OffsetIndexCache \
  --benchmark_min_time=0.5s --benchmark_repetitions=5 \
  --benchmark_enable_random_interleaving=true
```

### Tests

Parquet **227/227**, read integration **302/302**, read-with-index **68/68**, full pre-commit and diff checks passed. Tests cover default/false/true/invalid options, correct selected values, parsed-object reuse and lifetime. A focused ASan build of the Parquet sources and benchmark passed all six benchmark cases; dependencies were not rebuilt with ASan.

Benchmark fixtures are explicitly released before filesystem singletons at exit, preventing temporary-directory cleanup from accessing a destroyed IOHook.

| Validation for 8f56331b | Status |
|---|---|
| [Fork build/test matrix](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488362) | All 12 x86_64/aarch64 compiler and sanitizer jobs plus script-tests passed |
| [Fork GCC 8](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488237) / [pre-commit](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488390) | Passed |
| Local Apache RAT | 0 unknown licenses |
| Upstream PR workflows | Awaiting maintainer approval (`action_required`); not executed |

### API and Format

One default-off read option. No public header or storage-format changes.

### Documentation

`parquet_metadata_cache.rst` documents configuration, scope and memory cost.

### Generative AI tooling

Generated-by: OpenAI Codex


src/paimon/format/parquet/parquet_input_stream.h
```diff
@@ -78,21 +80,22 @@ class ParquetInputStream : public ArrowInputStreamAdapter {
...
+            const int64_t offset = position - range->first;
+            is_index = offset <= range->second && nbytes <= range->second - offset;
         }
-        --range;
-        if (position - range->first > range->second ||
-            nbytes > range->second - (position - range->first)) {
+        if (!is_index && !cache_data_) {
             return ArrowInputStreamAdapter::ReadAt(position, nbytes, out);
         }
         auto key = CacheKey::ForKind(file_uri_, position, static_cast<int32_t>(nbytes),
```
wangyong9999, line 97: The key here is whatever range Arrow happened to request: with pre-buffer on that is the coalesced range for this projection and page selection, so the same bytes read under a different projection or selection get a new entry instead of a hit, and with pre-buffer off a whole column chunk becomes one entry. There is no cap on entry size, so a plain `LruCache` shared with the footer/index entries gets flushed by one large chunk read.

Main just merged #272, which puts a `FileBlockCache` one layer below this (fixed-size blocks, one fetch per block, concurrent readers wait on the same fetch); this PR is based before it. The only thing that cache lacks is reuse across reader lifetimes. Backing its blocks with the caller-provided `Cache` (uri + block index) would give that with bounded entries and hits on overlapping reads, instead of a second cache over the same bytes keyed by request shape. At minimum this needs a rebase onto #272 and a note on how the two layers interact.

src/paimon/format/parquet/parquet_input_stream.h
```diff
@@ -103,24 +106,89 @@ class ParquetInputStream : public ArrowInputStreamAdapter {
...
+            return arrow::Status::IOError("Invalid Parquet cached range value");
         }
         std::memcpy(out, value.value()->GetSegment().Data(), nbytes);
         return nbytes;
     }
 
+    arrow::Future<std::shared_ptr<arrow::Buffer>> ReadAsync(const arrow::io::IOContext& io_context,
+                                                            int64_t position,
+                                                            int64_t nbytes) override {
+        if (!cache_data_ || !cache_ || file_uri_.empty() || position < 0 || nbytes <= 0 ||
```
wangyong9999, line 126: This guard and the `is_index` lookup below are a copy of the ones in `ReadAt`. One helper returning the cache key (or nullopt when the read is not cacheable) would serve both.

src/paimon/format/parquet/parquet_input_stream.h
wangyong9999: Agreed on the request-shaped key and shared-budget concerns. I synchronized this branch with main (including #272) through a merge commit, preserving the review history, and withdrew the exact-range data-cache layer rather than just documenting those limitations. The option, added ReadAsync path, supporting tests/docs and accompanying local-filesystem change are removed. The retained diff only covers reader-local parsed-index reuse, sparse page selection and reuse of direct-plan decisions.

One qualification after tracing #272: its FileBlockCache is a single-block fallback for reads not covered by registered prefetch ranges. Cross-block/large reads decline, registered prefetch fetches bypass it, Read() waits on the future, and single-flight is instance-local. Shared backing is a sensible direction, but adding Cache alone would not cover all current Parquet/native async paths. File-end block alignment also means that a shared key needs unambiguous immutable content/range identity, and bounded blocks alone do not isolate aggregate data occupancy from metadata.

The revised description separates these future design requirements from the current implementation and removes the old cache/async combination as evidence for the retained diff. The PR remains draft.

src/paimon/format/parquet/parquet_input_stream.h
wangyong9999: The duplication was real. After the design review I removed the entire added ReadAsync data-cache path, so ParquetInputStream now matches synchronized main and these duplicate guards/index lookups no longer exist in the diff. I have not added a helper for the withdrawn layer. Any future shared-block implementation should centralize its eligibility/key construction at that layer.

src/paimon/format/parquet/parquet_input_stream.h
wangyong9999: Verified the final tree at 0bc1b4fb after preserving the concurrent review update. The exact-range layer and option are absent; only OffsetIndex memoization, page selection and direct-plan decisions remain. I also removed unused ColumnIndex memoization and avoided repeated binary searches for ranges within already visited pages. Parquet 226/226 and filesystem 64/64 pass, including the existing short-read regression. The revised description records the local main/candidate/candidate/main measurements, including regressions and the lack of a general speedup. Resolving the withdrawn-cache concern; cross-lifetime block backing remains explicitly outside this PR.

src/paimon/format/parquet/parquet_input_stream.h
wangyong9999: Confirmed at 0bc1b4fb: ParquetInputStream is identical to current main, including removal of the added ReadAsync override. The duplicate eligibility/key logic is gone, so no replacement helper is needed. Resolving this thread.

wangyong9999:
Follow-up review completed at 016a9747. I found no additional production defect in the retained changes, but closed two validation gaps: the 1,025-row-group test checks index retention/release and decoding after the cache limit; the 1,001-page test checks every returned value under sparse/dense selection, coalesce/trim and both pre-buffer settings. Trim deliberately includes gaps within a selected page, so its oracle differs from exact bitmap selection.

Results: Parquet 228/228, read integration 294/294, read-with-index integration 68/68; full pre-commit and diff checks passed. The production implementation is unchanged by this follow-up. Both original review threads remain resolved. Performance limitations and draft status remain unchanged; these additional tests are correctness evidence, not a new speedup claim.

src/paimon/format/parquet/page_filtered_row_group_reader.cpp
```diff
@@ -100,6 +100,38 @@ std::optional<DataPageLayout> GetDataPageLayout(
     return DataPageLayout{column_chunk_offset, first_data_page_offset};
 }
 
+// Call only after GetDataPageLayout has validated the ordered page row offsets.
+template <typename Visitor>
+void VisitSelectedPages(const RowRanges& row_ranges,
```
wangyong9999, line 105: The scan this replaces is one of three linear passes over the same page vector on this path: `GetDataPageLayout` validates every page location before either caller gets here, and `ComputeCompressedRowRanges` walks every page again afterwards — the description keeps both as linear. So the cursor plus binary search removes one pass out of three, and the benchmark table shows nothing outside noise.

What it costs is a helper that is only correct while the row ranges stay ascending and disjoint and `first_row_index` stays strictly increasing, an invariant carried in a comment that callers have to honour, plus the 121-line equivalence test needed to hold it in place.

The OffsetIndex memoization is worth keeping — `RowGroupPageIndexReaderImpl::GetOffsetIndex` calls `OffsetIndex::Make` on every lookup, so that one removes real work. I would drop the page-cursor rewrite until a profile shows the scan matters.

src/paimon/format/parquet/file_reader_wrapper_test.cpp
```diff
@@ -311,6 +312,99 @@ TEST_F(FileReaderWrapperTest, PredicateReadsOnlyItsPageIndexesAndKeepsPayloadRea
...
+        ASSERT_THROW(first->GetOffsetIndex(-1), ::parquet::ParquetException);
+        ASSERT_THROW(first->GetColumnIndex(100), ::parquet::ParquetException);
+        ASSERT_TRUE(first->GetOffsetIndex(0));
+    }
+    ASSERT_FALSE(released_index.expired());
+    reader.reset();
+    ASSERT_TRUE(released_index.expired());
+}
+
+TEST_F(FileReaderWrapperTest, ParsedOffsetIndexesRespectRowGroupRetentionLimit) {
```
wangyong9999, line 344: This writes 1,025 single-row row groups to re-assert `kMaxRowGroupPageIndexReaders`, a constant this PR does not change, and pins it through `weak_ptr` expiry, which is the retention map's internals rather than the memoization being added.

`MissingPageIndexesRemainReadable` below never runs its index assertions at all: with `enable_page_index=false` the row group has no index ranges, so `PageIndexReaderImpl::RowGroup()` returns nullptr and the whole `if (indexes)` body is skipped. What is left is a 2,000-row read that existing tests already cover.

`ReusesParsedOffsetIndexesWithinFileReader` covers what actually changed; I would drop these two.

lxy-9602:
Thanks for the contribution. Based on the PR description, it seems this PR only improves raw Parquet read performance by less than 10%. I'm not sure whether that gain is enough to justify the corresponding changes. Also, @lucasfang has recently been working on Parquet performance optimizations. Could you help evaluate this change in that broader context?

wangyong9999:
Thanks @lxy-9602 for your concerns.  Looking forward for @lucasfang's input on whether this improvement is necessary.

lucasfang:
Thanks for narrowing this down. Correctness looks sound across all three changes; my concern is benefit vs. complexity, and I'd suggest splitting rather than merging as-is.

## Correctness (verified)

- `VisitSelectedPages` invariants hold: `RowRanges` guarantees ascending, disjoint, inclusive ranges (`Add`/`Union`/`Intersection` all maintain sorted-disjoint), and `first_row_index` is strictly increasing. The `next_page = first` advance dedupes pages shared by several ranges, so each selected page is visited exactly once — which is what both `MakeDataPageReadPlan` (no duplicate entries) and `ComputePageRanges` (no duplicate ranges) need.
- `direct_read_columns` is semantically equivalent to the old `offset_index && MakeDataPageReadPlan(...)` gate. The factory runs during `GetColumn`, before the leaf loop, and `MakeDataPageReadPlan` is a pure function of inputs that are now identical (memoized `GetOffsetIndex` returns the same object), so dropping the second validation is safe.
- `CachedRowGroupPageIndexReader` caches `nullptr` for missing indexes correctly, and `GetOffsetIndex(-1)` throws before `emplace`, so the cache is not polluted.

## The part worth keeping

OffsetIndex memoization is the only change with a demonstrated CPU benefit. Within one file reader, `GetOffsetIndex` is called from several sites for the same (row group, column) — `column_index_filter`, the factory/leaf-loop/`ComputePageRanges` in `page_filtered_row_group_reader`, and `parquet_file_batch_reader` — all sharing the same cached wrapper via `GetRowGroupPageIndexReader`. Arrow 17 re-runs `OffsetIndex::Make` (O(pages)) on every lookup, so memoizing removes real repeated deserialization. The `direct_read_columns` reuse (removing a duplicate `MakeDataPageReadPlan`) is a low-risk companion to this.

## Concerns

- Performance is not established. The PR's own benchmark table is within noise, with a regression in the first 10% case, and the description states "No general throughput gain is established." These are timings, not CPU profiles, and the isolated high-page-count measurement is still outstanding.
- `VisitSelectedPages` has poor benefit/complexity. This path already makes three linear passes over the same page vector: `GetDataPageLayout` validates every page, `VisitSelectedPages` selects, and `ComputeCompressedRowRanges` walks every page again. The cursor+binary-search removes one of three passes, so the overall cost stays O(pages). The cost is a helper that is only correct while ranges stay ascending/disjoint and `first_row_index` strictly increasing — an invariant carried by a comment callers must honor — plus a 121-line equivalence test. I'd drop this until a profile shows the selection scan is a real hotspot.
- Memory: after memoization, each retained row-group reader holds parsed OffsetIndex objects, up to the 1,024-reader limit. As the doc notes, that is a count limit, not a byte budget; for many-row-group, many-page files the resident parsed objects can be significantly larger than the prior byte cache. Bounded, but worth an explicit magnitude note or a soft cap on very large page indexes.
- Two tests add little:
  - `ParsedOffsetIndexesRespectRowGroupRetentionLimit` writes 1,025 single-row row groups to re-assert `kMaxRowGroupPageIndexReaders`, a constant this PR does not change, and pins it via `weak_ptr` expiry — the retention map's internals, not the new memoization.
  - `MissingPageIndexesRemainReadable`: with `enable_page_index=false`, `RowGroup()` returns nullptr, so the entire `if (indexes)` assertion body is skipped, leaving a 2,000-row read already covered elsewhere.
  - `ReusesParsedOffsetIndexesWithinFileReader` is the one that covers the actual change; keep it.

## Suggested resolution

- Keep: OffsetIndex memoization (`CachedRowGroupPageIndexReader`), the `direct_read_columns` reuse, `ReusesParsedOffsetIndexesWithinFileReader`, and the doc update.
- Drop (or back with a high-page-count CPU profile): the `VisitSelectedPages` page-cursor rewrite and its 121-line equivalence test.
- Drop: `ParsedOffsetIndexesRespectRowGroupRetentionLimit` and `MissingPageIndexesRemainReadable`.
- If memoization is kept, add an explicit note (or a soft cap) on the 1,024 x parsed-index memory upper bound.

wangyong9999:
Thanks @lucasfang and @lxy-9602. Addressed in c7579e7c: the PR now contains only OffsetIndex memoization, direct-plan decision reuse, the focused reuse test and documentation. Page selection is identical to current main; the cursor rewrite and the three extra tests are gone.

The doc gives a concrete memory estimate: 1,024 row groups × 10 accessed columns × 1,000 pages × 24 bytes ≈ 234 MiB for parsed locations, plus serialized buffers and container overhead. The count limit is not a byte cap.

Updated against main aae05ccb. Local Parquet 226/226, read integration 302/302, read-with-index 68/68, full pre-commit and Apache RAT passed. The PR/issue describe only the final scope and do not claim measured throughput gains for it.

src/paimon/format/parquet/page_filtered_row_group_reader.cpp
wangyong9999: Removed in c7579e7c. MakeDataPageReadPlan and ComputePageRanges now match current main; only the per-leaf direct-plan decision is reused. The page-selection test additions are also removed.

src/paimon/format/parquet/file_reader_wrapper_test.cpp
wangyong9999: Removed both tests and the fixture parameter added for the retention-limit test in c7579e7c. ReusesParsedOffsetIndexesWithinFileReader is the sole new test. The existing Parquet and read integration suites pass.

lucasfang:
LGTM

lxy-9602:
In our use cases, files may contain thousands of columns, and the current offset index cache could consume a significant amount of memory. Could you please validate the performance benefit of this optimization, and whether the issue could instead be mitigated by adding caching within the filesystem layer? If not, please at least add a switch and keep it disabled by default.

wangyong9999:
@lxy-9602 @lucasfang — verified revision **8f56331b**.

### Reviewer concerns

| Concern | Resolution |
|---|---|
| Memory growth on wide tables | `parquet.read.enable-offset-index-cache` defaults to **false**. Disabled reads do not retain parsed indexes. When enabled, retention grows with accessed columns and pages; the documentation gives the formula and a 234 MiB example. There is no byte cap. |
| Can filesystem caching replace it? | No: Arrow 17 calls `OffsetIndex::Make` again even when serialized bytes are already buffered. This option reuses parsed objects within a reader. |
| Complexity versus benefit | Retained parsed-index caching and direct-plan decision reuse. Page selection and validation remain unchanged. Removed the cursor rewrite and redundant retention/missing-index tests. |
| Is the gain demonstrated? | Yes for the many-page selective case below; no general scan or thousand-column speedup is claimed. |

### Performance evidence

100,000 rows, one 1-KiB string column, one row group, Zstd, dictionary off. Warm local filesystem; fresh reader per iteration. Same binary/file, cache off versus on, five randomized interleaved repetitions. Median wall time:

| Workload | Pages | Off | On | Interpretation |
|---|---:|---:|---:|---|
| Single-row lookup, 16 KiB pages | 6,250 | 2.445 ms | 0.892 ms | **63.5% lower latency; 2.74×** |
| Single-row lookup, 1 MiB pages | 98 | 0.831 ms | 0.802 ms | 3.5% improvement |
| Full scan, 16 KiB pages | 6,250 | 120.169 ms | 120.493 ms | No meaningful benefit |

Corresponding storage-byte counts are identical. Separate 99 Hz CPU samples show stacks containing `OffsetIndex::Make` decreasing from **79.9% to 45.0%**. The benchmark and reproduction command are in the PR. These measurements support opt-in use for selective reads over many pages, not remote/concurrent workloads or wide projections.

### Correctness and validation

- Rechecked option propagation, default-off allocation, exception/null handling, index lifetime and direct-plan reuse with caching both off and on; no additional correctness defect found.
- Local suites: Parquet **227/227**, read integration **302/302**, read-with-index **68/68**. Coverage includes absent/false/true/invalid settings, selected values, object reuse and release.
- The six benchmark cases pass a focused ASan build. Fixture cleanup now runs before filesystem singleton destruction. Dependencies in that local probe were not instrumented.
- Current-revision Apache RAT: **0 unknown licenses**; full pre-commit and diff checks passed.

### CI — same revision

| Check | Result |
|---|---|
| [Fork Build and Test](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488362) | **Passed:** all 12 GCC/Clang, debug/release, ASan/UBSan and TSan jobs across x86_64/aarch64, plus script-tests |
| [Fork GCC 8](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488237) | **Passed** |
| [Fork pre-commit](https://github.com/wangyong9999/paimon-cpp/actions/runs/35121488390) | **Passed** |
| Upstream PR workflows | **Awaiting approval** (`action_required`), not executed |

Could a maintainer approve the upstream workflows for 8f56331b? The approval API returns HTTP 403 for my account. The passing fork checks above do not replace that approval.

src/paimon/format/parquet/file_reader_wrapper.h
```diff
@@ -60,7 +60,7 @@ class FileReaderWrapper {
 
     static Result<std::unique_ptr<FileReaderWrapper>> Create(
         std::unique_ptr<::parquet::arrow::FileReader>&& reader, int64_t batch_size,
-        std::shared_ptr<arrow::MemoryPool> pool);
+        std::shared_ptr<arrow::MemoryPool> pool, bool enable_offset_index_cache = false);
```
lxy-9602, line 64: I believe we may have mentioned similar comments in previous PRs as well. To keep the production code style consistent, could we avoid default arguments, and when adding non-output parameters, keep `pool` as the last parameter? Thanks!

wangyong9999: Fixed in 38acb1c2.

- Removed the default argument from `FileReaderWrapper::Create`.
- Moved `enable_offset_index_cache` before `pool` in the declaration, definition and all callers; `pool` is now last.
- The read option still defaults to false; runtime behavior is unchanged.

Validation: Parquet 227/227 passed; changed-file pre-commit and diff checks passed.

lxy-9602 (approved):
+1

PR #320, opened 2026-09-10
feat(aggregate): support product aggregate function

### LINKED ISSUE #317: [Feature] Support the product aggregate function
## Search before asking

## Motivation

`product` is one of the aggregate functions Java Paimon supports, through `FieldProductAgg` and `FieldProductAggFactory`, but `FieldAggregatorFactory::CreateFieldAggregator` has no branch for it. A primary-key table configured with `'merge-engine' = 'aggregation'` and `'fields.<field>.aggregate-function' = 'product'` therefore cannot be opened at all by paimon-cpp: `AggregateMergeFunction::Create` builds one aggregator per value field up front, so the whole merge function fails with `Invalid: Use unsupported aggregation product or spell aggregate function incorrectly!` before a single row is read. That blocks both merge reads and compaction of tables a Java writer already produces.

## Solution

Port `FieldProductAgg`, following the shape of the existing `FieldSumAgg`: resolve the per-type arithmetic once in `Create()` into a `std::function`, so the merge path does not switch on the field type per row. `Agg` multiplies the accumulator by the input and `Retract` divides it, both keeping Java's null handling — `Agg` returns whichever side is non-null, while `Retract` returns the accumulator unchanged when either side is null, so retracting into a null accumulator stays null rather than producing a reciprocal the way `sum` produces a negation.

The supported types are Java's `NUMERIC` family: TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE and DECIMAL. Anything else is rejected by `Create()`, matching the `checkArgument` in `FieldProductAggFactory`.

- Integer arithmetic is exact. `__builtin_mul_overflow` catches products that leave the field type, and division checks for a zero divisor and for `MIN / -1`; each returns `Status::Invalid` at the points where Java throws `ArithmeticException`. Every width computes in its own type, so `300 * 200` overflows SMALLINT instead of silently widening. TINYINT is read back through `int8_t` first, because the variant holds it as a plain `char` whose signedness follows the ABI.
- Decimal arithmetic runs through `arrow::BasicDecimal256`. A 128-bit intermediate is not enough: the unscaled product of two `DECIMAL(38, 18)` values reaches 1e76, well past `int128`. Multiplying doubles the scale, so the result is brought back down with `ReduceScaleBy(scale, round=true)`, which rounds away from zero on ties and therefore matches Java's `BigDecimal.setScale(scale, HALF_UP)`. Division cancels the scale out, so the dividend is scaled up first and the truncated quotient is then rounded half up from the remainder.

Java's two decimal edge cases are reproduced rather than smoothed over, so a table stays readable the same way from either engine:

- A result the field precision can no longer hold aggregates to null, because that is what `Decimal.fromBigDecimal` returns for it.
- Retraction rejects a quotient with no finite decimal expansion, because `BigDecimal.divide` without a rounding mode throws there. This is narrower than rejecting every inexact division: `1.00 / 8.00` terminates at `0.125` and Java still rounds it half up to `0.13`, while `2.00 / 3.00` repeats and is refused. The test is whether the divisor, stripped of its factors of 2 and 5, divides the dividend.

## Anything else?

No public API (`include/paimon/`), storage format, or option change. `product` is already a legal value of the existing `fields.<field>.aggregate-function` option, it simply had no implementation behind it. The change adds `field_product_agg.{h,cpp}` under `src/paimon/core/mergetree/compact/aggregate/`, registers the name in `FieldAggregatorFactory`, and wires both into `src/paimon/CMakeLists.txt`.

Cross-checked against Java: the expectations in `ProductAggregationITCase` reproduce exactly, including `DECIMAL(5, 3)` `1.01 * 1.10 * 10.00 = 11.110` and `DECIMAL(4, 2)` `1.01 * 1.10 * 10.00 = 11.10`, where the intermediate rounds half up; likewise the product cases in `FieldAggregatorTest` (`agg(null, 10) == 10`, `agg(1, 10) == 10`, `retract(10, 5) == 2`, `retract(null, 5) == null`, and the overflow cases for all four integer widths).

SteNicholas:
### Purpose

Linked issue: close #317

`product` is one of the aggregate functions Java Paimon supports, through `FieldProductAgg` and `FieldProductAggFactory`, but `FieldAggregatorFactory::CreateFieldAggregator` has no branch for it. A primary-key table configured with `'merge-engine' = 'aggregation'` and `'fields.<field-name>.aggregate-function' = 'product'` is therefore unusable: the aggregators for every value field are built as a group, so the whole merge function fails with `Invalid: Use unsupported aggregation product or spell aggregate function incorrectly!` as soon as the aggregation logic is initialized — eagerly in `FileStoreWrite::Create` on the write path, and on the first merge on the read path. That blocks writes, merge reads and compaction of tables a Java writer already produces.

This PR ports `FieldProductAgg`, following the shape of the existing `FieldSumAgg`: the per-type arithmetic is resolved once in `Create()` into a `std::function`, so the merge path does not switch on the field type per row. `Agg` multiplies the accumulator by the input and `Retract` divides it, both keeping Java's null handling — `Agg` returns whichever side is non-null, while `Retract` returns the accumulator unchanged when either side is null, so retracting into a null accumulator stays null rather than producing a reciprocal the way `sum` produces a negation.

Supported types are Java's `NUMERIC` family: TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE and DECIMAL. Anything else is rejected by `Create()`, matching the `checkArgument` in `FieldProductAggFactory`.

- Integer arithmetic is exact. `__builtin_mul_overflow` catches products that leave the field type, and division checks for a zero divisor and for `MIN / -1`; each returns `Status::Invalid` at the points where Java throws `ArithmeticException`. Every width computes in its own type, so `300 * 200` overflows SMALLINT instead of silently widening. TINYINT is read back through `int8_t` first, because the variant holds it as a plain `char` whose signedness follows the ABI.
- Decimal arithmetic runs through `arrow::BasicDecimal256`. A 128-bit intermediate is not enough: the unscaled product of two `DECIMAL(38, 18)` values reaches 1e76, well past `int128`. Multiplying doubles the scale, so the result is brought back down with `ReduceScaleBy(scale, round=true)`, which rounds away from zero on ties and therefore matches Java's `BigDecimal.setScale(scale, HALF_UP)`. Division cancels the scale out, so the dividend is scaled up first and the truncated quotient is then rounded half up from the remainder.

Java's two decimal edge cases are reproduced rather than smoothed over, so a table reads the same way from either engine:

- A result the field precision can no longer hold aggregates to null, because that is what `Decimal.fromBigDecimal` returns for it.
- Retraction rejects a quotient with no finite decimal expansion, because `BigDecimal.divide` without a rounding mode throws there. This is narrower than rejecting every inexact division: `1.00 / 8.00` terminates at `0.125` and is still rounded half up to `0.13`, while `2.00 / 3.00` repeats and is refused. The test is whether the divisor, stripped of its factors of 2 and 5, divides the dividend — so `3.00 / 6.00` is accepted and a zero dividend always is.

Changes:

- `field_product_agg.h` / `.cpp` (new): `FieldProductAgg`, plus the file-local helpers it needs — `ToDecimal256` / `ToInt128` to move between `paimon::Decimal` and `arrow::BasicDecimal256`, `MultiplyExact` / `DivideExact` for the checked integer arithmetic, and `QuotientTerminates` for the finite-expansion test.
- `field_aggregator_factory.h`: register `product` next to `sum`.
- `src/paimon/CMakeLists.txt`: add the source and the test to their targets.

Cross-checked against Java: the expectations in `ProductAggregationITCase` reproduce exactly, including `DECIMAL(5, 3)` `1.01 * 1.10 * 10.00 = 11.110` and `DECIMAL(4, 2)` `1.01 * 1.10 * 10.00 = 11.10` where the intermediate rounds half up, and the product cases in `FieldAggregatorTest` (`agg(null, 10) == 10`, `agg(1, 10) == 10`, `retract(10, 5) == 2`, `retract(null, 5) == null`, and the overflow cases for all four integer widths).

### Tests

New UT in `field_product_agg_test.cpp`:

- `TestSimple`, `TestNull`, `TestSupportedTypes` — multiply and divide over every accepted type, and the null combinations for both directions.
- `TestIntegerRetractTruncatesTowardsZero` — negative divisor, and `10 / 3` / `-10 / 3` truncating towards zero rather than rounding or flooring.
- `TestDecimalRoundsHalfUp` — ties, below-tie, negative operands, a negative divisor, and the case where the truncated quotient is zero so the rounding step must take its sign from the operands.
- `TestDecimalRejectsNonTerminatingQuotient` — `2.00 / 3.00`, `1.00 / 7.00`, `0.01 / 1.50` rejected; `3.00 / 6.00` and `0.00 / 3.00` accepted.
- `TestDecimalWithZeroScale` — `DECIMAL(10, 0)`, where the rescale steps are no-ops.
- `TestDecimalWiderThanInt128Intermediate` — `DECIMAL(38, 18)`, whose unscaled intermediate does not fit an int128.
- `TestIntegerArithmeticErrors` — overflow, `MIN / -1` and division by zero for all four integer widths.
- `TestDecimalOverflowsToNull`, `TestDecimalDivisionByZero`, `TestFloatingPointDivisionByZero` (signed infinities including a negative-zero divisor, and `NaN` for `0 / 0`), `TestInvalidType`.

Also `FieldAggregatorFactoryTest.TestSimple` gains a `product` case, and `AggregateMergeFunctionTest.TestProduct` covers a multi-row merge (`2 * 3 * 5`) plus a `DELETE` retraction through the merge function.

New IT in `write_and_read_inte_test.cpp`: `TestPKProductAggMergesAndSurvivesCompaction` writes three commits with overlapping keys over an INT and a `DECIMAL(10, 2)` column, verifies the merge-on-read result, then full-compacts and verifies the same result. The decimal values are chosen so every intermediate product is exact at scale 2, since decimal rounding is not associative and the merge order differs between merge-on-read and compaction.

### API and Format

No. No header under `include/paimon/` is touched, and neither the storage format nor the protocol changes. `product` is already a legal value of the existing `fields.<field-name>.aggregate-function` option; it simply had no implementation behind it.

### Documentation

Yes. `docs/source/user_guide/primary_key_table.rst` had no merge engine section at all, only a passing mention of "the user-specified merge engine" under LSM trees. This PR adds a **Merge Engines** section with an **Aggregation** subsection covering the options, the fallback order, the available function names and retraction handling, and a **product** subsection covering the accepted types, the null semantics, integer exactness and truncation, IEEE 754 behaviour for FLOAT and DOUBLE, and the two DECIMAL boundary behaviours above.

The other aggregate functions are listed by name only. Documenting their semantics was left out deliberately rather than written from their names, since this change did not verify them.

### Generative AI tooling

Generated-by: Claude Code (claude-opus-5)

### DIFF
--- a/docs/source/user_guide/primary_key_table.rst
+++ b/docs/source/user_guide/primary_key_table.rst
@@ -80,3 +80,75 @@ merged according to the user-specified merge engine and the timestamp of each re
 New records written into the LSM tree will be first buffered in memory. When the
 memory buffer is full, all records in memory will be sorted and flushed to disk.
 A new sorted run is now created.
+
+
+Merge Engines
+-------------
+When Paimon sink receives two or more records with the same primary key, it
+merges them into one record to keep the primary key unique. The
+``merge-engine`` table option decides how. Paimon C++ supports ``deduplicate``
+(the default, which keeps the last record), ``partial-update``, ``first-row``
+and ``aggregation``.
+
+Aggregation
+~~~~~~~~~~~~~~
+The ``aggregation`` merge engine aggregates each value field across the records
+sharing a primary key, using the function configured for that field. Fields
+without a configured function fall back to
+``fields.default-aggregate-function``, and then to ``last_non_null_value``.
+
+.. code-block:: text
+
+   merge-engine = aggregation
+   fields.<field-name>.aggregate-function = <function>
+   fields.default-aggregate-function = <function>
+   fields.<field-name>.ignore-retract = true
+
+The available functions are ``sum``, ``product``, ``min``, ``max``,
+``first_value``, ``last_value``, ``first_non_null_value``,
+``last_non_null_value``, ``bool_and``, ``bool_or``, ``listagg``, ``collect``,
+``merge_map``, ``nested_update``, ``hll_sketch`` and ``theta_sketch``. Each one
+accepts only the field types it is defined for. An unknown function, or a
+function configured on an unsupported field type, is rejected when aggregation
+logic is initialized. Read paths that bypass aggregation may skip this
+validation.
+
+Not every function can process a retraction, that is a record whose row kind is
+``DELETE`` or ``UPDATE_BEFORE``. A function that cannot returns an error, unless
+``fields.<field-name>.ignore-retract`` is set, which drops the retraction
+instead.
+
+product
+^^^^^^^^^^^^^^
+Multiplies the values of a field. It accepts ``TINYINT``, ``SMALLINT``,
+``INT``, ``BIGINT``, ``FLOAT``, ``DOUBLE`` and ``DECIMAL``, and it supports
+retraction by dividing the accumulated value.
+
+.. code-block:: text
+
+   merge-engine = aggregation
+   fields.price.aggregate-function = product
+
+Null values are skipped, so aggregating a null into a field leaves the field
+unchanged, and the first non-null value seeds the product. Retracting into a
+field that is still null leaves it null rather than producing a reciprocal.
+
+Integer arithmetic is exact. A product or a quotient outside the range of the
+field type, a division by zero, and dividing the smallest value of the type by
+``-1`` all fail with an error instead of wrapping around. An integer quotient
+truncates towards zero, so retracting ``3`` from ``10`` yields ``3``.
+
+``FLOAT`` and ``DOUBLE`` follow IEEE 754 instead of reporting an error.
+Retracting a zero from a finite non-zero value yields an infinity carrying the
+sign of both operands, and retracting a zero from a zero yields ``NaN``.
+
+``DECIMAL`` has two boundary behaviors worth knowing, both matching Paimon
+Java:
+
+- The product and the quotient are rounded half up to the scale of the field.
+  A result that no longer fits the precision of the field aggregates to
+  ``NULL`` rather than failing, so a ``DECIMAL(4, 2)`` field holding
+  ``99.99 * 99.99`` becomes ``NULL``.
+- Retraction only accepts a quotient with a finite decimal expansion, and
+  returns an error otherwise. ``1.00 / 8.00`` is accepted and rounds to
+  ``0.13``, while ``2.00 / 3.00`` is rejected.
--- a/src/paimon/CMakeLists.txt
+++ b/src/paimon/CMakeLists.txt
@@ -325,6 +325,7 @@ set(PAIMON_CORE_SRCS
     core/mergetree/compact/aggregate/field_collect_agg.cpp
     core/mergetree/compact/aggregate/field_merge_map_agg.cpp
     core/mergetree/compact/aggregate/field_nested_update_agg.cpp
+    core/mergetree/compact/aggregate/field_product_agg.cpp
     core/mergetree/compact/aggregate/field_sketch_agg.cpp
     core/mergetree/compact/aggregate/field_sum_agg.cpp
     core/mergetree/compact/interval_partition.cpp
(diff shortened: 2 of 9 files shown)


lxy-9602 (approved):
+1

lxy-9602:
Thank you for the contribution and for improving the documentation! 👍

PR #322, opened 2026-09-10
refactor(parquet): generalize read type adaptation for BLOB

### LINKED ISSUE #321: [Feature] Support BLOB values in top-level MAP fields
### Search before asking

### Motivation

The Parquet read path now handles both timestamp conversion and compatible BLOB representations,
so the name `ParquetTimestampConverter` and its timestamp-specific method names no longer describe
its responsibility accurately.

Parquet `ARROW:schema` may restore the physical `binary` column of a Paimon BLOB as
`large_binary`. Identifying this as a compatible BLOB representation requires the target field
metadata. The existing recursive compatibility check passes only `DataType` below the root STRUCT,
so that metadata is lost when traversing MAP values.

Paimon C++ also rejects every nested BLOB field, although a BLOB can be the direct value of a
top-level MAP field. The compatibility fixtures currently group `ARRAY<BLOB>` and
`MAP<K, BLOB>` together, preventing the supported MAP case from being tested independently.

The Rust fixture stores raw BLOB bytes inline as Parquet `binary`. Java cannot interpret those bytes
as a Paimon `BlobDescriptor`, so C++ should reject this non-standard representation instead of
adding a special conversion for it.

### Solution

- Rename `ParquetTimestampConverter` to `ParquetReadTypeAdapter` while retaining its static utility
  structure.
- Keep the existing timestamp conversion implementation.
- Retain Arrow `Field` metadata during recursive compatibility checks so BLOB can be identified
  inside MAP values.
- Treat `large_binary` restored through `ARROW:schema` as compatible with a logical Paimon BLOB
  field.
- Move the pre-conversion and post-conversion type checks into `AdaptArray`.
- Allow BLOB as the direct value of a top-level MAP field while keeping `ARRAY<BLOB>` unsupported.
- Split ARRAY BLOB and MAP BLOB fixtures.
- Verify Python and Java standard `MAP<K, BLOB>` data.
- Explicitly reject Rust raw-binary `MAP<K, BLOB>` data because it is not encoded as a Paimon
  `BlobDescriptor`.

### Anything else?

No public API, storage format, or protocol changes are proposed.

Implementation PR: #322

zjw1111:
### Purpose

Linked issue: close #321

Rename `ParquetTimestampConverter` to `ParquetReadTypeAdapter`, because the Parquet read path now
handles both timestamp conversion and compatible BLOB representations.

The adapter keeps the existing static utility structure and timestamp conversion logic. Its
recursive compatibility check retains child `Field` metadata so it can distinguish a normal binary
field from a Paimon BLOB when traversing MAP values. A `large_binary` value restored through
Parquet `ARROW:schema` is treated as compatible with the logical BLOB read type.

`AdaptArray` now owns the conversion check, recursive conversion, and final output type validation.

This change also:

- Allows BLOB as the direct value of a top-level MAP field.
- Keeps `ARRAY<BLOB>` unsupported.
- Splits ARRAY BLOB and MAP BLOB compatibility fixtures.
- Verifies Python and Java standard `MAP<K, BLOB>` files.
- Explicitly rejects the Rust fixture that stores raw BLOB bytes inline as Parquet `binary`, because
  it is not a valid Paimon `BlobDescriptor` representation.

### Tests

- `pre-commit run --all-files`
- `cmake --build build --target paimon-parquet-format-test paimon-paimon-read-compat-inte-test -j 64`
- `LSAN_OPTIONS=detect_leaks=0 ./build/debug/paimon-parquet-format-test --gtest_filter='ParquetReadTypeAdapterTest.*'`
- `LSAN_OPTIONS=detect_leaks=0 ./build/debug/paimon-paimon-read-compat-inte-test --gtest_filter='Writers/PaimonReadCompatInteTest.ReadsCompatibleTypeValues/*:Writers/PaimonReadCompatInteTest.ReadsBlobValues/*:Writers/PaimonReadCompatInteTest.ReadsVectorValues/*:UnsupportedTypes/PaimonUnsupportedTypeInteTest.ReportsExpectedError/*ArrayBlob:PaimonReadCompatInteStandaloneTest.RejectsNonStandardRustMapBlob'`
- `LSAN_OPTIONS=detect_leaks=0 ./build/debug/paimon-core-test --gtest_filter='ArrowSchemaValidatorTest.TestBlobFieldPlacement:SchemaValidationTest.TestMapSharedShreddingRejectsBlobValue'`

### API and Format

No public API, storage format, or protocol changes. The adapter is internal to the Parquet reader.
The checked-in fixtures cover existing writer output only.

### Documentation

Updated the compatibility fixture README to describe ARRAY/MAP BLOB support and the non-standard
Rust inline MAP BLOB representation.

### Generative AI tooling


lxy-9602 (approved):
+1

PR #323, opened 2026-09-10
perf(manifest): selectively decode bucket entries with bounded Avro indexes

gripleaf:
### Purpose

Bucket-specific scan planning can spend substantial CPU materializing manifest entries for unrelated buckets. The motivating workload has 90 partitions and approximately 16K buckets per partition.

This change reuses `LateMaterializingFileBatchReader` to probe the version and bucket fields, and the total bucket count for inferred-bucket reads. It materializes payloads for the selected rows. Different or unknown bucket counts remain available for per-entry bucket filtering. Unsupported versions are retained even when their bucket does not match, then rejected by the existing version validation before bucket filtering.

- Selective reads work with cached manifest bytes or the underlying file stream; they are not conditional on a manifest-byte cache hit.
- When manifest bounds identify only the requested bucket, the scan uses a single read.
- Avro bitmap selection supports the first read, preserves physical row IDs, and uses a sequential bitmap iterator.
- A block index accelerates subsequent selection, with a 65,536-block limit and sequential fallback above that limit. Schema resets discard incomplete indexes while preserving completed and disabled indexes.
- Manifest selectors use named field-index constants. Schema validation checks names, indices, and types; incompatible selector layouts fall back to ordinary reading and schema alignment.
- Bucket pruning does not compare historical bucket-key schemas or traverse all historical schema IDs. It assumes stable bucket-key hashing across schema versions, matching Java's bucket-pruning approach. Schema reads required for statistics evolution remain unchanged.

The CPU profiles already posted in the discussion are earlier measurements and have not been remeasured for this follow-up. No end-to-end latency improvement is claimed. Performance evaluation of uncached remote reads and high-selection-ratio mixed-bucket manifests remains outstanding.

### Tests

Validation of follow-up commit `e15dd5ad` on 2026-09-15:

- Debug build of `paimon-avro-format-test` and `paimon-core-test` with the configured `-Wall -Werror` flags: passed.
- Full Avro suite: 76 tests passed.
- Full core suite: 2,098 tests passed.
- Focused manifest suite: 12 tests passed, including reordered selector fields and unsupported versions.
- Selected scan/read integration tests: 10 tests passed, covering selective manifest decoding, schema evolution, and append bucket-key point lookups.
- Changed-file pre-commit checks and `git diff --check`: passed.

Regression coverage includes first-read bitmap selection, sparse and empty selections, out-of-range row IDs, compressed blocks, schema resets, empty files, index capacity fallback, retained disabled indexes, reordered manifest selector fields, cached and uncached reads, invalid versions, bucket pruning without historical compatibility reads, ordinary schema evolution, and Java 0.9/1.1 manifest compatibility.

The full integration suite was not run. No performance benchmark was run for this follow-up.

### API and Format

No changes to public headers, storage formats, or protocols. Avro implements the existing precise bitmap-selection contract. Legacy manifest schema alignment remains supported. Reads no longer provide a bucket-key schema compatibility fallback for historical changes to bucket-key hashing.

### Documentation

No new configuration options. Comments describe per-entry bucket-count filtering, the stable bucket-key hashing assumption, and block-index reuse.

### Generative AI tooling


gripleaf:
We evaluated this optimization using the same query load on a Paimon table with 90 partitions and approximately 16K buckets per partition. The workload performs point lookups using `WHERE rowkey = value`, with the bucket inferred from the predicate.

**Before**

Manifest reading still spends substantial CPU decoding Avro records into Arrow arrays before filtering candidates. `AvroDirectDecoder::DecodeAvroToBuilder` accounts for **58.65%** of sampled CPU, and `PrepareBucketRead` does not appear in the profile.

image


**After**

The profile now includes `PrepareBucketRead`: the first pass reads only the fields required to identify candidate rows, and the second pass fully decodes only those candidates. Entries with historical bucket counts or schema IDs remain eligible for the existing compatibility checks.

image

Under the same query load, estimated CPU consumption decreased substantially:

| Call path (inclusive CPU) | Before (CPU cores) | After (CPU cores) | Reduction |
|---|---:|---:|---:|
| `ManifestFile::ReadInferredBucketEntries` | 30.74 | 2.50 | 91.9% |
| `AvroFileBatchReader::NextBatch` | 27.83 | 2.31 | 91.7% |
| `AvroDirectDecoder::DecodeAvroToBuilder` | 21.92 | 1.16 | 94.7% |

These estimates are derived from the 99 Hz profiles and normalized by the available sampling duration: 39 seconds before and 40 seconds after. The rows represent overlapping call paths and should not be added together.

The results demonstrate a substantial reduction in CPU spent decoding manifest metadata under the tested workload. They measure CPU consumption, rather than end-to-end query latency or a corresponding query speedup.

src/paimon/format/avro/avro_file_batch_reader.cpp
```diff
@@ -18,23 +18,63 @@
...
+            if (checked_pointer_cast<arrow::MapType>(file_type)->keys_sorted() !=
+                checked_pointer_cast<arrow::MapType>(read_type)->keys_sorted()) {
+                return false;
+            }
+            break;
+        case arrow::Type::STRUCT:
+        case arrow::Type::LIST:
+            break;
+        default:
+            return file_type->Equals(read_type);
```
wangyong9999, line 78: This rejects nested types that Avro can't represent exactly. TINYINT/SMALLINT are written as Avro `int`, so the file type here is `int32` while the data-schema read type is `int8`/`int16`; the decoder already handles that pair through its builder-type dispatch. Before this PR a `ROW<a TINYINT>` column in an Avro data file read fine, now `SetReadSchema` fails with "Avro full struct read requires matching field names, order and types" (reproduced locally with `struct<r: struct<a: int8, b: int16>>` against a file written from the same type). Top-level columns aren't type-checked at all, so `ARRAY<TINYINT>` keeps working while `ROW<a TINYINT>` breaks.

Suggest comparing only names, order and nesting kind here (that's all positional decoding needs) and leaving leaf type compatibility to the decoder, as the top-level path already does.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -422,15 +422,27 @@ Status FileStoreScan::ReadAndMergeBucketFileEntries(
...
+                        meta.FileName(), bucket, core_options_.GetBucket(), table_schema_->Id(),
+                        &bucket_entries));
+                } else if (meta.MinBucket() && meta.MaxBucket() &&
+                           meta.MinBucket().value() == bucket &&
+                           meta.MaxBucket().value() == bucket) {
+                    // Every entry belongs to this bucket; a projection pass cannot prune rows.
+                    PAIMON_RETURN_NOT_OK(
+                        manifest_file_->Read(meta.FileName(), /*filter=*/nullptr, &bucket_entries));
+                } else {
+                    PAIMON_RETURN_NOT_OK(manifest_file_->ReadBucketEntries(meta.FileName(), bucket,
```
wangyong9999, line 442: Lazy decode is on by default, so every cached explicit-bucket read now takes the probe pass regardless of how many buckets the manifest spans. The profile you posted is the 16K-bucket case; for the usual few-bucket tables this is a second full parse plus a second zstd decompression of nearly every block, to save materializing only a fraction of the rows, which can easily be a net loss. `meta.MinBucket()/MaxBucket()` is already available here, so gating the probe on the bucket span would keep the win for the wide case without taxing the common one.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -106,9 +108,175 @@ Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t buc
...
+    }
+    auto probe_schema = arrow::schema(probe_fields);
+    ArrowSchema probe_c_schema;
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportSchema(*probe_schema, &probe_c_schema));
+    auto probe_status = reader->SetReadSchema(&probe_c_schema, /*predicate=*/nullptr, std::nullopt);
+    if (!probe_status.ok()) {
+        if (!inferred_layout || (!probe_status.IsInvalid() && !probe_status.IsNotImplemented())) {
+            return probe_status;
+        }
+        // Nested projection is optional. Restore a full single pass when unsupported.
```
wangyong9999: The only manifest reader that reports precise bitmap selection is Avro, and it now accepts exactly this `_FILE._SCHEMA_ID` projection, so this fallback never runs. Just return `probe_status`.

src/paimon/core/operation/file_store_scan.cpp
gripleaf: It's no need to cache when you only have few manifest files. So, a second full parse plus a second zstd decompression is not expensive.

lxy-9602:
Thanks for the contribution and the performance improvement. Could we refactor this before merging?

Java’s `ManifestAvroReader` only uses the top-level `bucket` and `totalBuckets` fields for filtering. I’m not sure why `_FILE._SCHEMA_ID` is needed here, since `bucket-key` is immutable. Schema ID also seems too broad as a bucket-layout identifier.

Could we reuse or extend the existing late-materialization framework instead of implementing another probe/bitmap/reread flow in `ManifestFile`? The selector is roughly:

```text
_BUCKET = target
OR _TOTAL_BUCKETS IS NULL
OR _TOTAL_BUCKETS != expectedTotalBuckets
```

`_VERSION = 2` may still be validated for every row.

This would also keep Manifest-specific logic out of the generic Avro reader.

lxy-9602:
> end-to-end query latency or a corresponding query speedup.

The flame graph suggests that Avro decoding time drops significantly after the optimization. I’m also curious whether the end-to-end query latency shows a similarly noticeable improvement, since it’s not entirely clear what fraction of the overall query latency is spent in Avro decoding. Thanks!

gripleaf:
> > end-to-end query latency or a corresponding query speedup.
> 
> The flame graph suggests that Avro decoding time drops significantly after the optimization. I’m also curious whether the end-to-end query latency shows a similarly noticeable improvement, since it’s not entirely clear what fraction of the overall query latency is spent in Avro decoding. Thanks!

Thanks for raising this. In our workload, the impact on query latency also comes from contention in the executor used for manifest scanning. Paimon C++ submits a read task for each manifest being scanned and waits for these tasks to complete before proceeding. CPU-intensive Avro decoding keeps the executor’s worker threads occupied, causing additional scan tasks to queue up under concurrent requests.

The large number of manifest files amplifies this problem: our previous count was over 7 million manifest entry in total. Under sustained load, requests accumulate and end-to-end latency progressively increases.

Reducing manifest decoding CPU should therefore help by freeing executor capacity and reducing queueing delays, in addition to making individual scans cheaper. The flame graphs demonstrate the CPU reduction, but we have not yet quantified the corresponding end-to-end latency improvement under the same load.

src/paimon/common/reader/late_materializing_file_batch_reader.cpp
```diff
@@ -45,28 +46,42 @@
...
+            PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<ArrowSchema> schema, inner_->GetFileSchema());
+            PAIMON_RETURN_NOT_OK(SetReadSchema(schema.get(), nullptr, std::nullopt));
+        } else {
+            state_ = kNoLatMat;
+        }
     }
     if (state_ == kProbing) {
         PAIMON_RETURN_NOT_OK(ReadAndFilterProbeData());
-        if (matched_bitmap_.IsEmpty()) {
+        if (!row_ids_fit_) {
```
wangyong9999: Two of the branches in this function have no caller. `PrepareBucketRead` always calls `SetReadSchema` before the first `NextBatch`, so the `kInit` validation path above only serves `ValidatesWithoutReadSchema`. And the `row_ids_fit_` fallback protects against a file with more than 2^32 rows, which the rest of the read path (deletion vectors, range bitmaps, `RoaringBitmap32` selections everywhere) cannot handle either; before this PR the same spot just truncated. Suggest dropping both, with their tests, and returning an error on overflow if you want it explicit. `kFiltering` then only remains for the no-payload case.

src/paimon/common/reader/late_materializing_file_batch_reader.h
```diff
@@ -38,12 +40,22 @@ class PredicateFilter;
 // For convenience, we abbreviate `Later Materializing` as `LatMat`.
 // This reader is installed below the prefetch layer (see
 // AbstractSplitRead::CreateFileBatchReader) and performs probe/payload two-phase reads when a
-// predicate is pushed down through SetReadSchema; without a predicate it is a plain passthrough.
+// predicate is pushed down through SetReadSchema. Without a predicate or validation it is a
+// plain passthrough.
 class LateMaterializingFileBatchReader : public PrefetchFileBatchReader {
  public:
+    struct ProbeValidation {
```
wangyong9999: Every row of a manifest file carries the same `_VERSION`: the serializer writes the constant for the whole file and files are immutable. So validating the rows you materialize already rejects an unsupported file whenever it contributes anything to the scan, and a file that contributes nothing cannot hurt. If lxy-9602 is fine with that, this hook, the `kFiltering` state and their five tests can go, and `PrepareBucketRead` shrinks to building the selector predicate.

gripleaf:
> Thanks for the contribution and the performance improvement. Could we refactor this before merging?
> 
> Java’s `ManifestAvroReader` only uses the top-level `bucket` and `totalBuckets` fields for filtering. I’m not sure why `_FILE._SCHEMA_ID` is needed here, since `bucket-key` is immutable. Schema ID also seems too broad as a bucket-layout identifier.
> 
> Could we reuse or extend the existing late-materialization framework instead of implementing another probe/bitmap/reread flow in `ManifestFile`? The selector is roughly:
> 
> ```
> _BUCKET = target
> OR _TOTAL_BUCKETS IS NULL
> OR _TOTAL_BUCKETS != expectedTotalBuckets
> ```
> 
> `_VERSION = 2` may still be validated for every row.
> 
> This would also keep Manifest-specific logic out of the generic Avro reader.

Thanks for the suggestions. I’ve refactored the PR to reuse LateMaterializingFileBatchReader.

  ManifestFile now builds the selector you proposed:

  _BUCKET = target
  OR _TOTAL_BUCKETS IS NULL
  OR _TOTAL_BUCKETS != expectedTotalBuckets

  The existing framework handles probing, bitmap construction, and the payload pass. The Avro changes provide generic bitmap selection, physical row-ID tracking, and a bounded block index; they contain no Manifest field names or bucket-specific logic. The two-pass path runs only when the manifest bytes are retained in memory, avoiding a second remote read.

  I also agree that schema ID is too broad to identify a bucket layout. _FILE._SCHEMA_ID is no longer part of the probe. However, unchanged bucket-key names do not necessarily imply unchanged hashing: the regression test covers an append table whose bucket-key type changes from INT to BIGINT, where the same value maps to different buckets despite an unchanged
  bucket count.

  To preserve the scan’s conservative behavior, inferred-bucket pruning now has a scan-level compatibility check. It uses the maximum schema ID from the already-loaded manifest metadata as an upper bound and checks historical schemas against the current bucket-key field IDs, order, types, and bucket function. Compatible histories retain selective decoding; an
  incompatibility or schema-read failure triggers fallback. This is deliberately conservative and reuses the existing per-scan caches. Cold schema checks can add schema-file reads.

  One detail about _VERSION: the current implementation validates retained entries through the existing serializer, rather than validating every row during probing. Unsupported versions in retained entries still fail.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +92,79 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
             return Status::OK();
+        },
+        [this, bucket, expected_total_buckets](std::unique_ptr<FileBatchReader>* reader) {
+            return PrepareBucketRead(reader, bucket, expected_total_buckets);
         });
 }
 
+Status ManifestFile::PrepareBucketRead(std::unique_ptr<FileBatchReader>* reader, int32_t bucket,
+                                       const std::optional<int32_t>& expected_total_buckets) const {
+    if (!(*reader)->SupportPreciseBitmapSelection()) {
```
lxy-9602, line 139: It looks like `reader` is an output parameter. Please place output parameters at the end of the function parameter list.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +92,79 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
+        return Status::OK();
+    }
+    PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<ArrowSchema> c_schema, (*reader)->GetFileSchema());
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Schema> file_schema,
+                                      arrow::ImportSchema(c_schema.get()));
+    const auto& target_type = serializer_->GetDataType();
+    const std::string& bucket_name = target_type->field(3)->name();
```
lxy-9602: There are too many hard-coded indices like `3` and `4` here. Please replace them with descriptively named constants instead.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +92,79 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
+    auto bucket_field = file_schema->GetFieldByName(bucket_name);
+    if (!bucket_field || bucket_field->type()->id() != arrow::Type::INT32) {
+        return Status::OK();
+    }
+    if (expected_total_buckets) {
+        auto total_field = file_schema->GetFieldByName("_TOTAL_BUCKETS");
+        if (!total_field || total_field->type()->id() != arrow::Type::INT32) {
+            return Status::OK();
+        }
+    }
```
lxy-9602: Instead of manually validating each field type, could we use `PredicateValidator::ValidatePredicateWithSchema()` here for consistency?

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +92,79 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
 }
 
 Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t bucket,
-                                       std::vector<ManifestEntry>* entries) const {
+                                       std::vector<ManifestEntry>* entries,
+                                       const std::optional<int32_t>& expected_total_buckets) const {
+    // Readers without an in-memory selective probe still filter aligned Arrow columns
+    // before constructing ManifestEntry and DataFileMeta objects.
```
lxy-9602, line 108: Similarly, `entries` is an output parameter, and output parameters should generally be placed at the end of the parameter list. Please don’t insert the new parameter after it.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +92,79 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
-            ColumnarRow row(fields, pool_, /*row_id=*/0);
-            for (int64_t i = 0; i < batch->length(); i++) {
+        [this, bucket, expected_total_buckets,
+         entries](const std::shared_ptr<arrow::StructArray>& batch) -> Status {
+            ColumnarRow row(batch->fields(), pool_, /*row_id=*/0);
+            for (int64_t i = 0; i < batch->length(); ++i) {
                 row.SetRowId(i);
-                PAIMON_RETURN_NOT_OK(ManifestEntrySerializer::ValidateVersion(row.GetInt(0)));
-                if (ManifestEntrySerializer::GetBucket(row) != bucket) {
+                // Different or unknown bucket counts must reach the compatibility checks.
```
lxy-9602: I’m a bit curious why `ValidateVersion` was removed. It seems like the result should still be version-validated even after late materialization.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -464,6 +482,24 @@ Status FileStoreScan::ReadAndMergeBucketFileEntries(
...
+        PAIMON_ASSIGN_OR_RAISE(bool compatible, HasCompatibleBucketKeys(schema_id));
+        if (!compatible) {
+            return false;
+        }
+        if (schema_id == max_schema_id) {
+            break;
+        }
+    }
+    return true;
+}
```
lxy-9602: The current implementation scans all schemas, which could have a significant performance impact. In real-world environments, the schema may change many times, and some of those schema versions may not even produce a snapshot. Also, Java’s `BucketFilter` only considers (partition, bucket, totalBuckets) and does not check the data schema ID.

src/paimon/core/utils/objects_file.h
```diff
@@ -201,6 +206,9 @@ Status ObjectsFile<T>::ReadArrowBatches(
 
     PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<FileBatchReader> batch_reader,
                            reader_builder_->Build(file_input_stream));
+    if (prepare_reader && cached_bytes) {
+        PAIMON_RETURN_NOT_OK(prepare_reader(&batch_reader));
+    }
```
lxy-9602, line 222: It seems that Java performs late materialization for partition and bucket regardless of whether caching is enabled. I’m not sure what the benefit is of tying late materialization to cache in the current implementation. Alternatively, please provide test results showing whether enabling late materialization without cache has any negative impact.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -422,15 +422,33 @@ Status FileStoreScan::ReadAndMergeBucketFileEntries(
...
-                    manifest_file_->ReadBucketEntries(meta.FileName(), bucket, &bucket_entries));
+                if (inferred_bucket) {
+                    PAIMON_RETURN_NOT_OK(manifest_file_->ReadBucketEntries(
+                        meta.FileName(), bucket, &bucket_entries, core_options_.GetBucket()));
+                } else if (meta.MinBucket() && meta.MaxBucket() &&
+                           meta.MinBucket().value() == bucket &&
+                           meta.MaxBucket().value() == bucket) {
+                    // Every entry belongs to this bucket; a projection pass cannot prune rows.
+                    PAIMON_RETURN_NOT_OK(
+                        manifest_file_->Read(meta.FileName(), /*filter=*/nullptr, &bucket_entries));
```
lxy-9602: I don’t fully understand the logic here. If `min == max == bucket`, it seems we wouldn’t need late filtering at all, and applying it would only increase I/O.

src/paimon/format/avro/avro_file_batch_reader.cpp
```diff
@@ -180,6 +235,24 @@ Status AvroFileBatchReader::SetReadSchema(::ArrowSchema* read_schema,
...
+        !block_index_.empty()) {
+        selection_iterator_ = selection_bitmap_->Begin();
+        selection_end_ = selection_bitmap_->End();
+        const uint64_t selected_row = static_cast<uint32_t>(**selection_iterator_);
+        auto block =
+            std::upper_bound(block_index_.begin(), block_index_.end(), selected_row,
+                             [](uint64_t row, const auto& entry) { return row < entry.first; });
+        selected_block_ = std::distance(block_index_.begin(), block) - 1;
+    }
+    previous_row_ids_.clear();
```
lxy-9602, line 289: The current support for the selected bitmap in `avro_file_batch_reader.cpp` does not seem very maintainable. Please refactor this part. I’m only listing a few of the issues here:

- The implementation is tightly coupled with the block-optimization-based late materialization logic, and appears to assume that the reader will first do a full read without a bitmap and then re-read with a bitmap.
- `bitmap.Contains()` is relatively expensive. If the access pattern is sequential, it would be better to use an iterator instead.
- This feature introduces too many combinations of member state, and the validity of those states seems to rely entirely on implicit conventions. In addition, ReadRowsIntoBuilder() appears to be taking on too many responsibilities.

src/paimon/core/operation/file_store_scan.cpp
gripleaf: Thanks for pointing this out. I agree that repeatedly loading historical schemas can be costly. My plan is to reduce this overhead through #329, which allows scans of the same table and branch to share a `SchemaManager` through `TableScanResources`. After warm-up, subsequent scans can reuse cached schema objects, avoiding repeated schema-file reads and JSON parsing.

Sharing the manager amortizes schema-loading costs across scans, but it does not eliminate the initial loading cost or the traversal of historical versions, including unused ones. I propose keeping this correctness safeguard here and using #329 to reduce the repeated I/O overhead.

gripleaf: One additional detail: the existing C++ scan path already checks bucket-key compatibility when filtering deserialized manifest entries, through FilterManifestEntry() → HasCompatibleBucketKeys().

src/paimon/format/avro/avro_file_batch_reader.cpp
gripleaf: Thanks for the review. I’ve refactored the bitmap support introduced by this PR:

  - Bitmap selection works on the first read, without requiring a preceding full scan. AdvanceToRow() skips records sequentially when no completed block index is available; an existing index only accelerates positioning.
  - SelectionCursor owns the bitmap and its iterators, replacing repeated Contains() checks with sequential iteration and keeping their lifetimes together.
  - BlockIndex encapsulates block positions and explicit Building / Ready / Disabled states. This separates index lifecycle management from the current selection and projection.
  - The batch loop stays directly in NextBatch(). Positioning, row preparation, and decoding/skipping are handled by AdvanceToRow(), PrepareNextRow(), and ReadCurrentRow(), respectively. The separate ReadRowsIntoBuilder() wrapper has been removed.

src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,25 +100,89 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
...
+                {selector, PredicateBuilder::IsNull(total_index, total_name, FieldType::INT),
+                 PredicateBuilder::NotEqual(total_index, total_name, FieldType::INT,
+                                            Literal(expected_total_buckets.value()))}));
+    }
+    // Retain unsupported versions regardless of bucket so the consumer validates every version
+    // before bucket filtering, including when the probe would otherwise select no entries.
+    const std::string& version_name = target_type->field(kVersionFieldIndex)->name();
+    const int32_t version_index = file_schema->GetFieldIndex(version_name);
+    PAIMON_ASSIGN_OR_RAISE(
+        selector,
```
lxy-9602, line 163: I thought the version field should already be at `kVersionFieldIndex`. Otherwise, reading the data back and directly accessing `kVersionFieldIndex` would also be problematic. Why does the current flow still need to remap the index here via `file_schema->GetFieldIndex`?

src/paimon/core/operation/file_store_scan.cpp
lxy-9602: Your target scenario seems to be an online service, where queries may share some resources such as `TableScanResources`. However, in many other scenarios, such as Spark, each query runs as an independent task and there is no resource sharing, so this assumption does not always hold. The `HasCompatibleBucketKeys` logic was originally added by @wangyong9999  — could you please confirm whether Java has similar logic? If not, this change would also introduce additional small I/O operations. More importantly, in the current PR, the code iterates and reads regardless of whether the schema actually has corresponding data files, which seems too costly.

Since the bucket key is declared immutable, it should not be changed through ALTER TABLE in principle. If the current Java implementation still allows such corner cases to slip through schema validation, I think we should strengthen schema validation to prevent them there, rather than adding a lot of fallback handling in the scan path.

src/paimon/core/manifest/manifest_file.cpp
lxy-9602: Manifest format changes now only append new fields at the end, so the it seems `FieldIndex` should remain unchanged.

src/paimon/format/avro/avro_file_batch_reader.cpp
```diff
@@ -145,16 +147,116 @@ Result<BatchReader::ReadBatch> AvroFileBatchReader::NextBatch() {
...
+        state_ = State::kReady;
+    }
+}
+
+void AvroFileBatchReader::BlockIndex::Reset() {
+    if (state_ != State::kReady) {
+        blocks_.clear();
+        state_ = State::kBuilding;
+    }
+}
```
lxy-9602, line 181: For immutable files with more than 64K blocks, a full read later will repeatedly rebuild the 64K block indexes and then disable them again. It seems we could reset only the unfinished `kBuilding` state, while keeping `kReady` and `kDisabled` unchanged.

lxy-9602:
Thank you for the multiple rounds of updates and refactoring. When you have time, could you please make another pass based on the latest comments so we can proceed with the next round of review?

src/paimon/core/operation/file_store_scan.cpp
gripleaf: Thanks for the clarification. I confirmed that Java’s bucket-pruning path does not perform an equivalent historical schema compatibility check. I also discussed this with @wangyong9999, and we agreed that HasCompatibleBucketKeys() was overly defensive and could be removed.

The latest commit removes both HasCompatibleBucketKeys() and the full historical schema traversal introduced by this PR, along with the compatibility cache and schema-ID-based fallback. Bucket pruning now follows Java’s approach and no longer reads historical schemas for bucket-key compatibility checks. The related tests have been updated accordingly.

gripleaf:
> Thank you for the multiple rounds of updates and refactoring. When you have time, could you please make another pass based on the latest comments so we can proceed with the next round of review?

Thanks for the follow-up and the detailed reviews. I’ve made another pass against the latest comments and pushed the updates in a95f8eb6.
 
 The latest changes:
 
 - Remove historical schema traversal and bucket-key compatibility checks, following the discussion.
 
- Use fixed, named manifest field indices with PredicateValidator, and preserve version validation before bucket filtering.
- Keep completed and disabled Avro block indexes across schema resets.

 The updates are ready for another review. Thanks again for your time!

lxy-9602 (approved):
+1

gripleaf:
hi @lxy-9602 Could you please rerun the affected CI jobs when you have a chance? clang-debug-aarch64 failed because the ORC format and Mosaic downloads timed out. gcc-debug-x86_64 completed the build and tests successfully but stalled at “Complete job” until the job timeout. Both appear to be infrastructure issues. Thanks!

PR #327, opened 2026-09-11
feat(realtime): support spilling sealed segments

lxy-9602:
### Purpose

Linked issue: #158 

Add opt-in disk spill support for the default real-time store to reduce memory retained by sealed segments.

- Spill each sealed segment to a local Arrow IPC file.
- Release in-memory batches after spill while keeping pinned views readable.
- Support spilled query and commit readers for append-only and PK tables.
- Delete spill files after their last reference is released.
- Reuse common Arrow IPC utilities in the existing merge-tree spill path.
- Support configurable read batch sizes for real-time queries by slicing large in-memory and spilled batches.

### Tests

Added unit and integration coverage for:

- Spill configuration and validation.
- Memory and spilled reads.
- Append-only and PK deletion-vector commit/read flows.
- Spill-file lifetime during pinned reads and reclaim.
- PK reads before and after reclaim.
- Read batch slicing.

### API and Format
Adds `realtime.spill-enabled` and spill context fields.

No persistent table format change. Spill files are temporary local Arrow IPC files.

### Generative AI tooling

### DIFF
--- a/src/paimon/core/mergetree/spill_writer.cpp
+++ b/src/paimon/core/mergetree/spill_writer.cpp
@@ -18,29 +18,15 @@
 
 #include "paimon/core/mergetree/spill_writer.h"
 
-#include "paimon/common/utils/arrow/arrow_output_stream_adapter.h"
-#include "paimon/common/utils/arrow/arrow_utils.h"
 #include "paimon/common/utils/arrow/mem_utils.h"
-#include "paimon/common/utils/arrow/status_utils.h"
 #include "paimon/common/utils/scope_guard.h"
+#include "paimon/core/io/arrow_ipc_file.h"
 #include "paimon/core/mergetree/spill_channel_manager.h"
+#include "paimon/macros.h"
 
 namespace paimon {
 
-SpillWriter::SpillWriter(const std::shared_ptr<FileSystem>& fs,
-                         const std::shared_ptr<arrow::Schema>& schema,
-                         const std::shared_ptr<FileIOChannel::Enumerator>& channel_enumerator,
-                         const std::shared_ptr<SpillChannelManager>& spill_channel_manager,
-                         const std::string& compression, int32_t compression_level,
-                         bool use_threads, const std::shared_ptr<MemoryPool>& pool)
-    : fs_(fs),
-      schema_(schema),
-      channel_enumerator_(channel_enumerator),
-      spill_channel_manager_(spill_channel_manager),
-      compression_(compression),
-      compression_level_(compression_level),
-      use_threads_(use_threads),
-      arrow_pool_(GetArrowPool(pool)) {}
+SpillWriter::~SpillWriter() = default;
 
 Result<std::unique_ptr<SpillWriter>> SpillWriter::Create(
     const std::shared_ptr<FileSystem>& fs, const std::shared_ptr<arrow::Schema>& schema,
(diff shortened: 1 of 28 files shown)


src/paimon/core/mergetree/spill_writer.cpp
```diff
@@ -18,105 +18,62 @@
...
     const std::shared_ptr<FileSystem>& fs, const std::shared_ptr<arrow::Schema>& schema,
     const std::shared_ptr<FileIOChannel::Enumerator>& channel_enumerator,
     const std::shared_ptr<SpillChannelManager>& spill_channel_manager,
     const std::string& compression, int32_t compression_level, bool use_threads,
     const std::shared_ptr<MemoryPool>& pool) {
-    std::unique_ptr<SpillWriter> writer(new SpillWriter(fs, schema, channel_enumerator,
-                                                        spill_channel_manager, compression,
-                                                        compression_level, use_threads, pool));
-    PAIMON_RETURN_NOT_OK(writer->Open());
+    std::unique_ptr<SpillWriter> writer(new SpillWriter());
```
zjw1111: I think this is a bit strange, calling the empty constructor first and then calling open

src/paimon/core/realtime/arrow_realtime_store.cpp
```diff
@@ -431,13 +731,58 @@ Result<std::vector<std::unique_ptr<BatchReader>>> ArrowRealtimeStore::CreateComm
...
-    for (const StoredBatch& batch : arrow_segment->GetBatches()) {
-        readers.push_back(std::make_unique<StoredBatchReader>(batch.data, arrow_pool_));
+    readers.reserve(spilled_segment->GetBatches().size());
+    // PK merge-on-read may consume returned readers concurrently. Give every batch reader an
+    // independent file handle and RecordBatchFileReader instead of serializing on a shared one.
+    for (int32_t i = 0; i < static_cast<int32_t>(spilled_segment->GetBatches().size()); ++i) {
+        PAIMON_ASSIGN_OR_RAISE(std::unique_ptr<ArrowIpcFileReader> file_reader,
+                               ArrowIpcFileReader::Open(spilled_segment->GetFile()->GetFileSystem(),
+                                                        spilled_segment->GetFile()->GetPath(),
+                                                        /*use_threads=*/false, arrow_pool_));
```
zjw1111: [P2] Bound the number of open PK spill file handles

This loop immediately opens and retains one file handle per stored batch, even though all batches belong to the same IPC file. After many small PK writes in one segment, `CreateCommitReaders` can exceed the process file-descriptor limit before returning any readers, causing `PrepareCommit` to fail with `Too many open files` even for a small amount of data. `CreateQueryReaders` has the same pattern at lines 871–884.

Please bound the open handles while preserving independent reader state, for example with internally synchronized shared random-access reads or bounded handle reuse. Add coverage with many small batches and a constrained file-descriptor limit; the new PK spill test only uses two batches.

zjw1111 (commented):
One resource-management issue in the PK spill read path.

src/paimon/core/realtime/arrow_realtime_store.h
```diff
@@ -82,16 +87,28 @@ class ArrowRealtimeStore final : public RealtimeStore {
...
     class StoredBatchReader;
+    class SpillBatchReader;
     class AppendQueryBatchReader;
 
     Result<std::optional<BatchStatistics>> CollectStatistics(
         const std::shared_ptr<arrow::StructArray>& data) const;
 
-    static Result<bool> MayMatchStatistics(const StoredBatch& stored,
+    static Result<bool> MayMatchStatistics(int64_t row_count,
+                                           const std::optional<BatchStatistics>& statistics,
```
zjw1111, line 112: statistics - diff name in .h and .cpp

src/paimon/core/realtime/arrow_realtime_store.cpp
```diff
@@ -95,18 +176,95 @@ class ArrowRealtimeStore::Segment : public RealtimeSegmentHandle {
...
+        }
+        return result;
+    }
+
     std::vector<StoredBatch> batches_;
-    int64_t row_count_ = 0;
+    uint64_t memory_usage_ = 0;
+};
+
+class ArrowRealtimeStore::SpillFile {
```
zjw1111, line 216: Have you considered splitting each class into a separate file? Right now, this file is packed with more than a dozen mixed classes.

lxy-9602: May address this in the next refactor PR.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #329, opened 2026-09-13
feat(scan): share table metadata resources across scans

gripleaf:
### Purpose

Linked issue: N/A

Repeated scans of the same managed table rebuild schema metadata and reload snapshot metadata. Add `TableScanResources` so callers can retain and share these resources across scans of one table and branch.

- Share schema and snapshot managers, Arrow schemas, partition schemas, and primary-key field metadata. Each new scan discovers the latest schema ID; loaded schema versions and derived metadata remain cached for the resource object's lifetime.
- Cache snapshots by full file path in a 20-entry LRU. Replace the entire cache on the first access after 30 minutes, on explicit invalidation, and around snapshot deletion. In-flight reads cannot repopulate the replacement cache.
- Keep snapshot discovery and existence checks current. Historical commit lookup, timestamp searches, and retained-snapshot publication checks read files directly. Failed metadata loads remain retryable.
- Validate the physical table path, branch, and file-system identity. The resources supply the default branch and file system; conflicting explicit settings are rejected.
- Forward resources through `$ro` and `$audit_log`. Keep caller-supplied schemas out of shared schema caches, and retain independent filters, snapshot selection, limits, and streaming progress for each scan.

### Tests

Regression tests cover metadata reuse across schema versions, schema updates, concurrent scans, read failures and retries, independent scan state, branch and system-table handling, file-system validation, builder reset, snapshot LRU eviction, expiration, and invalidation overlapping reads.

Validation completed in the existing Debug build:

- `cmake --build build-devcontainer --target paimon-core-test -j 8`
- `./build-devcontainer/debug/paimon-core-test`: 2,216 tests passed.
- Pre-commit checks for changed files and `git diff --check`: passed.

Lance and Tantivy are disabled in this build.

### API and Format

- Add `TableScanResources::Create(table_path, file_system, branch)` in `paimon/table/source/table_scan_resources.h`.
- Add `ScanContextBuilder::WithTableResources()` and `ScanContext::GetTableResources()`.
- Add a required final `table_resources` parameter to the public `ScanContext` constructor. Direct constructor callers must supply it, using `nullptr` when sharing is unnecessary; existing builder calls remain valid.

No storage-format or protocol changes.

### Documentation

Document resource creation and reuse, validation rules, cache lifetime, and supported table types in the scan API guide and public headers. Concurrent use requires a thread-safe file system. Callers must recreate resources after fast-forward, table or branch recreation, or file-system access-configuration changes. Cached metadata does not prevent snapshot or data-file expiration.

### Generative AI tooling


src/paimon/core/utils/snapshot_manager.cpp
```diff
@@ -124,7 +127,9 @@ bool SnapshotManager::ExpiredSinceBoundaryWasRead(int64_t id) const {
 }
 
 Result<Snapshot> SnapshotManager::LoadSnapshot(int64_t snapshot_id) const {
-    return Snapshot::FromPath(fs_, SnapshotPath(snapshot_id));
+    return snapshot_cache_.Get(snapshot_id, [this](const int64_t& id) -> Result<Snapshot> {
```
wangyong9999: Caching snapshots by ID here applies to every `SnapshotManager`, not only the ones held by `TableScanResources`, and several callers rely on `LoadSnapshot` failing with `NotExist` to learn that a snapshot is gone.

`LatestSnapshotOfUserAtOrBefore` walks IDs downward and reads `NotExist` as the end of history — or, when the latest came from the catalog, as the case that must first pass `ExpiredSinceBoundaryWasRead`. A cache hit now makes an expired snapshot look live and lets it be reported as that user's last commit, which is exactly the outcome the comment there warns about ("could cause a recovering writer to commit twice"). `FindSnapshotBeforeTimestamp`'s binary search has the same dependency.

The cache also gets primed with the wrong entries: `ExpireSnapshots` loads every snapshot it is about to delete through this same manager, so those IDs are hot in the cache the moment their files disappear. The bypass this PR adds in `ExpireSnapshots::ExpireUntil` ("Cached metadata cannot prove that the current file has been published") is the same problem, noticed at one call site.

I would keep the cache out of `SnapshotManager` and put it where this PR's scope is — behind `TableScanResources`, or on a separate accessor that existence-sensitive callers do not go through.

src/paimon/core/operation/scan_context.cpp
```diff
@@ -53,10 +56,19 @@ ScanContext::ScanContext(const std::string& path, bool is_streaming_mode,
       table_schema_(table_schema),
       options_(options),
       cache_(cache),
-      format_table_(format_table) {}
+      format_table_(format_table),
+      table_resources_(table_resources) {}
 
 ScanContext::~ScanContext() = default;
 
+std::shared_ptr<FileSystem> ScanContext::GetSpecificFileSystem() const {
```
wangyong9999: The four sibling context classes (`read_context.h`, `write_context.h`, `commit_context.h`, `orphan_files_cleaner.h`) all keep this accessor as "the file system the caller explicitly set". Folding the resources' file system in changes what the name answers, and it disarms the check it feeds: `TableScan::Create` passes `GetSpecificFileSystem()` into `resources.Validate(...)`, so when the caller set none it compares the resources' own file system against itself and the mismatch branch cannot fire there. `Finish()` is then the only place that really checks, and it has already rejected the conflicting case.

Keeping the accessor literal and resolving the effective file system at the two or three call sites that need it would leave both the name and the check meaning something.

lxy-9602:
I’d like to better understand two things here:

- What are the actual benchmark results for this optimization, and how does it behave across different data sizes?
- It seems that many kinds of caching could potentially be handled by the filesystem layer. If caching the resource here mainly reduces JSON deserialization, I’m curious what proportion of the total time is actually spent on that part.

gripleaf:
> I’d like to better understand two things here:
> 
> * What are the actual benchmark results for this optimization, and how does it behave across different data sizes?
> * It seems that many kinds of caching could potentially be handled by the filesystem layer. If caching the resource here mainly reduces JSON deserialization, I’m curious what proportion of the total time is actually spent on that part.

here is  the benchmark results under our specific scenario(point lookup + hdd hdfs), it has abort 6% improvement.

| metrics | PR323 | + PR329 | + PR304/pre-buffer |
| --- | ---: | ---: | ---: |
| QPS | 71.9205 | 76.1071 | 54.9883 |
| Mean | 55.169 ms | 51.893 ms | 71.985 ms |
| P50 | 51.447 ms | 42.065 ms | 68.598 ms |
| P95 | 73.872 ms | 83.655 ms | 83.911 ms |
| P99 | 112.001 ms | 103.840 ms | 119.431 ms |

lxy-9602:
Thanks for the reply. We’ve recently been working on some adjustments to the REST catalog e.g., #274 , and I think the issue in the current PR might be addressed by introducing a `Table` abstraction. The `Table` could hold in-memory state such as schema and snapshot information, and also serve as the entry point for creating scan/read/write/commit operations. We can discuss this further later.

PR #332, opened 2026-09-14
perf(fs): skip getFileStatus on Jindo open when the file length is known

### LINKED ISSUE #331: [Feature] Skip the getFileStatus round-trip when opening a Jindo/OSS file with a known length
## Search before asking

## Motivation

The read path opens a data file with a length it already knows: the manifest carries the file size, and both `DataFileReaderFactory::Open` and `PrefetchFileBatchReaderImpl` pass a `FileStatus(path, size)` into `FileSystem::Open`. But `JindoFileSystem` only implements `Open(const std::string& path)`, and the base `FileSystem::Open(const FileStatus&)` default just forwards to it, so the known length is dropped.

On the Jindo/OSS side, opening a reader without a length makes the store issue its own `getFileStatus` request to resolve the size before it can serve reads. That is one extra full-latency RPC to the object store on the critical path of every file open — per data file, per reader — even though the caller already had the answer. For scans over many small files this open-time round-trip is a measurable and entirely avoidable part of read latency.

The arrow-based `ObjectStoreFileSystem` already avoids this: its `Open(const FileStatus&)` validates the length and hands it straight to the input stream, skipping the `HeadObject` it would otherwise issue. The Jindo backend has no equivalent fast path.

## Solution

Plumb the trusted length from `FileStatus` down to the Jindo store so open can skip its own `getFileStatus`, mirroring what `ObjectStoreFileSystem` already does.

- `JindoFileSystem` overrides `Open(const FileStatus&)`, validates the length is non-negative, and routes both `Open` overloads through a new protected `OpenReader(path, std::optional<int64_t> file_length)` hook. `Open(path)` passes `std::nullopt` and keeps the original behavior.
- When a length is present, `OpenReader` calls a new vendored `JdoFileSystem::openReader(path, file_length, result)` overload that sets `JDO_OPEN_OPTS_HAS_GET_FILE_STATUS=true` and `JDO_OPEN_OPTS_FILE_LENGTH=<len>` on the open options. The `HAS_GET_FILE_STATUS` hint is the gate that actually suppresses the store's `getFileStatus`; `FILE_LENGTH` alone has no effect.

The benefit is automatic for the existing read path, since it already passes a `FileStatus` with the manifest size.

## Anything else?

- Scope: the hint is honored by the cloud object-store open path (`JdcOpenCall`); local and JFS opens ignore these options, so their behavior is unchanged.
- The length is trusted, not re-validated at open: a stale or wrong length surfaces later as a short or failed read rather than at open time. In the read path the length comes from the manifest for the file being read, so it is authoritative, and a negative length is rejected up front.
- No new public API and no storage format or protocol change: `FileSystem::Open(const FileStatus&)` already existed, this only implements it in the Jindo subclass; the new `openReader` overload lives in vendored `third_party/`.

lucasfang:
### Purpose

Linked issue: close #331 

The read path opens data files with a length it already knows — the manifest carries the file size, and both `DataFileReaderFactory::Open` and `PrefetchFileBatchReaderImpl` pass `FileStatus(path, size)` to `FileSystem::Open`. But `JindoFileSystem` only implemented `Open(path)`, and the base `Open(const FileStatus&)` default forwards to it, so the known length was dropped and the Jindo store issued its own `getFileStatus` RPC to resolve the size on every open — one extra full-latency round-trip to OSS per file open, per reader. The arrow-based `ObjectStoreFileSystem` already avoids this via its `Open(const FileStatus&)`; this brings the Jindo backend in line.

Changes:

- `src/paimon/fs/jindo/jindo_file_system.h` / `.cpp`: override `Open(const FileStatus&)`, validate the length is non-negative, and hand the trusted length straight to a new `JdoFileSystem::openReader` overload. `Open(path)` keeps the original behavior.
- `third_party/jindosdk-nextarch/include/JdoFileSystem.hpp` / `src/JdoFileSystem.cpp`: add an `openReader(path, file_length, result)` overload that sets `JDO_OPEN_OPTS_HAS_GET_FILE_STATUS=true` and `JDO_OPEN_OPTS_FILE_LENGTH=<len>` on the open options. The `HAS_GET_FILE_STATUS` hint is the gate that suppresses the store's `getFileStatus`; `FILE_LENGTH` supplies the trusted size.

The benefit is automatic for the existing read path, since it already passes a `FileStatus`.

### Tests

- UT `src/paimon/fs/jindo/jindo_file_system_unit_test.cpp`: `OpenWithFileStatusRejectsNegativeLengthBeforeOpening` runs the production `Open(const FileStatus&)` against an uninitialized store and asserts a negative length is rejected before any open happens.
- Network-gated `JindoFileSystemTest.TestOpenWithKnownFileSizeReadsBackContent` (built only with `-DPAIMON_ENABLE_NETWORK_TESTS=ON`): writes known content, opens it through `Open(FileStatus)` with the real length, reads every byte back and compares, and asserts that over-reading a stale shorter length fails with an EOF error instead of silently returning truncated data (verified against real OSS). This exercises the production open path end to end, which the offline unit tests cannot reach.
- `FsType/FileSystemTest.TestOpenWithKnownFileSize` now reads the content back through the trusted-length open instead of only checking `Length()`, so every instantiated file system proves the stream serves the real bytes rather than echoing the length the caller passed in.
- Actual results: `cmake --build build-debug-bench --target paimon-fs-test -j 96` built clean; the offline suites (`FileSystemStaticTest`, `FsType/FileSystemTest`, `JindoFileSystemUnitTest`, `JindoUtilsTest`, `LocalFileTest`, `ResolvingFileSystemTest`) → 119 ran, 114 passed, 5 skipped (pre-existing environment skips), 0 failed; the OSS network suites (`JindoFileSystemTest` including the new read-back case, `JindoFileSystemFactoryTest`, `JindoFileSystemAsyncReadTest`, `JindoFileSystemPaginationTest`) → 7 ran, 7 passed against the configured test bucket. `git diff --check` and `pre-commit run --files <changed files>` clean.

### API and Format

Public API: unchanged. `FileSystem::Open(const FileStatus&)` already existed in `include/paimon/fs/file_system.h`; this only implements it in the `JindoFileSystem` subclass and adds an internal protected `OpenReader` hook. The new `JdoFileSystem::openReader` overload lives in vendored `third_party/` and is not a public Paimon header. Storage format and protocol: unchanged — this only affects how a reader is opened, not what is read or written.

### Documentation

No new table option or user-facing configuration key. The behavior is an internal open-path optimization; the trusted-length contract is documented in code comments on `OpenReader` and the new `openReader` overload.

### Generative AI tooling

Generated-by: Qoder


src/paimon/fs/jindo/jindo_file_system_unit_test.cpp
```diff
@@ -82,7 +119,24 @@ class TestJindoFileSystem : public jindo::JindoFileSystem {
...
  protected:
+    Result<std::unique_ptr<InputStream>> OpenReader(
+        const std::string& path, std::optional<int64_t> file_length) const override {
+        calls_.push_back("open_reader");
+        open_reader_paths_.push_back(path);
+        open_reader_lengths_.push_back(file_length);
+        std::unique_ptr<InputStream> input = std::make_unique<TestInputStream>();
+        return input;
+    }
+
```
lxy-9602: I’m a bit curious about the reason for this test case. Since `OpenReader` seems to be overridden here, how do we verify that the production `OpenReader` implementation is actually correct?

lxy-9602 (approved):
+1

PR #334, opened 2026-09-14
perf(read): open data file streams and build readers concurrently

### LINKED ISSUE #333: [Feature] Open data file streams and build readers concurrently on the read path
## Search before asking

## Motivation

Opening a data file and building its reader are dominated by remote I/O waits — an open round trip, then a footer read — but the read path did both one file at a time. Two spots serialized those waits. `PrefetchFileBatchReaderImpl::Create` opened the read-ahead cache's stream first and paid its round trip before dispatching the readers' opens. `AbstractSplitRead::CreateRawFileReaders` built the readers of a split's data files sequentially, so a split holding N files paid N sequential open-plus-footer round trips before the first batch could be produced. These are I/O-bound waits rather than CPU work, so running them concurrently cuts the wall-clock latency to the first batch, most visibly on object stores with high per-request latency.

## Solution

Open all of a prefetch reader's streams in one concurrent wave, then build the readers in a second concurrent wave. The build has to be a separate wave: each build task wraps its own stream in the read-ahead cache, so folding it into the open wave would let a task block a worker of the same pool on the cache's still-queued open.

Build the readers of a split's data files concurrently on a dedicated, lazily created `reader_build_executor` pool of 4 threads. It is separate from the prefetch executor to avoid a nested deadlock, and a `thread_local` guard keeps any nested `CreateRawFileReaders` reached while building a reader serial so it never blocks a worker of that same pool. `CollectAll` preserves submit order, so readers keep the split's file order, and the first error fails the whole build.

`SchemaManager::ReadSchema` is already safe to call concurrently because its cache is a `ConcurrentHashMap`, so schema-evolution reads need no extra locking. The reader-build parallelism is a fixed 4 with no configuration knob.

## Anything else?

No change to the storage format or protocol, and no public API change under `include/paimon/`: the reader-build parallelism is internal and is not exposed as an option. The 4-thread reader-build pool is created once per process on first use, so the pool size does not vary per split.

lucasfang:
### Purpose

Linked issue: close #333 

Opening a data file and building its reader are dominated by remote I/O waits — an open round trip, then a footer read — but the read path did both one file at a time. `PrefetchFileBatchReaderImpl::Create` opened the read-ahead cache's stream first and paid its round trip before dispatching the readers' opens, and `AbstractSplitRead::CreateRawFileReaders` built the readers of a split's data files sequentially, so a split holding N files paid N sequential open-plus-footer round trips before the first batch. Since these are I/O-bound waits rather than CPU work, this PR runs them concurrently to cut the latency to the first batch, most visibly on object stores with high per-request latency.

A prefetch reader now opens all of its streams in one concurrent wave, then builds the readers in a second concurrent wave. The build has to be a separate wave: each build task wraps its own stream in the read-ahead cache, so folding it into the open wave would let a task block a worker of the same pool on the cache's still-queued open. Both waves run on the read context's executor, which is sized to the prefetch parallelism, so no pool is spun up and torn down per file and the open/build concurrency follows the configured parallelism instead of a fixed default; that reuse is safe because neither wave's tasks wait on one another or on that executor, and `Create` never runs on one of its workers. The readers of a split's data files are built concurrently on a dedicated, lazily created `reader_build_executor` pool of 4 threads, separate from the read context executor the inner waves use so the two never block on the same pool; a `thread_local` guard keeps any nested `CreateRawFileReaders` reached while building a reader serial so it never blocks a worker of that same pool. `CollectAll` preserves submit order, so readers keep the split's file order and the first error fails the whole build. `SchemaManager::ReadSchema` is already safe to call concurrently because its cache is a `ConcurrentHashMap`, so schema-evolution reads need no extra locking.

The reader-build parallelism is a fixed 4 with no configuration knob. The 4-thread pool is created once per process on first use, so its size does not vary per split; a split with fewer files builds only that many readers.

### Tests

New `PrefetchFileBatchReaderImplTest.TestCreateOpensAllStreamsConcurrently` uses a barrier file system that fails unless every stream (one for the cache plus one per reader) is in flight at once, pinning down that the cache's open is no longer serialized ahead of the readers'. `ScanAndReadInteTest.TestWithAppendSnapshot5WithReaderBuildParallelism` reads a snapshot-5 bucket holding five files through the concurrent build, and `ReadInteTest.TestAppendReadWithSchemaEvolutionWithReaderBuildParallelism` reads splits that mix two schema ids per split, exercising concurrent `SchemaManager::ReadSchema`.

Verified in a Debug build (`-j 96`):

- `paimon-core-test --gtest_filter='CoreOptionsTest.*'`: 39 tests passed.
- `paimon-scan-and-read-inte-test --gtest_filter='*TestWithAppendSnapshot5WithReaderBuildParallelism*:*TestWithAppendSnapshot5/*'`: 16 tests passed.
- `paimon-read-inte-test --gtest_filter='*SchemaEvolution*'`: 97 tests passed.
- `pre-commit run --files <changed files>` and `git diff --check`: clean.

### API and Format

No. `include/paimon/defs.h` is byte-identical to main — the previously drafted `read.reader-build.max-parallel-num` option and `CoreOptions::GetReaderBuildMaxParallelNum()` were dropped, and the parallelism is now an internal fixed 4 rather than a public option. No storage format or protocol change.

### Documentation

No. This is an internal read-path optimization with no user-visible API, configuration, or behavior change.

### Generative AI tooling

Generated-by: Qoder


test/inte/scan_and_read_inte_test.cpp
```diff
@@ -604,6 +604,32 @@ TEST_P(ScanAndReadInteTest, TestWithAppendSnapshot5) {
...
+
+    ReadContextBuilder read_context_builder(table_path);
+    AddReadOptionsForPrefetch(&read_context_builder);
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<ReadContext> read_context, read_context_builder.Finish());
+    ASSERT_OK_AND_ASSIGN(auto table_read, TableRead::Create(std::move(read_context)));
+    ASSERT_OK_AND_ASSIGN(auto batch_reader, table_read->CreateReader(splits));
+    ASSERT_OK_AND_ASSIGN(auto read_result,
+                         ReadResultCollector::CollectResult(std::move(batch_reader)));
+    ASSERT_TRUE(read_result);
+    ASSERT_GT(read_result->length(), 0);
```
lxy-9602: I don’t see how `ASSERT_GT(read_result->length(), 0);` matches the comment saying it “has to return the very same rows in the very same order as the other snapshot-5 read above.” At a minimum, shouldn’t we compare the actual query results?

test/inte/read_inte_test.cpp
```diff
@@ -2818,6 +2818,53 @@ TEST_P(ReadInteTest, TestAppendReadWithSchemaEvolution) {
...
+    context_builder.EnablePrefetch(param.enable_prefetch)
+        .AddOption("test.enable-adaptive-prefetch-strategy",
+                   param.enable_adaptive_prefetch_strategy);
+    ASSERT_OK_AND_ASSIGN(auto read_context, context_builder.Finish());
+    ASSERT_OK_AND_ASSIGN(auto table_read, TableRead::Create(std::move(read_context)));
+    ASSERT_OK_AND_ASSIGN(auto batch_reader, table_read->CreateReader(data_splits));
+    ASSERT_OK_AND_ASSIGN(auto result_array,
+                         ReadResultCollector::CollectResult(std::move(batch_reader)));
+    ASSERT_TRUE(result_array);
+    ASSERT_GT(result_array->length(), 0);
```
lxy-9602: Please compare the actual results to verify that the read is correct.

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -64,6 +68,29 @@ class FileStorePathFactory;
...
+    ~BuildingReaderGuard() {
+        building_reader = false;
+    }
+
+    BuildingReaderGuard(const BuildingReaderGuard&) = delete;
+    BuildingReaderGuard& operator=(const BuildingReaderGuard&) = delete;
+};
+
+}  // namespace
+
```
lxy-9602: It seems that `BuildingReaderGuard` was introduced to handle recursive `CreateRawFileReaders` calls, but I don’t clearly see where such recursion happens in the current code. Could you clarify why this `thread_local` variable is needed?

lxy-9602 (approved):
+1

PR #336, opened 2026-09-14
perf(read): skip the payload Concatenate when late materialization matches a single run

### LINKED ISSUE #335: [Feature] Unpack predicate boolean kernel bitmaps a byte at a time
## Search before asking

## Motivation

Two predicate leaf functions evaluate a batch by running an `arrow::compute` kernel and reading back the `arrow::BooleanArray` it writes: `MultiLiteralsLeafFunction` (`IN` / `NOT IN`, via `IsIn`) and `NullFalseLeafBinaryFunction` (the comparison functions). A kernel returns a bitmap, one bit per row, but `LeafFunction::Test` returns `std::vector<char>`, one byte per row, so both call sites spread the bits over bytes with the same per-row loop: test `IsNull`, read `Value`, apply the negation `NOT IN` needs, store a byte. That is a shift, a mask and a byte store per row, duplicated across the two call sites, on the selection path every filtered batch goes through.

## Solution

Extract the spread into one helper, `ArrowUtils::UnpackBooleansToBytes(array, negate)`, and read the bitmap a byte at a time instead of a bit at a time:

- A compile-time table maps each of the 256 bitmap bytes to the eight bytes it expands to, so the aligned body produces eight rows per iteration with one lookup and one 8-byte store.
- A scalar head and tail cover the rows sharing a partial leading or trailing byte, which is where the array offset is not byte-aligned; a batch a kernel has just written is aligned, so it takes the fast body throughout.
- The offset and the validity bitmap are honoured exactly as `BooleanArray::Value()` and `Array::IsValid()` honour them, and a null row unpacks to 0 whatever the value bitmap holds for it, which is what both `IN` / `NOT IN` and every `NullFalseLeafBinaryFunction` require.

`MultiLiteralsLeafFunction` passes its `negate` through; `NullFalseLeafBinaryFunction` passes `negate=false`. The bytes each returns are unchanged.

## Anything else?

A property test that asserts the helper equals a row-by-row reference through the very accessors it replaces, over every `(offset, length)` slice of a bitmap whose value and null periods are not multiples of eight and both negate values, pins the offset, validity and negate handling against the definitions it optimizes. No change to any header under `include/paimon/`, the storage format, or the protocol: `ArrowUtils` is an internal utility under `src/paimon/common/utils/arrow/`.

lucasfang:
### Purpose

Linked issue: close #335

The payload pass of late materialization compacts every batch down to the rows that passed the probe pass: `LateMaterializingFileBatchReader::ReadPayloadBatch` turns the matched bitmap into one zero-copy slice per contiguous run (`ReaderUtils::GenerateFilteredArrayVector`), then merges the runs with `arrow::Concatenate`. When the matched rows form a single run there is nothing to merge, so `Concatenate` only copies the payload bytes a second time to move the slice back to offset zero — and that rebase is redundant, because `AssembleFullBatch` already passes every column it takes through `ArrowUtils::NormalizeArrayOffsets`. A single run is the common shape here: the payload pass re-reads with the matched rows as its selection, so a precisely filtering reader returns fully matched batches, and a range predicate over clustered data keeps whole batches at a time.

This PR hands a lone slice straight to `AssembleFullBatch` and keeps `Concatenate` for two runs or more. Neither half of the single-run case is worse than before: a run starting at the batch head reaches the output with no copy at all (`NeedsNormalization` is false, so normalization is a no-op), and a run starting mid-batch is still copied exactly once, by the per-column normalization instead of by `Concatenate`. The output is unchanged — the assembled batch is still built from normalized, equal-length columns, so the exported `ArrowArray` keeps the offset-zero, children-match-parent shape consumers rely on.

Known trade-off: where the run stays zero-copy it keeps the inner batch's buffers alive until the consumer releases the assembled batch. That retains more than the matched rows only when the run starts at the head without reaching the end of the batch, and stays bounded by the read batch size times the prefetch queue depth.

### Tests

New `LateMaterializingFileBatchReaderTest.DictionaryPayloadColumnKeepsEncoding`: a dictionary-encoded payload column (`dictionary<int32, utf8>`, built with `StringDictionary32Builder` so the indices match the declared type) read under `k >= 5` with a fixed batch size, which reaches both shapes of a lone run — one at batch offset 0 that passes through untouched, one at batch offset 2 that the per-column normalization copies through `CopyToZeroOffset`, the one type whose normalization is a full copy rather than a buffer rebase. The test asserts the output stays dictionary-encoded and decodes to the expected values. Row-level correctness of the single-run and multi-run paths is already covered by the existing `ContiguousSubsetAcrossBatches`, `ScatteredAlternatingMatch`, `NestedPayloadColumn`, and `WorksAsInnerOfPrefetchReader` cases.

Verified in a Debug build (`-j 96`):

- `paimon-common-test --gtest_filter='LateMaterializingFileBatchReaderTest.*' --gtest_repeat=20`: 17 tests passed on every repetition (the mock reader randomizes batch sizes elsewhere in the suite, so repetition covers varying run shapes).
- `paimon-common-test`: 1620 of 1622 tests from 185 suites passed, 2 skipped (pre-existing skips in `prefetch_file_batch_reader_impl_test.cpp`).
- `paimon-core-test`: 2073 tests from 233 suites passed.
- `pre-commit run --files <the two changed files>` and `git diff --check`: clean.

### API and Format

No. No header under `include/paimon/` is touched, and neither the storage format nor the protocol changes. The reader's externally observable output is byte-for-byte the same batch.

### Documentation

No. This is an internal read-path optimization with no user-visible API, configuration, or behavior change.

### Generative AI tooling

Generated-by: Qoder


src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -432,6 +464,56 @@ uint64_t ArrowUtils::GetArrayMemoryUsage(const std::shared_ptr<arrow::ArrayData>
...
+        const uint8_t valid_byte = validity == nullptr ? uint8_t{0xFF} : validity[byte];
+        const uint8_t value_byte = values[byte];
+        const uint8_t bits = negate ? static_cast<uint8_t>(~value_byte & valid_byte)
+                                    : static_cast<uint8_t>(value_byte & valid_byte);
+        std::memcpy(out + row, kExpandedBytes[bits].bytes, sizeof(ExpandedByte));
+    }
+    for (; row < length; row++) {
+        out[row] = unpack_row(row);
+    }
+    return is_valid;
```
lxy-9602, line 508: Consider use `arrow::util::bit_util::bits_to_bytes`

src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -1215,4 +1217,124 @@ TEST(ArrowUtilsTest, TestFlattenUnresolvableDictionaries) {
...
+    arrow::bit_util::SetBitsTo(value_bits->mutable_data(), /*offset=*/0, length, /*bits=*/true);
+    const arrow::BooleanArray array(length, value_bits, /*null_bitmap=*/nullptr,
+                                    /*null_count=*/length);
+
+    ASSERT_EQ(ArrowUtils::UnpackBooleansToBytes(array, /*negate=*/false),
+              std::vector<char>(static_cast<size_t>(length), 0));
+    ASSERT_EQ(ArrowUtils::UnpackBooleansToBytes(array, /*negate=*/true),
+              std::vector<char>(static_cast<size_t>(length), 0));
+    ASSERT_EQ(UnpackRowByRow(array, /*negate=*/true),
+              std::vector<char>(static_cast<size_t>(length), 0));
```
lxy-9602, line 1331: Isn’t this an invalid array? Could you confirm that arrow `Validate` should already fail for this case (see `arrow/array/validate.cc`)? If so, testing an invalid array here does not seem very meaningful.

lxy-9602 (approved):
+1

PR #338, opened 2026-09-14
perf(read): reduce metadata round trips during scan planning

### LINKED ISSUE #337: [Feature] Cut metadata round trips when planning a scan (known manifest sizes, listing probes, parallel manifest-list reads)
## Search before asking

## Motivation

Scan planning pays several object-store round trips that are avoidable, either because the answer is already in metadata it has read or because one call can answer what two are asking.

- **Manifest reads re-resolve a length the metadata already carries.** A manifest list records each manifest's `fileSize`, and a snapshot records its base/delta/changelog manifest-list sizes. But `ObjectsFile::Read*` opened these files with a bare `Open(path)`, so on a remote store every manifest and manifest list paid a `getFileStatus`/`HeadObject` round trip just to learn a length planning already had. This is the metadata-path counterpart of the data-file fast path in `Open(const FileStatus&)`.
- **`ListVersionedFileStatus` probes existence before listing.** It called `Exists(dir)` then `ListDir(dir)`. Every file system already lists a missing directory as an empty result rather than an error, so the probe only decided whether to make a call that answers the same question — an extra round trip on every schema/snapshot/versioned-file lookup.
- **Jindo `ListDir` asks the store twice.** It called `Exists(dir)` then `GetFileStatus(dir)`; a single `GetFileStatus` answers both "is it there" and "is it a directory".
- **`ScanMode::ALL` reads the base and delta manifest lists serially.** They are two independent files and neither read depends on the other, so the two metadata round trips are paid one after the other instead of together.

For scans over many manifests, or against a high-latency object store, these round trips are a measurable and entirely avoidable part of planning latency.

## Solution

- Thread an optional known length through the metadata read path: `ObjectsFile::Read`/`ReadIfFileExist`/`ReadArrowBatches`/`ReadFileSegment` gain a `std::optional<int64_t> file_size` (default `std::nullopt`), and a new `OpenForRead` helper opens with `Open(FileStatus(path, size))` when the length is known and falls back to `Open(path)` when it is not. `ManifestFile::ReadBucketEntries` forwards it, `ManifestList::ReadBase/Delta/ChangelogManifests` pass the sizes recorded on the snapshot, and `FileStoreScan` / `SnapshotFileCollector` pass each `ManifestFileMeta::FileSize()`.
- Drop the `Exists()` probe in `FileUtils::ListVersionedFileStatus` and list directly.
- Collapse Jindo `ListDir` to a single `GetFileStatus`, mapping the SDK not-found to an empty listing (as the other file systems do) and propagating any other error.
- In `FileStoreScan::ReadManifestsWithSnapshot`, read the base and delta manifest lists concurrently through the existing `executor_` (`Via` + `CollectAll`), preserving base-then-delta order.

The size fields stay optional, so a snapshot or manifest list written before they existed keeps reading through the `Open(path)` fallback — this is an optimization, not a new requirement on the metadata.

## Anything else?

- The length handed to `Open(FileStatus)` is trusted, not re-validated, so a stale or wrong size would surface as a short or failed read. Manifest and manifest-list files are write-once and never rewritten, and the size is recorded by the same commit that wrote the file, so it cannot go stale underneath the read; a negative length is rejected by the base `Open(const FileStatus&)`.
- The size fast path benefits stores that override `Open(const FileStatus&)` (today `ObjectStoreFileSystem`; `ResolvingFileSystem` forwards it). `JindoFileSystem` currently only overrides `Open(path)`, so for Jindo the manifest-size threading is inert until it gains that override; the Jindo win here is the single-call `ListDir`.
- No storage format or protocol change, and no new public API: the changed signatures are internal `src/paimon` helpers and the new parameter is defaulted.

lucasfang:
### Purpose

Linked issue: close #337 

Scan planning paid several avoidable object-store round trips: it re-resolved file lengths the metadata already records, probed existence before listing directories that already list as empty when missing, asked Jindo for a directory's status twice, and read the base and delta manifest lists one after the other. This removes all four.

Changes:

- `src/paimon/core/utils/objects_file.h`: `Read` / `ReadIfFileExist` / `ReadArrowBatches` / `ReadFileSegment` take an optional known `file_size` (default `std::nullopt`), and a new `OpenForRead` helper opens with `Open(FileStatus(path, size))` when the length is known and falls back to `Open(path)` when it is not.
- `src/paimon/core/manifest/manifest_file.{h,cpp}`: `ReadBucketEntries` forwards the optional size.
- `src/paimon/core/manifest/manifest_list.h`: `ReadBase/Delta/ChangelogManifests` pass the manifest-list sizes recorded on the snapshot.
- `src/paimon/core/operation/file_store_scan.cpp` and `src/paimon/core/snapshot_file_scan.cpp`: pass each `ManifestFileMeta::FileSize()` to the manifest reads; in `ScanMode::ALL`, read the base and delta manifest lists concurrently through the existing `executor_` (`Via` + `CollectAll`), preserving the base-then-delta order the serial `ReadDataManifests` produced.
- `src/paimon/core/utils/file_utils.cpp`: drop the `Exists()` probe before `ListDir` in `ListVersionedFileStatus`.
- `src/paimon/fs/jindo/jindo_file_system.cpp`: collapse `ListDir`'s `Exists()` + `GetFileStatus()` into a single `GetFileStatus`, mapping not-found to an empty listing and propagating any other error.

The size fields stay optional, so metadata written before they existed keeps reading through the `Open(path)` fallback — this is an optimization, not a new requirement on the metadata.

### Tests

- New UTs: `ManifestFileTest.TestReadPassesKnownSizeToOpen`, `ManifestListTest.TestReadDataManifestsOpensWithSizeFromSnapshot` and `ManifestListTest.TestReadDataManifestsWithoutSizesStillReads`, `SnapshotFileScanTest.TestListFilesOpensManifestsWithKnownLength`, `FileUtilsTest.TestListVersionedFilesListsWithoutProbingExistence` and `FileUtilsTest.TestListVersionedFilesMissingDirListsWithoutProbingExistence`. They assert the known length reaches `Open(FileStatus)`, that unknown sizes fall back to `Open(path)`, and that a listing no longer probes existence — including for a missing directory.
- Actual results: `cmake --build build-debug-bench --target paimon-core-test -j 96` and `--target paimon-fs-test -j 96` both built clean. `paimon-core-test` → `ManifestFileTest` / `ManifestListTest` / `SnapshotFileScanTest` / `FileUtilsTest` 37/37 passed; the scan suites (`*FileStoreScan*`, `*ScanTest*`, `ManifestFileMergerTest`, `SnapshotManagerTest`) 99/99 passed. `paimon-fs-test` → Jindo and ListDir suites 14/14 passed. `git diff --check` clean.

### API and Format

Public API under `include/paimon/` is unchanged, and there is no storage format or protocol change. The signatures that gained a parameter (`ObjectsFile`, `ManifestFile::ReadBucketEntries`) are internal `src/paimon` helpers, and the new `file_size` parameter is defaulted, so existing callers are unaffected. The change relies on the existing `FileSystem::Open(const FileStatus&)` contract, where the caller-supplied length is trusted rather than re-validated; the manifest and manifest-list files it applies to are write-once, so a length recorded at planning time cannot go stale underneath the read.

### Documentation

No new table option or user-facing configuration key, and no new feature to document. The optimization and the trusted-length rationale are documented in code comments at `OpenForRead`, `ListVersionedFileStatus`, and Jindo `ListDir`.

### Generative AI tooling

Generated-by: Qoder


src/paimon/core/manifest/manifest_file.cpp
```diff
@@ -90,7 +91,8 @@ Result<std::unique_ptr<ManifestFile>> ManifestFile::Create(
 }
 
 Status ManifestFile::ReadBucketEntries(const std::string& file_name, int32_t bucket,
-                                       std::vector<ManifestEntry>* entries) const {
+                                       std::vector<ManifestEntry>* entries,
+                                       std::optional<int64_t> file_size) const {
     return ReadArrowBatches(
```
lxy-9602, line 96: Please place the output parameter at the end of the parameter list.

src/paimon/core/manifest/manifest_file.h
```diff
@@ -63,8 +64,12 @@ class ManifestFile : public ObjectsFile<ManifestEntry> {
     Result<std::vector<ManifestFileMeta>> Write(const std::vector<ManifestEntry>& entries);
 
     /// Read a manifest file and deserialize only entries for the specified bucket.
+    ///
+    /// @param file_size Length of the manifest when the caller already has it from the manifest
+    ///                  list, which saves the read a metadata request on a remote store.
     Status ReadBucketEntries(const std::string& file_name, int32_t bucket,
-                             std::vector<ManifestEntry>* entries) const;
+                             std::vector<ManifestEntry>* entries,
+                             std::optional<int64_t> file_size = std::nullopt) const;
```
lxy-9602, line 74: Please avoid using default arguments in production code.

src/paimon/core/operation/file_store_scan.cpp
```diff
@@ -276,8 +276,33 @@ Status FileStoreScan::ReadManifests(std::optional<Snapshot>* snapshot_ptr,
...
+                Via(executor_.get(), [read_list]() { return read_list(/*base=*/false); }));
+            for (auto& metas : CollectAll(futures)) {
+                if (!metas.ok()) {
+                    return metas.status();
+                }
+                for (auto& meta : metas.value()) {
+                    manifests->emplace_back(std::move(meta));
+                }
+            }
+            return Status::OK();
```
lxy-9602, line 304: There are only two files here in total, one base file and one delta file. Is multithreading really necessary in this case?

src/paimon/core/utils/objects_file.h
```diff
@@ -63,11 +64,15 @@ class ObjectsFile {
 
     virtual ~ObjectsFile() = default;
 
+    /// @param file_size Length of the file when planning already knows it, which lets the read
+    ///                  skip the metadata request a bare `Open` issues on a remote store. Leave it
+    ///                  unset when the length is not known; the read then discovers it itself.
     Status Read(const std::string& file_name, const std::function<Result<bool>(const T&)>& filter,
-                std::vector<T>* result) const;
+                std::vector<T>* result, std::optional<int64_t> file_size = std::nullopt) const;
     Status ReadIfFileExist(const std::string& file_name,
```
lxy-9602, line 73: Similarly, please avoid placing output parameters in the middle of the parameter list, and avoid using default arguments.

src/paimon/fs/jindo/jindo_file_system.cpp
```diff
@@ -173,11 +173,18 @@ Result<FileStatus> JindoFileSystem::GetFileStatus(const std::string& path) const
...
+    // One status call answers what Exists() followed by GetFileStatus() asked the store twice:
+    // whether the path is there at all, and whether it is a directory. PAIMON_RETURN_NOT_OK_FROM_
+    // JINDO maps the SDK's not-found to Status::NotExist, which is what tells a missing directory
+    // (listed as empty, as the other file systems do) from a call that genuinely failed.
+    Result<FileStatus> dir_status = GetFileStatus(directory);
+    if (!dir_status.ok()) {
+        if (dir_status.status().IsNotExist()) {
+            return Status::OK();
+        }
+        return dir_status.status();
```
lxy-9602, line 185: Could you confirm that Jindo’s `NotExist` status is correctly mapped to a Paimon status, and that Jindo actually returns a `NotExist` status when the target does not exist?

src/paimon/core/utils/file_utils_test.cpp
```diff
@@ -27,6 +27,30 @@
...
+
+    Status ListDir(const std::string& directory,
+                   std::vector<BasicFileStatus>* file_status_list) const override {
+        ++list_dir_count;
+        return LocalFileSystem::ListDir(directory, file_status_list);
+    }
+
+    mutable int exists_count = 0;
+    mutable int list_dir_count = 0;
+};
```
lxy-9602: I think these tests are somewhat redundant and too tightly coupled to the current implementation. Even small changes to the I/O path could break it. I would not recommend adding this kind of test just for the sake of increasing test coverage.

src/paimon/fs/jindo/jindo_file_system.cpp
lucasfang: confirmed on both halves. The mapping is covered by JindoUtilsTest.TestMacroMapsStatusByErrorCode, which asserts JDO_FILE_NOT_FOUND_ERROR becomes Status::NotExist (and other codes become IOError). For the SDK side there was no UT, so I added the network-gated JindoFileSystemTest.TestMissingPathReportsNotExistAndMissingDirListsEmpty: against real OSS it asserts GetFileStatus on a missing path fails with IsNotExist(), and ListDir on a missing directory returns OK with an empty list. It passes locally (Jindo suites 13/13).

lxy-9602 (approved):
+1

PR #340, opened 2026-09-14
fix: respect parent validity in Arrow nullability checks

### LINKED ISSUE #339: [Bug] Nullability validation incorrectly checks children masked by null parents
### Search before asking

### Paimon-cpp version

master

### Minimal reproduce step

`ArrowUtils::CheckNullabilityMatch` recursively checks complete Arrow child arrays.

For nested Arrow types, a null parent may still have physical child values. These
children are logically hidden and should not participate in nullability validation.
However, if a hidden child is null and its field is non-nullable, the current
implementation rejects the batch.

This affects nested STRUCT, LIST, MAP, and VECTOR fields and is inconsistent with
Java's logical row semantics.

Expected behavior:

- Ignore child values masked by a null parent.
- Reject null children when the parent is valid and the child is logically visible.

### Anything else?

_No response_

lxy-9602:
### Purpose

Linked issue: close #339 

`ArrowUtils::CheckNullabilityMatch` previously validated all physical child values.
This incorrectly rejected null values hidden by a null nested parent, although those
values are not logically visible. This behavior was also inconsistent with Java.

- Track visible ranges while validating STRUCT, LIST, MAP, and VECTOR fields.
- Ignore child values masked by a null parent.
- Continue rejecting visible nulls in non-nullable fields.
- Unify VECTOR element validation with nested nullability validation.

### Tests
ArrowUtilsTest, TestCheckNullableMatchParentNullMasksChildren
ArrowUtilsTest, TestCheckNullableMatchRejectsVisibleChildNulls

### Generative AI tooling


src/paimon/common/utils/arrow/arrow_utils_test.cpp
```diff
@@ -220,6 +220,57 @@ TEST(ArrowUtilsTest, TestCheckNullableMatchWithStruct) {
...
+    std::shared_ptr<arrow::Array> vector_batch =
+        arrow::ipc::internal::json::ArrayFromJSON(arrow::struct_({vector_field}),
+                                                  R"([[[1.0, null, 3.0]]])")
+            .ValueOrDie();
+    ASSERT_NOK_WITH_MSG(
+        ArrowUtils::CheckNullabilityMatch(arrow::schema({vector_field}), vector_batch),
+        "VECTOR field embedding is invalid: CheckNullabilityMatch failed, field item not nullable "
+        "while data have null value");
+}
+
```
dalingmeng, line 273: [Tests] Add parent-null coverage for LIST / MAP / sliced / deeply-nestedThe new TestCheckNullableMatchParentNullMasksChildren and TestCheckNullableMatchRejectsVisibleChildNulls only cover STRUCT and VECTOR(FSL). But the VisibleRanges fix also governs LIST/MAP, and the old InnerCheckNullabilityMatch recursed into the whole list_array->values() for LIST without excluding values hidden under a null parent row — one of the main false-positive paths. Suggest adding these regression cases to lock in the fix:
LIST: a nullable list row is null while its value_field is declared not-nullable, and the values in that parent-null row's range contain null → should pass (no false positive); a null in a visible row → should be rejected.
MAP: parent map row is null, key/item not-nullable with nulls at the corresponding positions → should pass.
Sliced array (offset ≠ 0): validate a sliced nested array to confirm value_offset mapping and IsValid indexing remain correct.
Deeply nested (STRUCT > LIST > STRUCT): multi-level visible-range narrowing, where a mid-level null parent masks deeper child nulls.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -90,6 +89,153 @@ bool HasUndeclaredDictionaryChild(const std::shared_ptr<arrow::DataType>& type,
...
+                "VECTOR field {} is invalid: VECTOR holds {} elements while {} rows of dimension "
+                "{} require {}",
+                field->name(), vector_array->values()->length(), vector_array->length(),
+                vector_length, required_values));
+        }
+
+        VisibleRanges value_ranges = GetVisibleValueRanges(*vector_array, valid_ranges);
+        // Paimon VECTOR values cannot contain null elements, irrespective of the Arrow child
+        // field's declared nullability.
+        std::shared_ptr<arrow::Field> value_field = vector_type->value_field()->WithNullable(false);
```
dalingmeng, line 221: Nit / non-blocking: I noticed this replaces the previous delegation to VectorUtils::ValidateVectorElements with an inline implementation. Two small things I wanted to flag for your consideration (not required — up to you):
Error message: the inline version uses the generic nullability message, whereas the original ValidateVectorElements reported VECTOR cannot contain null elements, found one at row {i} position {j}, which pinpoints the offending element. That's handy when tracking down bad data, so this is a slight regression in diagnostics.
Read vs. write paths: the read path ([vector_file_batch_reader.cpp:121,175](file:///home/menglingda.mld/workspace/apache/paimon-cpp/src/paimon/core/io/vector_file_batch_reader.cpp#L121)) still calls ValidateVectorElements, so the same VECTOR with a null element now surfaces different messages depending on whether it's being read or written. Functionally fine — just a little inconsistent on the diagnostics side.
If you'd like to converge these, one low-cost option would be to add an overload of ValidateVectorElements that takes the visible ranges: the write path passes the ranges through, the read path keeps using the existing overload, and both sides share one implementation with a consistent error message. That said, if you feel the current implementation is good enough, feel free to disregard this.

src/paimon/common/utils/arrow/arrow_utils.cpp
```diff
@@ -90,6 +89,153 @@ bool HasUndeclaredDictionaryChild(const std::shared_ptr<arrow::DataType>& type,
...
+    }
+    if (null_count == array.length()) {
+        if (!nullable) {
+            return NullabilityMismatch(field);
+        }
+        return VisibleRanges{};
+    }
+
+    VisibleRanges valid_ranges;
+    for (const VisibleRange& range : visible_ranges) {
```
dalingmeng, line 140: Non-blocking / optional (up to you):IntersectWithValidity scans visible ranges bit-by-bit via array.IsValid(i). Correct as-is — just a perf note: Arrow's arrow::internal::SetBitRunReader (arrow/util/bit_run_reader.h) scans the validity bitmap word-by-word and yields runs of set bits directly, skipping all-ones words instead of testing each bit. Since validity bitmaps are usually mostly-ones and this runs on the main write path, it could help on large arrays — and its output is exactly the "valid ranges" we're building here.

    VisibleRanges valid_ranges;
    const uint8_t* bitmap = array.null_bitmap_data();
    for (const VisibleRange& range : visible_ranges) {
        const int64_t base = array.offset() + range.offset;  // IsValid(i) uses logical index; bitmap bit is array.offset()+i
        arrow::internal::SetBitRunReader reader(bitmap, base, range.length);
        int64_t covered = 0;
        for (auto run = reader.NextRun(); !run.AtEnd(); run = reader.NextRun()) {
            AppendVisibleRange(range.offset + run.position, run.length, &valid_ranges);  // position is relative to base
            covered += run.length;
        }
        if (!nullable && covered != range.length) {  // a visible-range null; can't use global null_count (nulls may sit under a null parent)
            return NullabilityMismatch(field);
        }
    }
    return valid_ranges;

Note this would be the first use of bit_run_reader.h in the repo, so feel free to skip if you'd rather not introduce a new pattern for a micro-optimization.

dalingmeng:
+1

SteNicholas (approved):
LGTM.

PR #342, opened 2026-09-14
perf(prefetch): prefetch the late-materialization payload ranges mid-read

### LINKED ISSUE #341: [Feature] Prefetch the late-materialization payload ranges instead of reading them on demand
## Search before asking

## Motivation

Late materialization reads a data file in two passes: a probe pass over the predicate fields, then a payload pass over the remaining fields for the matched rows only. The shared read-ahead cache is fed once per read-range generation through `PrefetchFileBatchReader::PreBufferRange()`, before any read starts. At that point the payload pass cannot know which pages hold the matched rows — that depends on the probe result — so `PreBufferRange()` only reports the probe ranges (there was an explicit TODO for exactly this). The payload pass is therefore never prefetched: every payload read misses the cache and waits for its own underlying IO, serialized against the decode, on the pass that touches the wide columns.

## Solution

Let a reader report byte ranges that only become known after reading has started, and let the shared cache register them mid-read.

- New `ReadAheadCache::AddRanges(ranges, expected_round)` registers ranges into an already-initialized cache and is safe to call repeatedly and concurrently with `Read()`. It merges the new ranges into the disjoint, offset-ordered pending list, registering only the parts no registered range covers and dropping the overlap (the round that registered it is already fetching those bytes), then rebuilds the per-range cached flags so an already-fetched range is not fetched twice. The registered part is cut at a new `CacheConfig` knob `late_range_size_limit` (default 8 MiB, smaller than the 32 MiB `range_size_limit`) so a large pass is fetched by several concurrent requests rather than one long one; a new `Warmup(from_offset)` starts fetching from the first newly-registered range instead of from the head.
- A registration round bounds the lifetime: every `Init()` opens a round identified by `RegistrationRound()`, and `AddRanges()` drops everything when `expected_round` is not the open round, so a pass that outlived its generation — the cache was reset for a new read-range generation, or released by `Close()` — registers nothing instead of prefetching bytes nobody reads. The round counter is monotonic across `Reset()` so a stale round is never mistaken for a new one.
- New `PrefetchFileBatchReader::PreBufferSink` and `SetPreBufferSink()`: `PrefetchFileBatchReaderImpl` installs a sink on each sub-reader that tags the reported ranges with the current round, calls `AddRanges`, and warms up from the first new range. `LateMaterializingFileBatchReader` reports the payload ranges through the sink once the probe pass has refined the inner reader's target pages, and surfaces a failure to compute them (they come from the file metadata) rather than swallowing it.
- New metrics `read-ahead-cache.late.registered` / `.registered-bytes` / `.dropped` / `.dropped-bytes`, counted after coalescing and splitting, so `registered-bytes` and `dropped-bytes` together account for every reported byte.

## Anything else?

Adds public API under `include/paimon/`: `PrefetchFileBatchReader::PreBufferSink` / `SetPreBufferSink()` and `CacheConfig::GetLateRangeSizeLimit()` / `SetLateRangeSizeLimit()`. `ReadAheadCache::AddRanges` / `RegistrationRound` / `Warmup(offset)` and the new counter names live in the internal header. No storage format or protocol change.

lucasfang:
### Purpose

Linked issue: close #341

The shared read-ahead cache is fed once per read-range generation through `PrefetchFileBatchReader::PreBufferRange()`, before any read starts. For late materialization the payload pass cannot know which pages hold the matched rows until the probe pass has run, so `PreBufferRange()` only reported the probe ranges (an explicit TODO covered this) and the payload pass - the one touching the wide columns - was never prefetched: every payload read missed the cache and waited for its own underlying IO. This change lets a reader report byte ranges that only become known after reading has started, and lets the cache register them mid-read.

- New `PrefetchFileBatchReader::PreBufferRangeCallback` + `SetPreBufferRangeCallback()` (default no-op). `PrefetchFileBatchReaderImpl` installs a callback on each sub-reader; `LateMaterializingFileBatchReader` reports the payload ranges once the probe pass has refined the inner reader's target pages, and propagates a failure to compute them (they come from the file metadata) rather than swallowing it.
- `ReadAheadCache::AddRanges(std::vector<ByteRange>&&)` registers ranges into a live cache and is safe to call repeatedly and concurrently with `Read()`. It merges into the single offset-ordered, disjoint range list, registering only the parts no registered range covers and dropping the overlap (those bytes are already being fetched), and returns the first newly-registered offset (`nullopt` when everything was already covered). `Warmup(from_offset)` then fetches from there instead of from the head.
- Lifetime is bounded by the impl's `cache_warmed_` flag rather than a round counter: `CleanUp()`/`Close()` and the reader-reuse entry points (`SetReadSchema`/`RefreshReadRanges`) clear it, so a report that outlives its read-range generation registers nothing instead of prefetching bytes nobody reads. The callbacks are detached in `Close()` and in the destructor after the background thread has been joined.
- Cache simplification bundled in: `ReadAheadCache` dropped its pimpl and collapsed the three parallel containers (`pending_ranges_`/`is_cached_`/`entries_`) into one `ranges_` vector, where an entry is published exactly when its buffer is set; `PreBuffer()` takes the write lock only when the prefetch window still holds an unpublished range and otherwise returns under a read-lock fast path; both release paths swap the entries/blocks out under the lock and wait on the in-flight futures outside it; `FileBlockCache` moved to a `std::shared_mutex` with a read-lock hit path and a double-checked write-lock miss; `AtomicCounterPair::Add` was renamed to `Record`.
- `CacheConfig` default tuning: `range_size_limit` lowered 32 MiB -> 8 MiB (now the single cap for both the up-front and the mid-read ranges; one range is one prefetch IO, so a smaller cap spreads a pass over several concurrent requests instead of one long serial one), and `hole_size_limit` raised 8 KiB -> 512 KiB (coalesce across the page-sized gaps a filtered read leaves between the pages it keeps, trading bytes for fewer round trips on remote storage).
- Metrics: the late-registration counters were dropped together with the round concept. The cache still exposes `read-ahead-cache.read.count`/`.bytes`/`.hits`/`.hit-bytes`/`.misses`/`.miss-bytes`, `.io.count`/`.bytes` and `.block.fetches`/`.fetch-bytes`/`.hits`/`.hit-bytes` through `BatchReader::GetReaderMetrics()`.

Risk / known limits: the callback captures the impl raw, which is safe because it owns the sub-readers and `ClearPreBufferRangeCallbacks()` runs after the background thread is joined. `AddRanges()` concurrent with `Read()` is memory-safe (publish happens under the write lock, lookups under the read lock). Prefetch and block fetches issue concurrent `ReadAsync` on the same underlying stream, so that stream must be thread-safe, as before.

Bundled unrelated to the feature: `page_filtered_row_group_reader.cpp` changes the up-front value-byte reservation cap from an absolute 16 MiB to 8x the file-bounded compressed chunk size (forged-footer hardening), and `retry_waiter_test.cpp` is removed (timing-based).

### Tests

- UT `read_ahead_cache_test.cpp` (`TestReadAheadCache`): new `AddRanges` cases `TestAddRangesRegistersNewRanges`, `TestAddRangesClipsIntersectingRange` (only the uncovered part is registered, the overlap dropped), `TestAddRangesDropsCoveredRange`, `TestAddRangesInterleavesWithRegisteredRanges` (result stays offset-ordered and disjoint), `TestAddRangesSplitsLargeRange` (cut at `range_size_limit` into concurrent IOs), `TestAddRangesKeepsCachedRanges` (published entries are not fetched twice), `TestWarmupFromOffset` and `TestCoalescesSmallHoles` (`hole_size_limit`), alongside the existing no-eviction, metrics, IO-error-propagation, in-flight-racing-reader, pre-buffer-window and block-cache cases.
- UT `late_materializing_file_batch_reader_test.cpp` (5 new): `PayloadPreBufferRangesReportedToCallback`, `NoCallbackReportWithoutLateMaterialization`, `EmptyMatchDoesNotReportPayloadRanges`, `PayloadPreBufferRangeErrorFailsRead`, and `PrefetchInnerRegistersPayloadPreBufferRanges` (end to end under the prefetch layer: probe ranges registered up front plus payload ranges registered mid-read reach the shared cache and are fetched).
- IT `read_inte_test.cpp::TestReadAheadCacheMetrics`: asserts the read/hit/miss/io counters are exposed and `hits > 0` exactly when a cache is created, absent otherwise.
- Test infra: `MockFileBatchReader` gained per-schema pre-buffer ranges and an injectable `PreBufferRange()` status; `MockInputStream::ReadAsync` completes inline so a cache-attached promise is always resolved instead of hanging reads/`ReleaseBuffers()`. The adaptive-sizer cases in `byte_range_combiner_test.cpp` were removed together with the adaptive sizer.

Validation run (release build): `paimon-common-test` 1637 passed / 2 skipped; `TestReadAheadCache.*` + `*LateMaterializing*` 48/48; `paimon-read-inte-test` `*ReadAheadCache*` 8/8; `paimon-parquet-format-test` `*Cache*:*PreBuffer*:*ReadHints*` 10/10; `git diff --check` clean.

### API and Format

New public API surface; no storage format or protocol changes:

- `include/paimon/reader/prefetch_file_batch_reader.h`: the `PreBufferRangeCallback` type alias and `virtual void SetPreBufferRangeCallback(PreBufferRangeCallback)` (default no-op).
- `include/paimon/utils/prefetch_cache_config.h`: no new knobs; two defaults change - `range_size_limit` 32 MiB -> 8 MiB and `hole_size_limit` 8 KiB -> 512 KiB (behavior, not signature).

`ReadAheadCache::AddRanges` / `Warmup(uint64_t)` and the counter names live in the internal header `src/paimon/common/utils/read_ahead_cache.h` (`ReadAheadCache` is internal, not under `include/`).

### Documentation

No new user-facing configuration option; the read-ahead cache stays enabled by default with no option to set, and the two `CacheConfig` defaults are internal tuning. Prefetch behavior is covered by the existing `docs/source/user_guide/prefetch.rst`. No documentation changes required.

### Generative AI tooling

Generated-by: Qoder


include/paimon/utils/prefetch_cache_config.h
```diff
@@ -34,8 +34,9 @@ namespace paimon {
 /// ReadAheadCache to balance memory usage, I/O efficiency, and latency hiding.
 class PAIMON_EXPORT CacheConfig {
  public:
-    /// Returns the maximum allowed size (in bytes) for a single cached range.
-    /// Defaults to 32 MiB.
+    /// Returns the maximum allowed size (in bytes) for a single cached range, both for the ranges
+    /// registered up front and, as the cap, for the ones registered mid-read.
+    /// Defaults to 8 MiB.
```
lxy-9602, line 39: Is the default 8 MB or 4 MB?

src/paimon/format/parquet/page_filtered_row_group_reader.cpp
```diff
@@ -383,13 +391,22 @@ Result<std::shared_ptr<arrow::ChunkedArray>> PageFilteredRowGroupReader::ReadFil
...
+                 chunk_bytes,
+                 SaturatingDoubleToInteger<int64_t>(
+                     static_cast<double>(chunk_compressed_bytes) *
+                     static_cast<double>(kMaxReservationDecompressionFactor))});
         }
 
+        std::cerr << "[reserve_value_bytes] col_idx=" << col_idx
+                  << " reserve_values=" << reserve_values
+                  << " reserve_value_bytes=" << reserve_value_bytes << std::endl;
+
```
lxy-9602: Could you please confirm whether this is intended to be output here?

src/paimon/format/parquet/page_filtered_row_group_reader.cpp
```diff
@@ -44,11 +46,15 @@ namespace paimon::parquet {
...
-constexpr int64_t kMaxMetadataValueBytesReservation = int64_t{16} * 1024 * 1024;
+/// Multiple of a column chunk's compressed size that a leaf may reserve up front in value bytes.
+/// The estimate is a heuristic over footer fields, and `total_uncompressed_size` is a bare claim a
+/// forged footer can inflate to make a leaf eagerly allocate gigabytes for a near-empty file.
+/// `total_compressed_size` is instead checked against the file length before any read (Arrow's
+/// ComputeColumnChunkRange, and GetDataPageLayout below, bound a chunk and its pages by it), so a
+/// factor of it is bounded by bytes that physically exist. A leaf decompressing within the factor
+/// is reserved in full; a more compressible one is capped and left to the builder's amortized
+/// doubling, which is cheap against the I/O and decode of a read that large.
+constexpr int64_t kMaxReservationDecompressionFactor = 20;
```
lxy-9602: This change removes the previous absolute cap of 16 MiB and replaces it with a limit of up to `20 * total_compressed_size`. `ResetLeaf()` now calls `ReserveValueBytes()` eagerly; for normal RowGroups with hundreds of MiB compressed size and several GiB uncompressed size, this may allocate several GiB up front before decoding even begins. Since the average is computed over the entire chunk, it can also significantly overestimate the actual output when skipped pages contain large values while matched pages contain only small values, causing selective queries to OOM directly. I’d suggest keeping an absolute cap, or integrating this with a controlled memory budget.

include/paimon/reader/prefetch_file_batch_reader.h
```diff
@@ -135,6 +136,20 @@ class PAIMON_EXPORT PrefetchFileBatchReader : public FileBatchReader {
...
+    /// PreBufferRange() covers what is known up front; a reader whose ranges depend on data it
+    /// has already read - the late-materialization payload pass only knows which pages hold the
+    /// matched rows once the probe pass has run - reports them through this callback instead.
+    using PreBufferRangeCallback =
+        std::function<void(std::vector<std::pair<uint64_t, uint64_t>>&&)>;
+
+    /// Installs the callback above, or clears it when `callback` is empty. By default a reader has
+    /// no late byte ranges to report and ignores the callback.
+    /// @param callback The callback to report late byte ranges through.
+    virtual void SetPreBufferRangeCallback(PreBufferRangeCallback callback) {}
```
lxy-9602, line 152: I think the current complexity mainly comes from placing `LateMaterializingReader` below `PrefetchReader`:

```text
PrefetchReader → LateMaterializingReader → FormatReader
```

Each sub-reader discovers payload ranges after probing and must report them back to the outer layer. This introduces callbacks, lifetime management, incremental `AddRanges()`, deduplication, and `Warmup(offset)`.

Could we invert the structure?

```text
LateMaterializingReader → PrefetchReader → FormatReader
```

Then late materialization can run two normal prefetch phases:

1. Configure and finish the probe phase.
2. Build the matched bitmap.
3. Reconfigure the prefetch reader for the payload phase.

This removes the late-range callback path and keeps responsibilities clearer. The main trade-off is a global probe barrier, so we should benchmark first-batch latency and total scan time. Unless the regression is significant, this design seems easier to understand and maintain.

lucasfang: I will refactor the decoupling of the read-ahead cache and the prefetch reader in subsequent steps.

lxy-9602 (approved):
+1

PR #348, opened 2026-09-15
feat: add FromView/ToView api for BlobView

lszskye:
### Purpose
add FromView/ToView api for BlobView

### Tests
common/data/blob_view_test.cpp

### API and Format
include/paimon/data/blob_view.h


lxy-9602 (approved):
+1

PR #349, opened 2026-09-15
perf(aggregate): use hash lookup for scalar field aggregators

### LINKED ISSUE #363: [Feature] Optimize FieldCollectAgg and FieldMergeMapAgg lookup performance
### Search before asking

### Problem

`FieldCollectAgg` and `FieldMergeMapAgg` currently use linear scans with
`FieldAggregateUtils::Equals` for distinct-value and key lookup.

For large arrays/maps, repeatedly scanning previously collected values causes
quadratic behavior.

For example:

- `FieldCollectAgg` scans previously collected elements to detect duplicates.
- `FieldMergeMapAgg` scans existing keys through `FindKey` before
  overwriting or inserting an entry.

The implementation already contains a TODO noting that these lookups should use
type-aware hashing for hashable types, similar to the Java implementation.

### Proposed improvement

Add a hash-backed lookup path for types with well-defined semantic hashing,
while preserving the existing aggregation behavior.

The optimized path should:

- process both accumulator and input values;
- preserve first-seen order for `FieldCollectAgg`;
- preserve key order and input-overwrite semantics for `FieldMergeMapAgg`;
- hash STRING/BINARY by content;
- preserve FLOAT/DOUBLE semantics:
  - all NaNs compare equal;
  - `-0.0` and `+0.0` remain distinct;
- support TIMESTAMP consistently with its equality semantics;
- retain the existing linear fallback for DECIMAL and constructed types;
- retain linear lookup for small cardinalities where hash-table construction
  overhead is higher.

`FieldNestedUpdateAgg` can be considered separately as follow-up work.

### Performance

Focused benchmarks using the production aggregators show that the quadratic
lookup becomes significant once cardinality grows.

At 192 elements/entries:

| Workload | Current | Hash lookup | Speedup |
| --- | ---: | ---: | ---: |
| Collect INT64 | 127.5 us | 10.1 us | 12.6x |
| Collect STRING | 101.3 us | 11.9 us | 8.5x |
| MergeMap INT64 | 73.5 us | 15.8 us | 4.7x |
| MergeMap STRING | 96.6 us | 18.6 us | 5.2x |

At 8192 elements/entries:

| Workload | Current | Hash lookup | Speedup |
| --- | ---: | ---: | ---: |
| Collect INT64 | 109169.3 us | 375.3 us | 290.9x |
| Collect STRING | 180821.9 us | 646.0 us | 279.9x |
| MergeMap INT64 | 115265.6 us | 664.5 us | 173.5x |
| MergeMap STRING | 181807.6 us | 903.6 us | 201.2x |

Small-cardinality performance remains approximately unchanged by retaining the
existing linear path below the hash threshold.

### Related PR

#349

### Anything else?

_No response_

VietCT04:
### Purpose

Linked issue: #363 

Optimize distinct lookup in `FieldCollectAgg` and key lookup in
`FieldMergeMapAgg`.

The current implementation performs linear `FieldAggregateUtils::Equals`
scans while processing accumulator and input values. For large arrays/maps,
this leads to quadratic behavior.

This change adds a semantic hash-backed lookup path for supported types while
preserving the existing C++ aggregation semantics:

- process both accumulator and input, including duplicate accumulators;
- preserve first-seen order in `FieldCollectAgg`;
- preserve key order and input-overwrite semantics in `FieldMergeMapAgg`;
- hash STRING/BINARY by content;
- preserve floating-point semantics where all NaNs compare equal and `-0.0`
  remains distinct from `+0.0`;
- hash TIMESTAMP consistently with its equality semantics;
- keep DECIMAL and constructed types on the existing linear fallback;
- retain the linear path for small cardinalities to avoid hash-table overhead.

`FieldNestedUpdateAgg` is intentionally left unchanged.

A fixed threshold of 192 elements/entries is used for the hash path. This
keeps small inputs close to the current implementation while providing a clear
benefit once cardinality grows.

Focused benchmarks, using the actual production aggregators and pinned CPU
runs, show the crossover and large-cardinality scaling:

| Workload | N | Baseline | Hash path | Speedup |
| --- | ---: | ---: | ---: | ---: |
| Collect INT64 | 192 | 127.5 us | 10.1 us | 12.6x |
| Collect STRING | 192 | 101.3 us | 11.9 us | 8.5x |
| MergeMap INT64 | 192 | 73.5 us | 15.8 us | 4.7x |
| MergeMap STRING | 192 | 96.6 us | 18.6 us | 5.2x |
| Collect INT64 | 8192 | 109169.3 us | 375.3 us | 290.9x |
| Collect STRING | 8192 | 180821.9 us | 646.0 us | 279.9x |
| MergeMap INT64 | 8192 | 115265.6 us | 664.5 us | 173.5x |
| MergeMap STRING | 8192 | 181807.6 us | 903.6 us | 201.2x |

At `N=96`, performance remains approximately neutral:

| Workload | Baseline | Hash-enabled implementation |
| --- | ---: | ---: |
| Collect INT64 | 18.0 us | 18.2 us |
| Collect STRING | 26.5 us | 23.4 us |
| MergeMap INT64 | 20.6 us | 20.3 us |
| MergeMap STRING | 31.2 us | 32.1 us |

Repeated one-element merges also improve substantially once the accumulator
becomes large, despite rebuilding the temporary hash index on each `Agg` call:

| Workload | K | Baseline | Hash-enabled implementation | Speedup |
| --- | ---: | ---: | ---: | ---: |
| Collect INT64 | 512 | 81248.9 us | 13183.1 us | 6.2x |
| Collect INT64 | 2048 | 5121360.1 us | 149463.3 us | 34.3x |
| MergeMap INT64 | 512 | 111933.1 us | 18205.9 us | 6.1x |
| MergeMap INT64 | 2048 | 6049633.9 us | 297083.7 us | 20.4x |

At `K=128`, Collect remains approximately neutral (`1499.4 us -> 1557.4 us`).

### Tests

Added and ran focused unit tests covering:

- duplicate accumulator values;
- duplicate input values;
- duplicate accumulator and input map keys;
- accumulator/input overlap;
- first-seen ordering;
- map key ordering and overwrite behavior;
- null handling;
- STRING/BINARY content equality across different backing storage;
- FLOAT/DOUBLE signed zero, infinities, and multiple NaN payloads;
- TIMESTAMP equality/hash behavior;
- DECIMAL fallback;
- constructed-type fallback.

The hash-path tests use cardinalities above the threshold so they exercise the
optimized implementation directly. DECIMAL and constructed-type tests also
use large cardinalities to verify that they remain on the linear fallback
because of type support rather than the size threshold.

Results:

- `FieldCollectAggTest.*`: 13/13 passed
- `FieldMergeMapAggTest.*`: 5/5 passed
- relevant aggregate/factory tests: 10/10 passed
- `git diff --check` passed

A full integration/dependency rebuild was not run because it would rebuild the
bundled dependencies. No end-to-end compaction performance claim is made.

### API and Format

### Documentation

No documentation changes required. This is an internal performance
optimization with unchanged aggregation semantics.

### Generative AI tooling

Generated-by: OpenAI Codex

Close #363


src/paimon/core/mergetree/compact/aggregate/field_aggregate_utils.cpp
```diff
@@ -123,6 +128,25 @@ Result<bool> EqualMaps(const std::shared_ptr<InternalMap>& lhs,
...
+    Bits bits;
+    std::memcpy(&bits, &value, sizeof(value));
+    if (std::isnan(value)) {
+        if constexpr (sizeof(T) == sizeof(uint32_t)) {
+            bits = static_cast<Bits>(0x7fc00000U);
+        } else {
+            bits = static_cast<Bits>(0x7ff8000000000000ULL);
+        }
+    }
+    return std::hash<Bits>{}(bits);
```
lxy-9602: Consider using the floating-point utilities in `math.h` (for example, `CanonicalizeFloatingPoint` or `CanonicalizeFloatToIntBits`) instead of reimplementing similar logic.

src/paimon/core/mergetree/compact/aggregate/field_aggregate_utils.cpp
```diff
@@ -244,4 +268,67 @@ Result<bool> FieldAggregateUtils::Equals(const VariantType& lhs, const VariantTy
...
+        default:
+            return false;
+    }
+}
+
+size_t FieldAggregateUtils::Hash(const VariantType& value,
+                                 const std::shared_ptr<arrow::DataType>& type) {
+    assert(IsHashableType(type));
+    size_t result = std::hash<int>{}(static_cast<int>(type->id()));
+    if (DataDefine::IsVariantNull(value)) {
```
lxy-9602, line 278: For consistency across the codebase, please use `int32_t` instead of `int`.

src/paimon/core/mergetree/compact/aggregate/field_collect_agg_test.cpp
```diff
@@ -60,6 +63,11 @@ Result<std::unique_ptr<FieldCollectAgg>> MakeCollectAgg(bool distinct) {
     return FieldCollectAgg::Create(arrow::list(arrow::int32()), options, "f", GetDefaultPool());
 }
 
+Result<std::unique_ptr<FieldCollectAgg>> MakeDistinctAgg(
+    const std::shared_ptr<arrow::DataType>& element_type);
+
+VariantType Array(std::vector<VariantType> values);
+
```
lxy-9602, line 77: It seems you could move the function definition earlier, or rearrange the tests, to reduce the need for forward declarations.

src/paimon/core/mergetree/compact/aggregate/field_merge_map_agg.cpp
lxy-9602, line 182: I’m a bit curious: are there any plans to optimize the `Retract` method for `merge_map` aggregation using hashing?

lxy-9602:
Thank you for the hash optimization work on the aggregation! Could you please add an issue for this in the current PR and link it in the description? Thanks!

VietCT04:
@lxy-9602 Thanks for the review! I’ll create an issue for this and link it in the PR description.

src/paimon/core/mergetree/compact/aggregate/field_merge_map_agg.cpp
VietCT04: Good point. I kept Retract out of scope for this PR to keep the change focused. It still uses linear lookup, so I’d prefer to handle it as a follow-up optimization if profiling shows it is worthwhile.

VietCT04:
Hi @lxy-9602 , I’ve addressed the review comments and pushed the updates. Could you please take another look when you have a chance? Thanks!

lxy-9602 (approved):
+1

PR #357, opened 2026-09-16
feat(catalog): support writing and committing to a branch

### LINKED ISSUE #351: [Feature] Support committing snapshots to a branch
### Search before asking

### Motivation

In Java Paimon a commit is aimed at a branch: `SnapshotCommit#commit` takes `branch`, `CatalogSnapshotCommit` re-targets the table identifier at it (`new Identifier(database, table, branch)`), and `RenamingSnapshotCommit` writes the snapshot file into that branch's snapshot directory (`snapshotManager.copyWithBranch(branch)`). #326 ported the rest of `CatalogSnapshotCommit` but left the branch out, because the C++ request carried no identifier for a branch to apply to.

Paimon C++ cannot commit to a branch at all today:

- `SnapshotCommit::Commit()` has no branch parameter, so `CatalogSnapshotCommit` always commits to the identifier its context was built with, and `RenamingSnapshotCommit` always writes to the snapshot directory of the manager it was built with.
- `FileStoreCommit::Create()` builds the `SchemaManager` and the `SnapshotManager` for the main branch whatever the `branch` option says, while `FileStoreWrite::Create()` is already branch-aware. A writer aimed at `dev` therefore reads that branch's schema and snapshots while the commit publishes on main.
- `CommitContextBuilder` and `WriteContextBuilder` refuse a non-main branch outright once a catalog is configured ("a commit through a catalog reads the main branch of the table"), so a catalog table's branch can be neither written nor committed.
- `ReadContext::GetBranch()` reflects only `WithBranch()`, so a read aimed at a branch through the `branch` option alone - which is how a scan of one is aimed, and how `RestCatalog` reports a branch table's options - takes the main branch's schema, while the splits, the path factory and the split readers follow the option.

The integration tests show the gap: a branch is fabricated by copying the `schema` and `snapshot` directories, and `test/inte/pk_compaction_inte_test.cpp` writes and compacts on a branch but can only verify the commit messages, "since we don't support branch commit and scan".

### Solution

Port the branch through the snapshot-commit layer, then resolve the branch once for every entry point.

- `SnapshotCommit::Commit(base_snapshot_uuid, snapshot, branch, statistics)`, as in Java. `CatalogSnapshotCommit` commits to `Identifier(database, data table name, branch)` so the catalog answers for the branch, with the main branch staying the bare table name, matched ignoring case as the identifier the Java client builds is. `RenamingSnapshotCommit` writes the snapshot file and the `LATEST` hint through a manager of the branch asked for, mirroring `copyWithBranch`. `FileStoreCommitImpl` passes `snapshot_manager_->Branch()`: the branch whose metadata the commit reads is the one it publishes on, which is the convention its real-time offsets already follow.
- New `Identifier(database, table, branch)`, equivalent to the Java constructor, building the `<table>$branch_<branch>` object name.
- New `BranchManager::ResolveBranch(identifier, options, asked_branch, operation)`: the identifier, the `branch` option and `WithBranch()` may each name the branch, an absent or empty name is the main branch, and two different names are refused rather than silently chosen between. `CommitContextBuilder`, `WriteContextBuilder` and `ReadContextBuilder` all resolve through it, and with a catalog a non-main branch has to be named in the identifier, since the catalog is asked for the branch's schema and snapshot by identifier.
- `FileStoreCommit::Create()` builds the `SchemaManager`, the format-table dispatch and the `SnapshotManager` for the resolved branch, and puts it back into the merged `CoreOptions` so that what reads through them reads the same branch - manifest entry cache keys are scoped by it. `FileStoreWrite::Create()` does the same, which also fixes a `WithBranch()` write whose compaction reads went through main-scoped options.

### Anything else?

Behaviour changes: a file-system commit with `branch=dev` publishes to that branch instead of silently to main; `WithCatalog()` accepts `tbl$branch_dev`; `WriteContext::GetBranch()` and `ReadContext::GetBranch()` return the resolved branch, so `WithBranch("")` reads back as `main`.

A branch keeps its own schema, snapshots and real-time offsets under `branch/branch-<name>` while data and manifests stay shared under the table path, as in Java. Two consequences worth documenting: the branch's schema is read from the table directory rather than from the catalog, so a branch committed to through a catalog needs its schema published there; and expiration walks the retained snapshots of the branch it commits to while the data files are shared with the other branches of the table.

Public API touched: the new `Identifier` constructor, and the documented behaviour of `CommitContextBuilder::WithCatalog()`, `WriteContextBuilder::WithBranch()` and `ReadContextBuilder::WithBranch()`. No storage format or protocol change.

SteNicholas:
### Purpose

Linked issue: close #351

Java Paimon addresses a branch by the `<table>$branch_<branch>` object name and hands that branch down to `SnapshotCommit#commit(baseSnapshotUuid, snapshot, branch, statistics)`, while `AbstractFileStoreTable#switchToBranch` reads the branch's schema from `branch/branch-<name>` through a `FileSystemSchemaManager`. Paimon C++ had none of this on the commit path: `FileStoreCommit::Create()` built its `SchemaManager` and `SnapshotManager` without a branch, so a commit configured with `branch=dev` read the main schema and published its snapshot under the main branch's directory, and a write through a catalog refused a branch outright with `a write through a catalog requires the main branch`.

This PR makes a branch a first-class target of the write and commit path, for a file-system commit, a catalog commit and a request-only commit alike.

**Resolving the branch**

- `BranchManager::ResolveBranch()` resolves the branch an operation is aimed at from every source that may name it: the table identifier, the `branch` option and, where a builder has one, `WithBranch()`. Names that disagree are refused rather than chosen between, and `tbl$branch_main`, `branch=main`, an empty and a blank name all mean the main branch. A read resolves it the same way, so the `branch` option now names a read's branch as `WithBranch()` already did.
- Naming a branch only in the option while the identifier of `WithCatalog()` names the bare table is refused: the catalog would answer for the table while the files were written for the branch.

**Addressing the branch on a catalog**

- `Identifier(database, table, branch)` builds the `<table>$branch_<branch>` object name, folding every spelling of `main` into the bare table name as the Java client does. `CatalogUtils::BranchIdentifier()` rebuilds it for every catalog call, so what is read and what is committed name one object; `RestCatalog` now goes through the same builder instead of splicing the name itself.
- `SnapshotCommit::Commit()` takes the branch. `CatalogSnapshotCommit` commits to the branch's object name, and `RenamingSnapshotCommit` writes the snapshot through a manager of that branch.
- `UseRESTCatalogCommit()` still builds a body that names no branch, so the caller sends it to the branch's own URL: the commit endpoint of `tbl$branch_dev`, not the one of `tbl`.
- Two kinds of name a catalog cannot tell apart from another object are refused where a builder sees them: a branch differing from `main` by case alone, whose snapshots would go to a directory of their own, and a branch holding a `$`, which reads back as a system table of another branch.

**Schema of a branch**

- A branch keeps its schema, snapshots and real-time offsets under `branch/branch-<name>`, while data and manifests stay shared under the table path. A write or a commit aimed at a branch reads the schema published there rather than the catalog's, as `switchToBranch` does, so a catalog serving only the table's own current schema still takes the branch's snapshots. Only the main branch reloads the catalog's schema id for a new snapshot.
- This library creates no branch: one has to exist, with its schema published under `branch/branch-<name>`, before a write or a commit can be aimed at it. Catalog branch management (creating, deleting and merging a branch) remains unimplemented.

**Expiration**

`Expire()` and snapshot loading follow the commit's branch, so expiring on a branch manages that branch's snapshots and its `EARLIEST` hint. As in Java's `ExpireSnapshotsImpl`, only the retained snapshots of that branch are read while every branch shares the table's data files, so a file that only another branch still refers to is deleted. This is stated in `include/paimon/file_store_commit.h` and called out as a warning in the catalog guide; cross-branch reachability belongs to orphan-file cleaning, as it does upstream.

### Tests

New and extended UT:

- `identifier_test.cpp`: `BranchConstructor` (object name, round trip, and every spelling of `main` folding to the bare table); a blank branch added to `InvalidEmptySystemTableNameParts`.
- `branch_manager_test.cpp`: `TestResolveBranch` (each source, agreement, disagreement, names that would leave the table root), `TestCheckCatalogAddressableBranch`.
- `catalog_snapshot_commit_test.cpp`: `TestCommitToBranch` (branch object name per attempt, the branch of the attempt winning over the one the identifier was built for, `MAIN` refused); `TestBuildRequestWithoutCatalog` extended so a refused attempt leaves no request behind.
- `renaming_snapshot_commit_test.cpp`: `TestCommitToBranch` (snapshot and `LATEST` under the branch, main untouched, and a commit aimed back at main from a branch manager).
- `commit_context_test.cpp` / `write_context_test.cpp`: `TestCatalogAddressesBranchByIdentifier`, `TestBranch`. `read_context_test.cpp`: `TestBranch`.
- `file_store_commit_impl_test.cpp`: `TestCommitToBranch`, `TestRequestOnlyCommitToBranch`, `TestCatalogCommitToBranchTakesTheBranchSnapshot`, `TestBranchCommitReadsTheSchemaFromTheBranch`, `TestBranchCommitNamesTheBranchWhenItFails`, `TestCatalogCommitCanonicalizesTheMainBranchIdentifier`, `TestBranchExpireLeavesMainSnapshotMetadataAlone`.
- `file_store_write_test.cpp`: `TestCatalogWriteAsksTheCatalogOnlyForTheMainBranchSchema`.
- `rest_catalog_test.cpp`: `ListSnapshots` extended with a blank branch, which is the bare table as `main` and `MAIN` are.

New and extended IT:

- `write_inte_test.cpp`: `TestBranchCommitAndRead` writes and commits on `dev` and on main, then reads each back.
- `pk_compaction_inte_test.cpp`: `WriteAndCompactWithBranch` now commits the compaction to `branch-rt` and scans the branch for the compacted file, instead of verifying the commit messages only.

### API and Format

Yes.

- `include/paimon/catalog/identifier.h`: new `Identifier(database, table, branch)` constructor.
- `include/paimon/file_store_commit.h`: `GetLastCommitTableRequest()` now clears the request of the previous attempt, so an attempt that failed before building one returns an error rather than the request before it. `Expire()` documents that only the retained snapshots of its own branch are read.
- `include/paimon/commit_context.h`, `include/paimon/write_context.h`, `include/paimon/read_context.h`: `WithCatalog()`, `WithBranch()` and the `branch` option document how a branch is named and how disagreeing names are refused. A commit or a catalog write aimed at a branch is accepted where it was refused before.
- No storage format change: `<table>$branch_<branch>` and `branch/branch-<name>` are the existing Java spellings.

### Documentation

Yes.

- `docs/source/user_guide/catalog.rst`: a new **Committing to a branch** section with **Naming the branch**, **What a branch keeps and what it shares** and **Addressing a branch without a catalog client**; the expiration warning covering data files shared by every branch; and a note that branch management covers creating, deleting and merging a branch while committing to an existing one is supported.
- `docs/source/user_guide/write.rst`: the main branch loads its current schema from the catalog while a branch reads the schema published under `branch/branch-<name>`, and a request-only commit aimed at a branch has to be sent to that branch's URL.

### Generative AI tooling

Generated-by: Claude Code (claude-opus-5[1m])


src/paimon/core/operation/file_store_commit.cpp
```diff
@@ -272,11 +295,12 @@ Result<std::unique_ptr<FileStoreCommit>> FileStoreCommit::Create(
             options.LegacyPartitionNameEnabled(), external_paths, global_index_external_path,
             options.IndexFileInDataFileDir(), ctx->GetMemoryPool()));
 
-    auto snapshot_manager = std::make_shared<SnapshotManager>(options.GetFileSystem(), root_path);
+    auto snapshot_manager =
```
lucasfang, line 298: `FileStoreCommit::Create()` passes a branch-scoped `SnapshotManager` (`file_store_commit.cpp:298`, `:330`) to `ExpireSnapshots`, whose protection sets (`GetManifestSkippingSet` / `GetDataFileSkippingSet`) come only from that branch's `retained_snapshots`. Expiration never consults a tag and never scans other branches, while data/manifest files are shared across branches. So expiring `dev` deletes a file that has left `dev`'s retained set even when `main` still references it, and `main` then fails to read.

**Repro:** create `dev` from a `main` tag, compact/overwrite on `dev`, then `Expire()`.

The docs claim this matches Java's `ExpireSnapshotsImpl`, but that is only half true. Java collects tagged snapshots *before* deleting files (apache/paimon#9220), and a branch created from a tag carries that tag — which is exactly what protects the shared base files. C++ already has a branch-aware `TagManager` but expiration ignores it. Java is equivalent only for files referenced by another branch's *untagged* live snapshots.

1. This PR's `expire_snapshots.cpp` diff is log-only, so the missing tag protection is a pre-existing gap that also hits `main`; the PR newly makes it reachable on branches.
2. `TestBranchExpireLeavesMainSnapshotMetadataAlone` cannot catch it — `main` and `dev` commit the same append fixtures, so no shared file is actually deleted, and it only checks snapshot JSON/hints, not read integrity.

Suggested fix (in order):
1. Fold `TagManager`'s tagged snapshots into the skipping sets, matching Java's `ExpireSnapshotsImpl`.
2. Otherwise, refuse or hard-warn at runtime on branch expiration until that lands.
3. Add a regression test that compacts/overwrites on `dev` so a file truly leaves its retained set while `main` still references it, then reads `main` back in full.

A documentation warning alone is not enough for a silent data-loss path.

test/inte/pk_compaction_inte_test.cpp
```diff
@@ -1954,35 +1954,77 @@ TEST_F(PkCompactionInteTest, WriteAndCompactWithBranch) {
...
+        ASSERT_NE(split_impl, nullptr);
+        for (const std::shared_ptr<DataFileMeta>& file : split_impl->DataFiles()) {
+            if (file->file_name == compact_after[0]->file_name) {
+                ASSERT_EQ(split_impl->SnapshotId(), 3);
+                compacted_splits.push_back(split);
+                break;
+            }
+        }
+    }
+    ASSERT_EQ(compacted_splits.size(), 1u);
```
lxy-9602: Reconstructing the split inside the integration test feels a bit unusual. Would it make more sense to just verify it normally with `ASSERT_EQ(split_impl->SnapshotId(), 3);`, and then use the full split for reading?

docs/source/user_guide/catalog.rst
```diff
@@ -224,7 +222,73 @@ restored by ``RollbackToAsLatest``. Serialize rollback and expiration through
...
+
+   Expiration reads the retained snapshots of the branch it runs on and of no
+   other, while data files are shared by every branch of the table, so a file
+   that only another branch still refers to is not preserved by that reference
+   and is deleted. This holds for the main branch as much as for the others:
+   expiring the main branch deletes a file only a branch refers to. Expire only
+   where the retained snapshots of the branch cover every file the other branches
+   still read, or keep those files reachable from the expiring branch through the
+   upstream coordinator.
+
```
lxy-9602, line 236: Could we add some safeguards here? For example, perhaps expiration should be unsupported for tables that have branches.

src/paimon/core/catalog/renaming_snapshot_commit.h
```diff
@@ -47,10 +48,17 @@ class RenamingSnapshotCommit : public SnapshotCommit {
...
+                        const Snapshot& snapshot, const std::string& branch,
                         const std::vector<PartitionStatistics>& statistics) override {
         PAIMON_ASSIGN_OR_RAISE(std::string json_str, snapshot.ToJsonString());
-        std::string snapshot_path = snapshot_manager_->SnapshotPath(snapshot.Id());
+        // The snapshot directory belongs to a branch, so a commit aimed at another one writes
+        // through a manager of that branch.
+        std::shared_ptr<SnapshotManager> snapshot_manager = snapshot_manager_;
+        if (BranchManager::NormalizeBranch(branch) != snapshot_manager_->Branch()) {
+            snapshot_manager = std::make_shared<SnapshotManager>(
+                snapshot_manager_->Fs(), snapshot_manager_->RootPath(), branch);
```
lxy-9602: I’d like to understand whether, in the current production code, `snapshot_manager_` can actually belong to a different branch than the input `branch`. If that is not possible, could we make this check stricter and fail explicitly here?

src/paimon/core/operation/file_store_commit_impl_test.cpp
```diff
@@ -721,6 +722,194 @@ TEST_F(FileStoreCommitImplTest, TestCatalogCommitTakesTheSnapshot) {
...
+    CatalogCommitSpec spec;
+    spec.catalog = catalog;
+    spec.identifier.emplace("db", "tbl$branch_dev");
+    spec.table_id = "table-uuid";
+    ASSERT_OK_AND_ASSIGN(std::unique_ptr<FileStoreCommit> commit, CreateCatalogCommit(spec));
+
+    Status failed = commit->Commit(CommitMessagesOfRound(1));
+    ASSERT_NOK_WITH_MSG(failed, "catalog unreachable");
+    ASSERT_NE(failed.ToString().find("(branch dev, through catalog"), std::string::npos)
+        << failed.ToString();
```
lxy-9602: `ASSERT_NOK_WITH_MSG`

test/inte/pk_compaction_inte_test.cpp
SteNicholas: Done in 0af0eea. The test no longer picks the split holding the compacted file: it asserts `SnapshotId() == 3` for every split of `branch-rt` and reads all of them, so the expected result is now every row of the branch.

docs/source/user_guide/catalog.rst
SteNicholas: Done in 0af0eea. `Expire()` now returns `NotImplemented` before reading any snapshot or deleting any file when it runs on a branch other than main, or when it runs on main and finds another branch under `branch/branch-<name>` of the table path; that error lists the branches found.

A branch held only by a catalog, or one created while expiration runs, is not found by this check, so `FileStoreCommit::Expire()` and the catalog and clean guides call that out and ask to serialize branch creation and expiration.

`TestExpireOnBranchIsNotSupported` and `TestExpireOnMainOfTableWithBranchesIsNotSupported` cover each refusal and check that no snapshot, manifest or data file is deleted.

src/paimon/core/catalog/renaming_snapshot_commit.h
SteNicholas: It cannot in the current production code: `FileStoreCommit::Create()` builds one branch-aware `SnapshotManager` and hands it both to `RenamingSnapshotCommit` and to `FileStoreCommitImpl`, which passes `snapshot_manager_->Branch()` to `SnapshotCommit::Commit()`.

So a mismatch can only be a caller bug, and in 0af0eea it returns `Invalid` instead of switching to a manager of the other branch. `RenamingSnapshotCommitTest.TestBranchHasToMatchSnapshotManager` covers both directions and checks that nothing is written. This is stricter than Java's `RenamingSnapshotCommit`, which switches with `copyWithBranch`.

src/paimon/core/operation/file_store_commit_impl_test.cpp
SteNicholas: Done in 0af0eea: both checks use `ASSERT_NOK_WITH_MSG` now.

lxy-9602 (approved):
+1

lucasfang (approved):
+1

PR #366, opened 2026-09-17
perf(read): warm up only raw bytes by default

### LINKED ISSUE #365: [Feature] Default WarmupLevel to RAW instead of DECODED
## Search before asking

## Motivation

`WarmupLevel::DECODED` is the current default, so every read that goes through the prefetching reader starts a background decode loop on the next file while the current one is still being consumed. That is the most aggressive level: it materializes decoded Arrow batches for a file the query has not reached yet, so the peak footprint of a read carries the decoded batches of the current file plus those of the next one, per reader in flight. A merge-on-read section with `k` sorted runs warms one file per leaf, which multiplies that extra decoded buffer by `k`. The cost is paid up front and is wasted whenever the scan stops early — a `LIMIT`, or a predicate selective enough to empty a split — because the warmed file is then never read. The latency it hides is dominated by the remote fetch, which the cheaper `RAW` level already covers, so the default trades a large amount of memory for the smaller, decode-side part of the win.

## Solution

Make `WarmupLevel::RAW` the default and leave `DECODED` as an explicit opt-in for latency-bound readers that can afford the memory.

- `ReadContextBuilder` defaults `warmup_level_` to `RAW`, in both the member initializer and `Reset()`, so a reused builder keeps the same default.
- `DataFileReadOptions::warmup_level` defaults to `RAW` as well, which is what an internal caller that builds read options directly gets.
- The enum and `SetWarmupLevel()` documentation move the "this is the default" note to `RAW`.

No enum value, signature, or plumbing changes: `NONE`, `RAW` and `DECODED` all keep their behavior, and a caller that sets a level explicitly is unaffected.

## Anything else?

One behavior consequence worth stating: `RAW` warms through the read-ahead cache and falls back to `NONE` when that cache is disabled, so a reader built with `SetReadAheadCacheEnabled(false)` now gets no warmup by default, where it previously got a background decode. Such a reader has to ask for `DECODED` explicitly to keep warming. This is a default-value change to public API behavior in `include/paimon/`; no signature, storage format, or protocol change.

lucasfang:
### Purpose

Linked issue: close #365 

`WarmupLevel::DECODED` was the default, so every prefetching read started a background decode loop on the next file while the current one was still being consumed, holding the decoded batches of two files at once per reader in flight — multiplied by the number of sorted runs in a merge-on-read section, since `LoserTree` warms one file per leaf. That memory is committed before the query reaches the file and is wasted when the scan stops early, while most of the latency it hides is the remote fetch that the cheaper `RAW` level already covers.

This changes the default to `WarmupLevel::RAW` and leaves `DECODED` as an explicit opt-in. No enum value, signature, or plumbing change; every level keeps its behavior, and a caller that sets a level explicitly is unaffected.

- `ReadContextBuilder` defaults `warmup_level_` to `RAW` in both the member initializer and `Reset()`, so a builder reused after `Finish()` gets the same default.
- `DataFileReadOptions::warmup_level` defaults to `RAW`, which is what an internal caller that fills read options directly gets.
- The `WarmupLevel` enum and `ReadContextBuilder::SetWarmupLevel()` docs move the "this is the default" note to `RAW`.

One consequence to note: `RAW` warms through the read-ahead cache and falls back to `NONE` when that cache is off, so a reader built with `SetReadAheadCacheEnabled(false)` now warms nothing by default, where it previously got a background decode. Such a reader has to ask for `DECODED` explicitly.

### Tests

No new case: the level-specific behavior is unchanged, and the existing coverage iterates all three levels rather than relying on which one is the default — `PrefetchFileBatchReaderImplTest.TestWarmupLevelNone` / `.TestWarmupLevelRaw` / `.TestWarmupLevelDecoded` (each over `{parquet, orc}` x `{read-ahead cache on, off}`), `DataFileReaderFactoryTest.OpenForwardsWarmupLevelToPrefetchReader`, and `SystemTableTest.TestNewReadPropagatesWarmupLevel`.

Updated the cases that pin the default: `ReadContextTest.TestDefaultValue` and `ReadContextTest.TestSetWarmupLevel` (the post-`Finish()` reset now expects `RAW`). `ReadContextTest.TestSetContent` set `RAW`, which is now the default and would no longer catch a dropped setter, so it sets `DECODED` instead.

Ran: `paimon-common-test` (1654 cases, 1652 passed, 2 pre-existing skips) and `paimon-core-test` (2187 cases, all passed), plus the directed `ReadContextTest.*:SystemTableTest.*` and `DataFileReaderFactoryTest.*` filters. `pre-commit run --files <changed files>` and `git diff --check` are clean. Integration tests and benchmarks were not run.

### API and Format

A default-value change to public API behavior in `include/paimon/`: `ReadContextBuilder::SetWarmupLevel()` now defaults to `WarmupLevel::RAW`. No signature, ABI, storage format, or protocol change.

### Documentation

Not a new feature. The doc comments on `WarmupLevel` and `SetWarmupLevel()` are updated in place, which is what `docs/source/api/read.rst` renders.

### Generative AI tooling

Generated-by: Qoder


lxy-9602 (approved):
+1

PR #368, opened 2026-09-17
fix(global-index): align scores with row IDs after filtering

### LINKED ISSUE #367: [Bug] Global Index scores become misaligned after deletion-vector or predicate filtering
### Search before asking

### Paimon-cpp version

`main` at commit `70165ff127b9129234c81deab745332d0e12cdfa`.

### Minimal reproduce step

1. Create a data-evolution table with row tracking and deletion vectors enabled. Write 12 rows with `f0 = 0..11` and global row IDs `100..111`.
2. Supply a scored Global Index result through `ScanContextBuilder::SetGlobalIndexResult`: candidate row IDs `100, 101, 102, 104, 105, 106, 107, 108, 109, 110, 111`, with each score equal to `row_id + 0.5`.
3. Read `f0` and `_INDEX_SCORE` with predicate filtering enabled and the predicate `f0 != 2 AND f0 < 10`. Repeat with and without projecting `_ROW_ID`.
4. Also exercise deletion-vector filtering, both by itself and together with the predicate. For example, delete file-relative positions `1, 4, 5, 6, 7, 11`.
5. Repeat after a partial-column update that requires merging files covering the same row-ID range.

Each surviving row should keep its own Global Index score after predicate or deletion-vector filtering. In the predicate-only example, row ID `104` should have score `104.5`.

Instead, score completion advances a cursor through the original score vector for surviving rows. Once an earlier candidate is filtered out, subsequent rows can receive another candidate's score; row ID `104` receives `102.5` in the regression example.

The result should be identical whether the user projects `_ROW_ID` or not. An internally added row-ID column must not appear in the user's output. Indexed splits without scores should continue returning null `_INDEX_SCORE` values.

### Anything else?

The issue affects Global Index score completion combined with deletion vectors and/or predicates. Regression coverage exercises single-file reads and reads that merge columns from multiple files, using Parquet, ORC and Avro. File Index score handling is outside this issue's scope.

zjw1111:
### Purpose

Closes #367.

Global Index score completion currently consumes scores sequentially after deletion-vector or predicate filtering has removed candidate rows. A surviving row can therefore receive another row's score; for example, filtering out row ID 102 can give row ID 104 the score 102.5 instead of 104.5.

Match scores to surviving global row IDs. Add `_ROW_ID` to the internal read schema when needed and remove it before returning batches, while preserving explicit row-ID projections and keeping the shared read schema unchanged. Indexed splits without scores continue to return null `_INDEX_SCORE` values.

Validate scored row ranges before constructing the lookup, rejecting invalid endpoints, overlaps and score-count mismatches. This change covers Global Index; File Index behavior is outside its scope.

### Tests

- `CompleteIndexScoreBatchReaderTest.*`: sparse and non-monotonic row IDs, bitmap filtering, multiple batch sizes, both row-ID projections, unselected null scores, and invalid score/schema inputs.
- `IndexedSplitTest.*`: range overlap, invalid endpoints, boundary values and score counts.
- `DataEvolutionSplitReadTest.*`: existing split-reader coverage.
- `DataEvolutionTableTest.TestGlobalIndexScoresWithSingleFileFiltering` and `TestGlobalIndexScoresWithMergedFileFiltering`: Global Index combined with DV, predicates and both together; with/without `_ROW_ID`; empty results; scored and unscored splits reused with the same `TableRead`. Parameterized over Parquet/ORC/Avro and manifest-cache settings.
- Final local validation: 35 focused unit tests, 138 Global Index integration tests and 216 Data Evolution integration tests passed. The Data Evolution suite skipped 18 existing cases because of format capabilities.
- Changed-file pre-commit checks and `git diff --check` passed.
- Regression control using the original sequential-score path failed all 12 new integration cases with score mismatches.

Debug builds used ASan with `ASAN_OPTIONS=detect_leaks=0` because LeakSanitizer cannot scan `/proc` in the sandbox. The complete repository test suite and optional Lumina/Lucene/Tantivy/Mosaic backend configurations were not run.

Commands:
```bash
ASAN_OPTIONS=detect_leaks=0 ./build/debug/score-alignment-test
ASAN_OPTIONS=detect_leaks=0 ./build/debug/paimon-global-index-test
ASAN_OPTIONS=detect_leaks=0 ./build/debug/paimon-data-evolution-table-test
```
The focused unit executable links the existing score-reader, indexed-split and data-evolution split-reader test translation units.

### API and Format

No public API under `include/`, storage format or protocol changes. The internal `CompleteIndexScoreBatchReader` constructor now accepts scores keyed by global row ID.

### Documentation

No new user-facing feature or configuration. Internal reader preconditions and score-alignment behavior are documented in the class header.

### Generative AI tooling

Generated-by: OpenAI Codex CLI 0.154.0


src/paimon/core/global_index/indexed_split_impl.h
```diff
@@ -98,7 +100,22 @@ class IndexedSplitImpl : public IndexedSplit {
...
+                }
+                if (range.to == std::numeric_limits<int64_t>::max()) {
+                    return Status::Invalid(
+                        "Row id range upper bound must be less than INT64_MAX in scored indexed "
+                        "split.");
+                }
+                if (previous_end && range.from <= previous_end.value()) {
+                    return Status::Invalid("Duplicate row id in scored indexed split.");
+                }
+                previous_end = range.to;
```
lxy-9602: Could you clarify the purpose of this added part? I couldn’t find similar logic in Java `IndexedSplit`.

src/paimon/core/operation/data_evolution_split_read.cpp
```diff
@@ -215,11 +215,32 @@ Result<std::unique_ptr<BatchReader>> DataEvolutionSplitRead::CreateReader(
...
+                }
+            }
+            remove_row_id = read_schema->GetFieldIndex(SpecialFields::RowId().Name()) < 0;
+            if (remove_row_id) {
+                PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(
+                    read_schema, read_schema->AddField(read_schema->num_fields(),
+                                                       DataField::ConvertDataFieldToArrowField(
+                                                           SpecialFields::RowId())));
+            }
+        }
```
lxy-9602, line 237: I’m wondering whether the mapping from id to score, as well as whether to remove `row_id`, could be handled in `CompleteIndexScoreBatchReader` instead. We may have similar needs later for the global PK index, and doing it there could reduce the amount of split-read-specific implementation. Also, could `CompleteIndexScoreBatchReader` be placed in `FileBatchReader`? In that case, `SetReadSchema` could naturally take care of adding or removing the `row_id` field as needed.

src/paimon/core/global_index/indexed_split_impl.h
lxy-9602: The `rowRanges` produced by the Java generation path should already have strictly increasing start positions and no overlaps. Given that, is sorting here actually necessary?

test/inte/data_evolution_table_test.cpp
```diff
@@ -402,6 +404,130 @@ class DataEvolutionTableTest : public ::testing::Test,
...
+                        }
+                        std::shared_ptr<arrow::ChunkedArray> expected;
+                        ASSERT_TRUE(arrow::ipc::internal::json::ChunkedArrayFromJSON(
+                                        arrow::struct_(expected_fields),
+                                        {fmt::format("[{}]", fmt::join(expected_rows, ","))},
+                                        &expected)
+                                        .ok());
+                        ASSERT_TRUE(expected->Equals(*result))
+                            << "actual=" << result->ToString()
+                            << "\nexpected=" << expected->ToString();
```
lxy-9602, line 536: I find the current test a bit hard to follow. Could we at least simplify the correspondence between the inputs and the expected results, for example by specifying it more explicitly?

src/paimon/common/global_index/complete_index_score_batch_reader_test.cpp
```diff
@@ -150,14 +154,106 @@ TEST_F(CompleteIndexScoreBatchReaderTest, TestReadWithNullScores) {
...
+                        .ok());
+        for (int32_t batch_size : {1, 3, 10}) {
+            SCOPED_TRACE(fmt::format("remove_row_id={}, batch_size={}", remove_row_id, batch_size));
+            auto inner = std::make_unique<MockFileBatchReader>(
+                src_array, src_array->type(), RoaringBitmap32::From({0, 2, 3}), batch_size);
+            inner->EnableRandomizeBatchSize(false);
+            auto reader = std::make_unique<CompleteIndexScoreBatchReader>(
+                std::move(inner),
+                std::unordered_map<int64_t, float>{
+                    {100, 1.0f}, {101, 2.0f}, {102, 3.0f}, {108, 9.0f}, {109, 10.0f}},
```
lxy-9602: Could we move the map before `expected_json`? Otherwise, `expected_json` is a bit hard to understand on its own.

src/paimon/core/operation/data_evolution_split_read.cpp
zjw1111: Let's defer this refactoring until the global PK index actually needs to reuse this logic. I'll keep the current structure in this PR and revisit the shared abstraction when that use case is implemented.

lszskye (approved):
+1

lxy-9602 (approved):
+1

PR #370, opened 2026-09-17
feat(scan): support file indexes in real-time and PK scans

lxy-9602:
### Purpose

Linked issue: #158 

This change enables file-index pruning for real-time reads and aligns scan-time index handling across append-only and primary-key tables.

- Apply embedded and external file indexes in the real-time merge reader.
- Combine bitmap index selections with deletion vectors for PK tables.
- Add scan-time pruning with embedded file indexes for both append-only and PK tables.
- Keep external index evaluation in the reader stage, consistent with Java Paimon.

### Tests

- append-only and PK+DV real-time reads using embedded and external bitmap indexes;
- scan pruning when min/max statistics retain a file but its embedded bitmap index filters it;
- reader-stage bitmap filtering when a split is retained;
- predicate reconstruction after a field rename through schema evolution.

### Generative AI tooling


src/paimon/core/operation/raw_file_split_read.cpp
```diff
@@ -189,91 +180,4 @@ Result<bool> RawFileSplitRead::Match(const std::shared_ptr<Split>& split,
...
-        PAIMON_ASSIGN_OR_RAISE(
-            file_index_result,
-            FileIndexEvaluator::Evaluate(data_schema, predicate, data_file_path_factory, file,
-                                         options_.GetFileSystem(), pool_));
-        PAIMON_ASSIGN_OR_RAISE(bool is_remain, file_index_result->IsRemain());
-        if (!is_remain) {
-            return std::unique_ptr<FileBatchReader>();
-        }
-    }
-    // prepare selection bitmap for index
```
zjw1111, line 209: Do these comments still need to be kept? They have been deleted now after the migration.

zjw1111 (commented):
One performance concern in the merge-reader fallback.

src/paimon/core/operation/abstract_split_read.cpp
```diff
@@ -138,13 +168,92 @@ Result<std::vector<std::unique_ptr<FileBatchReader>>> AbstractSplitRead::CreateR
...
+        return std::unique_ptr<FileBatchReader>();
+    }
+
+    ::ArrowSchema c_read_schema;
+    PAIMON_RETURN_NOT_OK_FROM_ARROW(arrow::ExportSchema(*read_schema, &c_read_schema));
+    PAIMON_RETURN_NOT_OK(file_reader->SetReadSchema(&c_read_schema, predicate, actual_selection));
+
+    std::unique_ptr<FileBatchReader> reader;
+    if (!file_reader->SupportPreciseBitmapSelection() && actual_selection) {
+        reader = std::make_unique<ApplyBitmapIndexBatchReader>(std::move(file_reader),
```
zjw1111, line 247: This changes the DV-only merge path from `ApplyDeletionVectorBatchReader` to `ApplyBitmapIndexBatchReader`. The latter starts at `bitmap_.Begin()` for every batch and advances through all selected row IDs preceding that batch. With a sparse DV, its complement contains almost every row, so reading N rows in batches of B requires O(N² / B) bitmap iterator increments, instead of the previous per-row DV membership checks. Both Parquet and ORC report `SupportPreciseBitmapSelection() == false`, so this regression also affects reads with file indexes disabled or no predicate.

zjw1111 (approved):
+1

SteNicholas (approved):
LGTM.

PR #373, opened 2026-09-18
fix(data issue): fix dictionary encoded binary read as empty

### LINKED ISSUE #381: [Feature] dictionary encoded binary support
### Search before asking

### Motivation

Currently paimon-cpp does not handle dictionary for binary type. It's OK for the default parquet reader because it  would not return a dictionary for the binary type.
But for customed parquet readers or if there're `ARROW:schema` meta data in the parquet footer, it's possible to return dictionary for a binary type. In that case, there'll be data error since ColumnarUtils::GetView returns empty.

### Solution

_No response_

### Anything else?

_No response_

zhangweilst:
### Purpose

Dictionary encoded binary will be read out empty in that ColumnarUtils::GetView didn't handle Binary type correctly.
This patch fix this data error issue and add tests.

Linked issue: close #381 

Fix dictionary encoded binary read data issus.

### Tests

Added UT in the patch.

### API and Format

No

### Documentation

No


lxy-9602:
Thanks a lot for the fix. Could you also create a corresponding issue for this PR to describe the problem?

I’d also like to better understand your use case: are you using your own Parquet reader in your environment? It seems the default Parquet reader would not return a dictionary for the binary type.

Also, there appear to be several places in the current codebase that forcibly assume a string dictionary, which may have similar issues. Would you prefer to address those in this PR as well, or leave them for follow-up PRs?

zhangweilst:
> Thanks a lot for the fix. Could you also create a corresponding issue for this PR to describe the problem?
> 
> I’d also like to better understand your use case: are you using your own Parquet reader in your environment? It seems the default Parquet reader would not return a dictionary for the binary type.
> 
> Also, there appear to be several places in the current codebase that forcibly assume a string dictionary, which may have similar issues. Would you prefer to address those in this PR as well, or leave them for follow-up PRs?

We're indeed using an internal parquet reader for performance trade-offs, which will read the dictionary encoded binary into a dictionary. But the default parquet reader can also be possible to end up with a dictionary if there're `ARROW:schema` meta data in the parquet footer. It's not a common scenario though.

I guess I can address those 'forcibly assume a string dictionary' issues in this PR as well

lxy-9602:
> I guess I can address those 'forcibly assume a string dictionary' issues in this PR as well

Indeed, both Paimon-cpp and Java currently do not write out `ARROW:schema` by default, which has masked this issue. Thanks for the clarification — I look forward to the complete fix. Please feel free to ping me once it’s ready. Thanks!

zhangweilst:
> > I guess I can address those 'forcibly assume a string dictionary' issues in this PR as well
> 
> Indeed, both Paimon-cpp and Java currently do not write out `ARROW:schema` by default, which has masked this issue. Thanks for the clarification — I look forward to the complete fix. Please feel free to ping me once it’s ready. Thanks!

Updated the whole fix. Since most of the added code are tests, it looks bigger than it really is. Please help to review

src/paimon/core/casting/casting_utils.cpp
```diff
@@ -23,6 +23,19 @@
...
+Result<std::shared_ptr<arrow::Array>> CastingUtils::DecodeDictionary(
+    const std::shared_ptr<arrow::Array>& array, arrow::MemoryPool* pool) {
+    if (array->type_id() != arrow::Type::DICTIONARY) {
+        return array;
+    }
+    const auto& dictionary_type = checked_cast<const arrow::DictionaryType&>(*array->type());
+    std::shared_ptr<arrow::DataType> value_type = dictionary_type.value_type();
+    if (value_type->id() == arrow::Type::LARGE_STRING) {
+        value_type = arrow::utf8();
+    }
```
lxy-9602, line 35: Could `value_type->id()` be large binary? If so, should it be cast to binary? Since in paimon-cpp, large binary specifically refers to the blob type.

src/paimon/common/data/columnar/columnar_utils.h
```diff
@@ -73,13 +73,15 @@ class ColumnarUtils {
...
-                    checked_cast<arrow::StringArray*>(typed_array->dictionary().get());
+                    checked_cast<arrow::BinaryArray*>(typed_array->dictionary().get());
                 return dictionary->GetView(dict_index);
-            } else if (value_type_id == arrow::Type::type::LARGE_STRING) {
+            } else if (value_type_id == arrow::Type::type::LARGE_STRING ||
+                       value_type_id == arrow::Type::type::LARGE_BINARY) {
                 auto dictionary =
-                    checked_cast<arrow::LargeStringArray*>(typed_array->dictionary().get());
+                    checked_cast<arrow::LargeBinaryArray*>(typed_array->dictionary().get());
                 return dictionary->GetView(dict_index);
```
lxy-9602, line 85: Here we call the dictionary’s `GetView()` directly. If the index itself is valid but the referenced dictionary slot is null, `DictionaryArray::IsNull()` still returns false. `ColumnarRow`, `ColumnarArray`, and `ColumnarRowRef`’s `IsNullAt()` only check the outer index validity, so the null slot will later be read as a zero-length view, which is indistinguishable from a valid empty byte string.

So I’d like to confirm whether there are cases where the dictionary indices are not null, but the dictionary entries themselves are null? Since many other places already take this into account and have tests for it.

src/paimon/format/parquet/parquet_file_batch_reader_test.cpp
```diff
@@ -1983,11 +2101,8 @@ TEST_F(ParquetFileBatchReaderTest, TestDictionaryPassthrough) {
 }
 
 TEST_F(ParquetFileBatchReaderTest, TestDictionaryPassthroughSkipsBinaryColumn) {
-    // Parquet stores STRING and BINARY in the same BYTE_ARRAY leaf and dictionary-encodes both, so
-    // the gate has to exclude BINARY by logical type. It does, because nothing downstream can read
-    // `dictionary(int32, binary)`: ColumnarUtils::GetView() asserts on it and returns an empty view
-    // in a release build, and LiteralConverter rejects it. `f8` is the control - same physical
-    // type, same pages, and it is forwarded - so this fails if the exclusion is ever widened back.
+    // Keep the passthrough policy limited to STRING even though consumers also support binary
+    // dictionaries. Both columns use BYTE_ARRAY and dictionary pages; f8 is the STRING control.
```
lxy-9602, line 2105: Please add a failing Paimon example table under `test_data` and add an inte test in `scan_and_read_inte_test.cpp` for this issue, for example a table whose Paimon type is `binary` but is internally stored as `dictionary<binary>` and uses `store_arrow_schema`. I found that `ParquetReadTypeAdapter` still seems to reject this kind of data. Since the current test only covers the Parquet format layer, it may miss some framework-level validation or flow.

PR #376, opened 2026-09-21
feat(format): support Vortex file format

### LINKED ISSUE #319: [Feature] Support Vortex file format
### Search before asking

### Motivation

Apache Paimon Java supports Vortex through the paimon-vortex module, a Rust-based columnar file format with extensible compression encodings and lazily loaded summary statistics. Paimon C++ cannot currently access tables using this format.

### Solution

Add Vortex support to the Paimon C++ file-format SPI:
- Register the vortex file-format factory.
- Implement compatible schema, reader, and writer handling.
- Support column projection, predicate pushdown, and precise row selection.
- Validate the same supported and unsupported logical types as Java.
- Add C++ round-trip tests and Java-produced file compatibility tests.

### Anything else?

Vortex is implemented in Rust, so the integration can follow the Mosaic Rust FFI patterns (#238).
Java reference module: https://github.com/apache/paimon/tree/master/paimon-vortex

zhouyc-ali:
### Purpose

Linked issue: close #319

Adds optional read and write support for the Apache Paimon Vortex file format, integrated through
the upstream Vortex Rust FFI and gated by `PAIMON_ENABLE_VORTEX`.

- Add the Vortex file reader, writer, format factory, stats extractor, and Arrow C Data Interface
  conversion.
- Add a callback-based IO bridge so Vortex reads and writes directly through paimon's
  `InputStream`/`OutputStream` (works with any `FileSystem`: local, OSS, S3, Jindo), instead of
  buffering whole files in memory on read or staging a local temporary file on write. The stock
  Vortex C API cannot route IO through paimon's filesystem, so the bridge is added to `vortex-ffi`
  as an injected Rust module (`crates/vortex_callback_io/`) plus a one-line patch
  (`cmake_modules/vortex.diff`).
- Support projection and configurable output batch sizes; reset the reader to the beginning when
  `SetReadSchema` is called.
- Read the file row count for `FileInfo`; Vortex exposes no per-column statistics to Paimon
  (consistent with the Java implementation).
- Validate Vortex-incompatible schema types (MAP, MULTISET, VARIANT, BLOB).
- Register Vortex in the file-format factory and bypass the generic `PrefetchFileBatchReader`
  (Vortex performs random-access reads through its own callbacks).
- Enable Vortex in CI: add `-DPAIMON_ENABLE_VORTEX=ON` to `build_paimon.sh`, install the `stable`
  Rust toolchain (Vortex MSRV 1.95), and build `flatc` 25.12.19 (`ci/scripts/setup_flatc.sh`), which
  Vortex needs to compile its FlatBuffers schemas at build time.

Predicate/projection pushdown into Vortex is not implemented; all rows and columns are scanned and
projected to the read schema.

### Tests

- Vortex unit tests: round-trip read/write across multiple batch sizes; primitive, string, binary,
  date, timestamp, decimal and list types; row-count statistics extraction; projected reads; and IO
  fault injection at every read and write position (verifying paimon IO errors surface through the
  FFI callback boundary rather than an opaque Vortex error).
- Rust unit tests (`cargo test -p vortex-ffi callback_io`): callback read round-trip, callback
  write→read round-trip, and context release on open failure.
- Integration test `TestVortexJavaCompatibility`: reads an append table written by Paimon Java 2.0.0
  (`paimon-vortex-format` + `paimon-vortex-jni`).
- Full CTest suite passes (33/33 targets); clang-format, cpplint, cmake-format and rustfmt clean.

### API and Format

Adds the optional Vortex file-format factory and the `PAIMON_ENABLE_VORTEX` CMake option. It reads
and writes the existing Apache Paimon Vortex file format through the upstream Vortex Rust FFI and
does not change the Vortex binary format or any existing public API under `include/`. Building Vortex
requires a Rust toolchain (>= 1.95) and `flatc` 25.12.19; both are set up in CI.

### Documentation

Updated `docs/source/user_guide/format_table.rst` to list Vortex among the formats a format table
does not reach, and added `crates/vortex_callback_io/README.md` describing the injected module and
its build wiring.

### Generative AI tooling

Generated-by: Qoder


src/paimon/core/table/format/format_table.cpp
```diff
@@ -67,7 +67,7 @@ Result<FormatTable::Format> FormatTable::ParseFormat(const std::string& file_for
         return Format::ORC;
     }
     if (normalized == "csv" || normalized == "text" || normalized == "json" ||
-        normalized == "mosaic") {
+        normalized == "mosaic" || normalized == "vortex") {
         return Status::NotImplemented(
```
lxy-9602, line 71: Can we double-check if Lance supports this? If it doesn't, we should add this `NotImplemented` here.

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+}
+
+// Rebuilds `array` with every StringView/BinaryView column (which Vortex's Arrow export produces)
+// converted to the standard string/binary type. Arrow 17 has no cast kernel between the view and
+// standard types, so the leaves are rebuilt with builders and the parents reconstructed, recursing
+// through structs and lists. Arrays with no view-typed data are returned unchanged. `array` must be
+// offset-normalized (offset 0) so the rebuilt parents and children stay consistent.
+Result<std::shared_ptr<arrow::Array>> NormalizeViewArray(const std::shared_ptr<arrow::Array>& array,
+                                                         arrow::MemoryPool* pool) {
+    switch (array->type_id()) {
```
lxy-9602, line 91: Please add a TODO here to track this potential performance hotspot for future optimization.

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+            const auto& list = checked_cast<const arrow::LargeListArray&>(*array);
+            PAIMON_ASSIGN_OR_RAISE(std::shared_ptr<arrow::Array> values,
+                                   NormalizeViewArray(list.values(), pool));
+            if (values.get() == list.values().get()) {
+                return array;
+            }
+            return std::make_shared<arrow::LargeListArray>(
+                arrow::large_list(array->type()->field(0)->WithType(values->type())), list.length(),
+                list.value_offsets(), values, list.null_bitmap(), list.null_count(), list.offset());
+        }
```
lxy-9602, line 163: It seems that LARGE_LIST is not supported?

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+      input_context_(std::move(input_context)),
+      session_(std::move(session)),
+      data_source_(std::move(data_source)),
+      scan_(std::move(scan)),
+      file_schema_(NormalizeViewSchema(file_schema)),
+      struct_type_(struct_type),
+      total_rows_(total_rows),
+      pool_(pool),
+      arrow_pool_(arrow_pool),
+      metrics_(std::make_shared<MetricsImpl>()) {}
```
lxy-9602, line 282: I'm curious about the reason for the Paimon pool here. Looks like this pool is unused?

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+    }
+}
+
+// Recursively projects `array` onto `target_type` (the read schema's type for this level),
+// selecting struct children by field name and pruning nested struct/list/fixed-size-list elements
+// so the returned array's type equals `target_type`. Returns `array` unchanged when its type
+// already matches. Leaf types are expected to match after view normalization; an unexpected leaf
+// mismatch is returned as-is for the caller's schema check to surface.
+Result<std::shared_ptr<arrow::Array>> ProjectArrayToType(
+    const std::shared_ptr<arrow::Array>& array,
```
lxy-9602, line 188: This function can be replaced with `NestedProjectionUtils::AlignArrayToReadType`. Also, does Vortex not support column pruning? Do we need to read everything back and then prune it?

lxy-9602: If you plan to support this in the next PR, please add a TODO to indicate that this issue is known and will be addressed later. If not supoort, there’s no need to add this kind of fallback; we can simply avoid testing scenarios like nested subcolumn pushdown.

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+    CloseInternal();
+}
+
+void VortexFileBatchReader::ReleaseStream() {
+    if (stream_active_) {
+        if (current_stream_.release != nullptr) {
+            current_stream_.release(&current_stream_);
+        }
+        current_stream_ = ::ArrowArrayStream{};
+        stream_active_ = false;
```
lxy-9602, line 534: Use `ArrowArrayStreamRelease`

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+      metrics_(std::make_shared<MetricsImpl>()) {}
+
+Result<std::unique_ptr<VortexFileBatchReader>> VortexFileBatchReader::Create(
+    const std::shared_ptr<InputStream>& input, int32_t batch_size,
+    const std::shared_ptr<MemoryPool>& pool, const std::shared_ptr<arrow::MemoryPool>& arrow_pool) {
+    if (input == nullptr || pool == nullptr || batch_size <= 0) {
+        return Status::Invalid(
+            "Vortex reader requires non-null input and memory pool, and positive batch size");
+    }
+    // Vortex reads the file through positional callbacks into `input`, so nothing is staged here.
```
lxy-9602, line 291: Why not check `arrow_pool`?

src/paimon/format/vortex/vortex_file_batch_reader.cpp
```diff
@@ -0,0 +1,559 @@
...
+    ::ArrowSchema* read_schema, const std::shared_ptr<Predicate>& predicate,
+    const std::optional<RoaringBitmap32>& selection_bitmap) {
+    if (read_schema == nullptr) {
+        return Status::Invalid("Vortex read schema is nullptr");
+    }
+    (void)selection_bitmap;
+    // Projection and predicates are not pushed into Vortex: the whole file is scanned and
+    // NextBatch projects the result to `read_schema`. Reading restarts from the first row, so drop
+    // the single-use scan and any open stream; the next read re-creates the scan.
+    PAIMON_ASSIGN_OR_RAISE_FROM_ARROW(std::shared_ptr<arrow::Schema> imported_read_schema,
```
lxy-9602, line 489: Why doesn’t `selection_bitmap` support pushdown either? Vortex’s advantage over Parquet is point lookup, but if we disable its point-lookup capability, then introducing Vortex becomes meaningless. Similarly, at minimum, please add a TODO here.

include/paimon/type_fwd.h
```diff
@@ -38,7 +38,10 @@ class Result;
 
 class Status;
 
-class MemoryPool;
+// paimon::MemoryPool is a real public type (see paimon/memory/); this forward declaration is
+// intentional. The NOLINT silences bugprone-forward-declaration-namespace, which otherwise fires in
+// any translation unit that includes this header and happens to use arrow::MemoryPool.
+class MemoryPool;  // NOLINT(bugprone-forward-declaration-namespace)
```
lxy-9602, line 45: Why was this changed here? It looks like the other plugins are fine.

src/paimon/format/vortex/vortex_format_writer.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace paimon::vortex {
+
+/// Writes a Vortex file.
+///
+/// Bytes go straight to the paimon `OutputStream` through `vx_callback_sink_open`, the
+/// callback-based sink this repository adds to vortex-ffi: Vortex writes into `output_context_`,
+/// which forwards to the stream. Vortex's own sink (`vx_array_sink_open_file`) can only create a
+/// local file, which would mean staging the file on local disk and copying it back on finish.
+class VortexFormatWriter : public FormatWriter {
```
lxy-9602, line 49: ”Vortex's own sink (`vx_array_sink_open_file`) can only create a local file, which would mean staging the file on local disk and copying it back on finish.“ This probably doesn’t need to be in the class comment. Please check for similar issues as well.

src/paimon/format/vortex/vortex_stats_extractor.cpp
```diff
@@ -0,0 +1,74 @@
...
+    }
+    vx_estimate row_count{};
+    vx_data_source_get_row_count(data_source.get(), &row_count);
+    if (row_count.type == VX_ESTIMATE_UNKNOWN) {
+        return Status::Invalid("Vortex file did not report a row count");
+    }
+    const auto rows = static_cast<int64_t>(row_count.estimate);
+    // Empty column stats: Vortex keeps its statistics internal and does not surface them to Paimon.
+    return std::make_pair(ColumnStatsVector(), FileInfo(rows));
+}
```
lxy-9602, line 72: If there is no statistics information, the returned value should be a `ColumnStatsVector` containing one element for each field, with each element being an empty stats object. Please refer to `LanceStatsExtractor`. The current implementation returns an empty `StatsVector`, which may cause issues when reading, since the field idx is used for the check.

test/inte/scan_and_read_inte_test.cpp
```diff
@@ -3690,6 +3690,49 @@ TEST_F(ScanAndReadInteTest, TestMosaicJavaAndPythonCompatibility) {
...
 #endif
 
+#ifdef PAIMON_ENABLE_VORTEX
+TEST_F(ScanAndReadInteTest, TestVortexJavaCompatibility) {
+    // Fixture written by Paimon Java 2.0.0 with file.format=vortex; see
+    // test/test_data/vortex/append_java_compat.db/append_java_compat/README.md. This verifies that
+    // paimon-cpp reads back a Vortex table produced by the Java implementation.
+    arrow::FieldVector fields = {
+        arrow::field("_VALUE_KIND", arrow::int8()),
+        arrow::field("id", arrow::int32()),
```
lxy-9602, line 3700: Please add complete compatibility tests for Vortex supporting all field types, similar to `test_data/mosaic`.

test/test_data/vortex/append_java_compat.db/append_java_compat/README.md
```diff
@@ -0,0 +1,24 @@
...
+(2, NULL)
+(3, "three")
+(4, "four")
+(5, "five")
+
+Generated by a standalone Java program using the Paimon batch write API
+(BatchWriteBuilder / BatchTableWrite / BatchTableCommit) on JDK 17 with
+--add-opens=java.base/java.nio=ALL-UNNAMED (required by Arrow Java memory).
+Read back and verified by paimon-cpp in
+ScanAndReadInteTest.TestVortexJavaCompatibility.
```
lxy-9602, line 24: Please remove this. This test data may be used by other tests in the future, so it shouldn’t be coupled to `TestVortexJavaCompatibility`.

test/inte/scan_and_read_inte_test.cpp
```diff
@@ -3690,6 +3690,49 @@ TEST_F(ScanAndReadInteTest, TestMosaicJavaAndPythonCompatibility) {
...
+        arrow::field("name", arrow::utf8()),
+    };
+    std::shared_ptr<arrow::DataType> data_type = arrow::struct_(fields);
+    std::shared_ptr<arrow::Array> expected_array =
+        arrow::ipc::internal::json::ArrayFromJSON(data_type, R"([
```
lxy-9602, line 3705: Also, I see write support has already been added. Are you planning to leave the combined read/write `TEST_P` cases from the other integration tests for a follow-up commit?

src/paimon/format/vortex/vortex_file_format_test.cpp
```diff
@@ -0,0 +1,520 @@
...
+                                                        /*selection_bitmap=*/std::nullopt),
+                          i);
+
+        bool eof = false;
+        while (!eof) {
+            Result<BatchReader::ReadBatch> batch = reader.value()->NextBatch();
+            CHECK_HOOK_STATUS(batch.status(), i);
+            eof = BatchReader::IsEofBatch(batch.value());
+        }
+        run_complete = true;
```
lxy-9602, line 473: If the read eventually succeeds, please verify that the final read result matches the expectation. You can use `ReadResultCollector` for this.

src/paimon/format/vortex/vortex_file_format_test.cpp
```diff
@@ -0,0 +1,520 @@
...
+
+        ::ArrowArray ffi_array = {};
+        ASSERT_TRUE(arrow::ExportArray(*data, &ffi_array).ok());
+        CHECK_HOOK_STATUS(writer.value()->AddBatch(&ffi_array), i);
+        CHECK_HOOK_STATUS(writer.value()->Finish(), i);
+        CHECK_HOOK_STATUS(output_stream->Close(), i);
+        run_complete = true;
+        break;
+    }
+    ASSERT_TRUE(run_complete);
```
lxy-9602, line 517: Similarly, when the write eventually succeeds, please also verify that the written data can be read back and matches the expectation.

PR #377, opened 2026-09-21
feat: expose BlobDescriptor as public API

Smith-Cruise:
### Purpose

Expose the existing `BlobDescriptor` API to external consumers by moving its header into the public include tree. This allows integrations to detect, deserialize, and inspect blob descriptors without duplicating the binary format parser.

The implementation and serialized representation are unchanged. Internal includes now use the public header path.

### Tests

No need

### API and Format

Adds the installed public header `paimon/data/blob_descriptor.h`. Existing exported API are unchanged.

### Generative AI tooling


include/paimon/data/blob_descriptor.h
lxy-9602: Please move `kMagic` and `kMinDescriptorLength` into the `.cpp` file to keep the implementation details hidden.

Smith-Cruise: Done in 5c718d8. Both kMagic and kMinDescriptorLength now live in the .cpp anonymous namespace.

src/paimon/common/data/blob_descriptor.cpp
```diff
@@ -31,6 +31,13 @@
 #include "paimon/status.h"
 
 namespace paimon {
+namespace {
```
Smith-Cruise, line 34: anonymous namespace

lxy-9602 (approved):
+1

PR #378, opened 2026-09-21
fix(read): support Parquet TIME columns

### LINKED ISSUE #292: [Feature] Support reading Paimon TIME columns
### Search before asking

- [x] I searched the issues and found nothing similar.

### Motivation

Reading a Paimon table containing a `TIME` column fails during schema deserialization:

```text
Invalid: deserialize failed, possibly type incompatible:
parse data type failed, error msg: Invalid: Unsupported type: TIME
```

The Parquet table was created with PyPaimon 2.0.0 using an Arrow `time32[ms]` column containing `00:00:00.123`, `23:59:59.999`, and `NULL`. PyPaimon successfully writes and reads these values, but reading through the Vane Paimon extension fails.

Tested with paimon-cpp commit `366323740cc13b5e113d9465080624b28e3880e8`, bundled in extension commit `b0ac989`.

### Solution

Support Paimon `TIME` schema parsing and the corresponding Arrow time representation in the read path. Add a Parquet round-trip test covering millisecond precision, boundary values, and `NULL`.

The current [type parser](https://github.com/apache/paimon-cpp/blob/a42670e066bba60100317cc0aa5a854b2273bf98/src/paimon/common/types/data_type_json_parser.cpp) still has no `TIME` handling branch.

### Anything else?

The failure was reproduced through Vane, not a standalone paimon-cpp executable.

QuakeWang:
### Purpose

Closes #292.

TIME schemas were rejected, and field mapping required a logical cast type even for identical Arrow types. This prevented reading existing Parquet TIME columns.

Parse and validate TIME as time32[ms], matching Java Paimon, and skip cast lookup for identical types. Millisecond values and nulls are preserved, including PyPaimon's TIME(0).

### Tests

- Passed: 7 parser, 85 core, 69 Parquet reading tests, the new TIME integration test, and an actual PyPaimon 2.0.0 table scan/read.
- Full-repository pre-commit and git diff --check passed.
- macOS core/read tests required static relinking. Full common tests hit an existing compilation error; full Parquet tests aborted in a VECTOR test. Lance was not enabled.

### API and Format

No public API or storage format changes. Scope is reading existing Parquet TIME columns.

### Documentation

No documentation changes.

### Generative AI tooling


lxy-9602:
Thank you for the contribution! Before reviewing this PR in detail, I’d like to confirm the intended scope of the implementation. Does this PR only support reading append tables? It seems primary key tables may require additional changes, for example around `FieldsComparator`.

Also, although the `TIME` schema can pass validation during creation, it currently cannot be serialized when create table— for example, calling `ToJsonString()` fails with an `unknown type` error.

QuakeWang:
> Thank you for the contribution! Before reviewing this PR in detail, I’d like to confirm the intended scope of the implementation. Does this PR only support reading append tables? It seems primary key tables may require additional changes, for example around `FieldsComparator`.
> 
> Also, although the `TIME` schema can pass validation during creation, it currently cannot be serialized when create table— for example, calling `ToJsonString()` fails with an `unknown type` error.

@lxy-9602 Yes, this PR currently targets reading existing append-only Parquet tables. I should have made that clearer in the description. You’re right about PK tables. 

I checked the merge path, and `RowToArrowArrayConverter` also needs TIME support, beyond the comparator changes. 

I missed the serialization gap. I’d like to address that here and leave PK support for a separate PR. Does that scope work for you? For serialization, we’ll also need to decide how to handle precision, since the current mapping to `time32[ms]` loses the original `TIME(p)` precision.

lxy-9602:
> I missed the serialization gap. I’d like to address that here and leave PK support for a separate PR. Does that scope work for you? For serialization, we’ll also need to decide how to handle precision, since the current mapping to `time32[ms]` loses the original `TIME(p)` precision.

Thanks for your reply! Yeah, I agreed that PK support should be a separate pr. I think C++ can follow the paimon-rust design: preserve `TIME(p)` in the Paimon logical schema and treat Arrow `time32[ms]` only as the execution/storage representation. Serialization should use the preserved logical precision instead of inferring it from the Arrow type. If introducing a separate logical type model is out of scope for this PR, storing the original precision in Arrow field metadata could be a transitional solution. Does this look good to you?

QuakeWang:
Updated in c23c267. TIME precision is now preserved in Arrow field metadata and used for serialization, with `TIME(0)` as the default when metadata is absent. Added round-trip tests and updated the compatibility tests. PK support will stay in a separate PR.

src/paimon/common/types/data_type.h
```diff
@@ -37,6 +38,14 @@ namespace paimon {
 
 class DataType : public Jsonizable<DataType> {
  public:
+    static constexpr char TIME_PRECISION[] = "paimon.time.precision";
+
```
lxy-9602, line 43: Could you please update the naming of the new variables to follow the `kTimePrecision` format?

src/paimon/common/types/data_type.h
```diff
@@ -37,6 +38,14 @@ namespace paimon {
...
+    // Arrow carries milliseconds, while metadata preserves the declared TIME precision.
+    // Arrow-only schemas use Paimon's default precision of zero.
+    static Result<int32_t> GetTimePrecision(
+        const std::shared_ptr<arrow::DataType>& type,
+        const std::shared_ptr<const arrow::KeyValueMetadata>& metadata);
```
lxy-9602: It seems we could pass `arrow::Field` directly here?

src/paimon/core/schema/arrow_schema_validator.cpp
```diff
@@ -208,6 +210,10 @@ Status ArrowSchemaValidator::ValidateField(const std::shared_ptr<arrow::Field>&
         case arrow::Type::type::DATE32:
         case arrow::Type::type::TIMESTAMP:
             break;
+        case arrow::Type::type::TIME32:
+            PAIMON_RETURN_NOT_OK(
+                DataType::GetTimePrecision(field->type(), field->metadata()).status());
```
lxy-9602: `PAIMON_RETURN_NOT_OK(DataType::GetTimePrecision(field->type(), field->metadata()));` seems ok.

test/inte/read_inte_test.cpp
```diff
@@ -88,6 +92,94 @@
...
+            }
+            auto expected = arrow::StructArray::Make(expected_columns, expected_fields);
+            ASSERT_TRUE(expected.ok());
+            ASSERT_TRUE(
+                result->Equals(std::make_shared<arrow::ChunkedArray>(expected.ValueOrDie())))
+                << result->ToString();
+        }
+    }
+}
+
```
lxy-9602: It seems this PR currently only plans to support read operations. Is write support out of scope for now?

- If write is supported: We could refer to the tests in `write_and_read_inte_test.cpp` for end-to-end read/write testing. Manually constructing raw data files and metadata for integration tests is a bit too complex. Also, I assume statistics (like parquet_stats_extractor) are not supported yet for writing?
- If only read is planned: I noticed you've already added the coverage in the `compat_inte` tests, so it seems this extra case is no longer necessary.
- Time Type as Partition Column: Does the Time type support being used as a partition column? If so, it seems there are other places needing adjustments and extra tests. We can either leave this for a follow-up PR, or explicitly block it in `schema_validation.cpp` for now.

lxy-9602 (approved):
+1

PR #380, opened 2026-09-21
fix(types): assign omitted field IDs in type JSON

### LINKED ISSUE #347: [Bug] Configured Variant shredding schema without field IDs is rejected unlike Java Paimon
### Search before asking

### Paimon-cpp version

- Apache Paimon C++ v0.3.0 (`efbfc848`).
- The problem is also present on current `main` (`ccfe346670f32b82d7c9139a8f3d356c33dc5b7d`).
- Java behavior used for comparison: Apache Paimon 1.4.2.

### Minimal reproduce step

Create a logical schema with a top-level Variant column and configure `variant.shreddingSchema` with a valid ROW type whose fields omit every `id`, for example:

```json
{
  "type": "ROW",
  "fields": [
    {
      "name": "payload",
      "type": {
        "type": "ROW",
        "fields": [
          {"name": "age", "type": "INT"},
          {"name": "profile", "type": {
            "type": "ROW",
            "fields": [{"name": "city", "type": "STRING"}]
          }}
        ]
      }
    }
  ]
}
```

Pass this value to `VariantShreddingWritePlan::FromConfiguredSchema` (normally through `variant.shreddingSchema` or its Parquet fallback option) and open a writer.

Paimon-cpp rejects the configuration while deserializing the first `DataField`, because `DataField::FromJson` unconditionally requires the `id` key. In an embedding engine the propagated exception is equivalent to:

```text
key 'id' must exist
```

The same schema is accepted by Java Paimon 1.4.2. `DataTypeJsonParser.parseDataType` starts an `AtomicInteger` at `-1` and assigns deterministic IDs in traversal order when all field IDs are omitted.

**Expected behavior**

Paimon-cpp should accept the same configured Variant shredding schema JSON as Java Paimon:

1. If every ROW field omits `id`, assign deterministic IDs in the same traversal order as Java, starting at 0.
2. If every ROW field provides `id`, preserve those IDs.
3. If IDs are only partially present, reject the schema with an explicit diagnostic such as `Partial field id is not allowed.`

**Actual behavior**

Paimon-cpp accepts only the second form. A valid all-ID-omitted schema accepted by Java is rejected before the Variant shredding plan can be created.

This creates a cross-language compatibility gap for engines that consume table options written by Java/Spark. Such engines must currently route an otherwise supported Variant write back to the Java SDK solely because the configured schema omits optional field IDs.

### Anything else?

The mismatch comes from the generic JSON parsing path:

- [`VariantShreddingWritePlan::FromConfiguredSchema`](https://github.com/apache/paimon-cpp/blob/ccfe346670f32b82d7c9139a8f3d356c33dc5b7d/src/paimon/common/data/variant/variant_shredding_write_plan.cpp#L185) delegates the configured ROW JSON to `DataTypeJsonParser::ParseType`.
- [`DataTypeJsonParser::ParseRowType`](https://github.com/apache/paimon-cpp/blob/ccfe346670f32b82d7c9139a8f3d356c33dc5b7d/src/paimon/common/types/data_type_json_parser.cpp#L703) deserializes every ROW child as a `DataField`.
- [`DataField::FromJson`](https://github.com/apache/paimon-cpp/blob/ccfe346670f32b82d7c9139a8f3d356c33dc5b7d/src/paimon/common/types/data_field.cpp#L67) requires `id` unconditionally.

A contained fix could normalize the configured schema before normal DataType parsing: recursively count ID presence across ROW fields; generate preorder IDs only when none are present; preserve the current error behavior for partial IDs. Alternatively, the generic `DataTypeJsonParser` can implement Java's all-or-none semantics if that behavior is desired for every DataType JSON consumer.

Suggested regression coverage:

- all IDs omitted, including nested ROW/ARRAY/MAP descendants;
- all IDs explicitly present;
- partially present IDs are rejected;
- configured shredding write/read round trip uses the generated IDs in the physical Parquet schema.

This is separate from #344: that issue concerns missing IDs in the generated physical shredded Variant subtree, whereas this issue concerns parsing IDs in the user-configured logical shredding schema.

SteNicholas:
### Purpose

Linked issue: close #347

Java's `DataTypeJsonParser` numbers the ROW fields of a parsed type when none of them carries an id, so a configured `variant.shreddingSchema` whose fields all omit their ids is accepted by Java Paimon but was rejected here while deserializing the first `DataField` (`key 'id' must exist`). An embedding engine therefore had to route an otherwise supported Variant write back to the Java SDK only because the configured schema omitted optional field ids.

This threads a field id assigner through the data type JSON parsing, mirroring Java's all-or-none rule:

- ids are generated in traversal order starting at 0 when no ROW field carries one;
- explicit ids are preserved at every level;
- a type that mixes both forms is rejected with `Partial field id is not allowed.`

One deliberate deviation from Java: Java rejects the mix only when a generated id comes first, so an explicit id followed by an omitted one silently produces duplicate ids there, while both orders are rejected here. Deserializing a stored table schema still requires every field to carry its own id, as before.

### Tests

- `DataTypeJsonParserTest.ParseTypeRowTypeGeneratesMissingFieldIds`: ids generated in traversal order across nested ROW, ARRAY and MAP descendants.
- `DataTypeJsonParserTest.ParseTypeRowTypeSuccess`: non-sequential explicit ids preserved at every level.
- `DataTypeJsonParserTest.ParseTypeComplexTypeFailure`: partial ids rejected in either order, alongside the malformed complex type cases.
- `DataFieldTest.FromJson` and `DataFieldTest.FromJsonFailed`: a stored schema still requires an id on every field, names and descriptions keep embedded NULs, and each malformed field key reports its own diagnostic.
- `VariantShreddingWritePlanFactoryTest.ConfiguredSchema`: `variant.shreddingSchema` and its `parquet.variant.shreddingSchema` fallback plan the same physical schema, field ids included, whether the configured schema carries ids or omits them.
- `VariantShreddingWritePlanFactoryTest.ConfiguredSchemaRejectsPartialFieldIds`.
- `VariantParquetTest.ShreddedWriteAndReadRoundTrip`: the new `configured-without-ids` mode asserts the generated ids reach the Parquet physical schema and that the shredded values round trip.

### API and Format

`include/paimon/defs.h` only gains documentation for `variant.shreddingSchema`; there is no public symbol, storage format or protocol change. A configured schema that already carries field ids produces exactly the same physical schema as before.

### Documentation

`docs/source/user_guide/data_types.rst` documents that the configured shredding schema's fields either all carry an `id` or all omit one, in which case the ids are assigned in traversal order starting at 0.

### Generative AI tooling

Generated-by: Claude Opus 5


lxy-9602 (approved):
+1

lxy-9602:
Thanks to @SteNicholas  for the fix. @suxiaogang223  , could you please check again whether the issue #347 is resolved after this PR is merged? Thanks!

